PR #1, opened 2025-03-24
ci: add workflow to build debos recipe

lool:
Build debos recipe in a fresh container; image is currently discarded.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -1,0 +1,90 @@
+name: Build debos recipe
+
+on:
+  # run on pull requests to the main branch
+  pull_request:
+    branches: [main]
+  # run on pushes to the main branch
+  push:
+    branches: [main]
+  # run daily at 8:30am
+  schedule:
+    - cron: '30 8 * * *'
+  # allow manual runs
+  workflow_dispatch:
+
+# only need permission to read repository; implicitely set all other
+# permissions to none
+permissions:
+  contents: read
+
+defaults:
+  # run all commands from the debos-recipes directory
+  run:
+    working-directory: debos-recipes
+
+env:
+  INCUS_IMAGE: images:debian/trixie/arm64
+  INCUS_NAME: debos
+
+# cancel in progress builds for this workflow triggered by the same ref
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-debos:
+    runs-on: [self-hosted, arm64, debbuilder]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # this is the default in our self-hosted runners
+      - name: Make sure Incus is setup
+        run: |
+          set -x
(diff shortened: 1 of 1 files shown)


lool:
While there are some self-hosted runners configured on this repo, there were no arm64 ones at the time of submitting this pull request, so the new workflow defined here didn't run for this pull request yet :)

ndechesne:
> While there are some self-hosted runners configured on this repo, there were no arm64 ones at the time of submitting this pull request, so the new workflow defined here didn't run for this pull request yet :)

I think it did not run because you are limiting the pull_request to changes in debos-recipes/* . So this PR would not trigger the workflow. I don't think we should have the filtering based on path here.

.github/workflows/debos.yml
```diff
@@ -0,0 +1,86 @@
...
+    paths: ['debos-recipes/*']
+  # run daily at 8:30am
+  schedule:
+    - cron: '30 8 * * *'
+  # allow manual runs
+  workflow_dispatch:
+
+# only need permission to read repository; implicitely set all other
+# permissions to none
+permissions:
```
ndechesne, line 20: why do you need that?

.github/workflows/debos.yml
```diff
@@ -0,0 +1,86 @@
...
+defaults:
+  # run all commands from the debos-recipes directory
+  run:
+    working-directory: debos-recipes
+
+env:
+  INCUS_IMAGE: images:debian/trixie/arm64
+  INCUS_NAME: debos
+
+# cancel in progress builds for this workflow triggered by the same ref
```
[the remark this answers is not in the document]
lool: If you push repeatedly to the same pull request

.github/workflows/debos.yml
lool: By default, everything is allowed because potentially workflows could be used for anything, but we know what the workflow will do

.github/workflows/debos.yml
```diff
@@ -0,0 +1,89 @@
+name: Build debos recipe
+
+on:
+  # run on pull requests to the main branch if debos recipes are changed
+  pull_request:
+    branches: [main]
+    paths: ['.github/workflows/debos.yml', 'debos-recipes/*']
```
ndechesne, line 7: I really don't know why you want to limit the CI like that. Yes, we are saving a 'build' if we only update the README.. but the risk is that we are missing an important change (e.g. a change in script folder). I think it's ok to just test all PRs. We can always limit later, if it's really a problem, no?

lool: Sure, this is to be conservative in not triggering things needlessly. We'll be adding a lot of downstream tests over time, and we might build other artifacts from this repo, but one could argue this is premature optimization.

ricardosalveti: Yeah, we can limit it later if needed.

lool: I've removed the paths in both event triggers and updated comments; a build with an arm64 self-hosted runner also just passed before this change

ndechesne (approved):

PR #3, opened 2025-03-28
ci: debos: Update OS packages first

lool:
During some GitHub workflow runs, incus exec commands are randomly
interrupted by:
    Error: websocket: close 1006 (abnormal closure): unexpected EOF

It's possible that this is due to an OS update happening during the run,
so pre-emptively do the updates first, which is a good idea for bug
fixes and predicatibility anyway.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -40,6 +40,15 @@ jobs:
         with:
           fetch-depth: 0
 
+      # make sure we have latest packages first, to get latest fixes and to
+      # avoid an automated update while we're building
+      - name: Update OS packages
+        run: |
+          set -x
+          sudo apt update
+          sudo apt -y upgrade
+          sudo apt -y full-upgrade
+
       # this is the default in our self-hosted runners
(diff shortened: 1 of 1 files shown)


ndechesne (approved):

ricardosalveti:
> It's possible that this is due to an OS update happening during the run,
> so pre-emptively do the updates first, which is a good idea for bug
> fixes and predicatibility anyway.

Is there a way to block the auto update? Doing apt-get update / upgrade on every run will cause the builder behavior to potentially change as well.

lool:
There might also be a gcp agent doing some magic.

doanac:
I could also update the base OS image. Then your CI jobs could go a little faster and be on something more predictable. And if there's some way to disable any type of background update (assuming that could be at play - i'm somewhat skeptical) - I'm happy to add that.

ricardosalveti:
Yeah, let's try with this update approach on every run to see if that works around the issue, then we can isolate.

ricardosalveti (approved):

PR #4, opened 2025-03-28
ci: debos: upload artifacts to fileserver

lool:


.github/workflows/debos.yml
```diff
@@ -97,3 +99,18 @@ jobs:
...
+          # curl will be used to talk to fileserver; should be installed by
+          # default
+          sudo apt -y install curl
+          # create a directory for the current run
+          dir="${FILESERVER_DIR}/${GITHUB_RUN_ID}"
+          mkdir -vp "${dir}"
+          # copy output files
+          cp -v disk.img "${dir}"
+          # instruct fileserver to publish this directory
+          url="${FILESERVER_URL}/${GITHUB_RUN_ID}/"
```
ndechesne, line 114: GITHUB_RUN_ID is unique within the repo. We seem to publish builds from this repo at the same level as builds from meta-qcom. that could clash no?

ndechesne:
> We could:
> 
> 1. do multiple debos runs with different sector sizes (wasteful IMO)
> 2. generate an image for the other sector size at the end of a debos build, and always publish both
> 3. defer the conversion to flashing time / end-users and keep building only one image
> 

I think we should publish working (and tested) images, so we cannot defer to the user to do the final conversion.  We can do either 1 or 2 above, but I would prefer 1. it seems safer, in case debos ends up doing something 'more' with the sector size than what a custom/local script could do. 

Could debos be split into 2 steps, first to prepare the rootfs, and then to create the image ? so that we 'waste' less? 

> There's also a side question of generating the other assets needed for flashing, but that is a larger topic.

Right now most of the boot assets , qdl xml files, .. are managed/integrated in meta-qcom, and 'tied' to our Yocto builds, but it's not ideal. We would like to have them in https://github.com/qualcomm-linux/qcom-ptool instead, so that they are managed independently of meta-qcom.

.github/workflows/debos.yml
lool: Ah indeed, I assumed the run ids were globally unique, but the doc says unique per repo.

There's also the case for reruns, so a run id is actually not a unique build id; so we should also use GITHUB_RUN_ATTEMPT to have unique build ids.

lool:
> > We could:
> > 
> > 1. do multiple debos runs with different sector sizes (wasteful IMO)
> > 2. generate an image for the other sector size at the end of a debos build, and always publish both
> > 3. defer the conversion to flashing time / end-users and keep building only one image
> 
> I think we should publish working (and tested) images, so we cannot defer to the user to do the final conversion. We can do either 1 or 2 above, but I would prefer 1. it seems safer, in case debos ends up doing something 'more' with the sector size than what a custom/local script could do.
> 
> Could debos be split into 2 steps, first to prepare the rootfs, and then to create the image ? so that we 'waste' less?

Yes, if we want to use debos to do the conversion, then it's a good idea to architect it in two steps.

> > There's also a side question of generating the other assets needed for flashing, but that is a larger topic.
> 
> Right now most of the boot assets , qdl xml files, .. are managed/integrated in meta-qcom, and 'tied' to our Yocto builds, but it's not ideal. We would like to have them in https://github.com/qualcomm-linux/qcom-ptool instead, so that they are managed independently of meta-qcom.

Is there a place where this is discussed/being spec-ed that I can tune into?

.github/workflows/debos.yml
lool: There is no globally unique run id provided by GitHub, so some repo id needs to be used. There is a repository id as a number, and there is the repository name which is the nicest IMO, but it's controller by the submitter of pull request.

I've found some non-official repositories outlining the GitHub user/org naming rules, and these explanations in the GitHub docs about user names:
https://docs.github.com/en/enterprise-cloud@latest/admin/managing-iam/iam-configuration-reference/username-considerations-for-external-authentication

Basically usernames are alphanumeric and can contain hyphens and underscores, but never two hyphens or start with an hyphen, so that should be safe to use in directory names and URLs.

ndechesne: @doanac , FYI.

ndechesne:
> > > There's also a side question of generating the other assets needed for flashing, but that is a larger topic.
> > 
> > 
> > Right now most of the boot assets , qdl xml files, .. are managed/integrated in meta-qcom, and 'tied' to our Yocto builds, but it's not ideal. We would like to have them in https://github.com/qualcomm-linux/qcom-ptool instead, so that they are managed independently of meta-qcom.
> 
> Is there a place where this is discussed/being spec-ed that I can tune into?

We have https://github.com/qualcomm-linux/meta-qcom/issues/714. but really not much was done so far.

.github/workflows/debos.yml
```diff
@@ -97,3 +99,21 @@ jobs:
           EOF
           ) | sudo incus exec "${INCUS_NAME}" -- sh
 
+      - name: Upload artifacts to fileserver
```
ndechesne, line 103: @doanac to review/approve this change. In particular to check if the following would work with the fileserver we have:

    + id=qualcomm-linux/qcom-deb-images-14170644209-1
    + dir=/srv/gh-runners/quic-yocto/builds/qualcomm-linux/qcom-deb-images-14170644209-1
    + mkdir -vp /srv/gh-runners/quic-yocto/builds/qualcomm-linux/qcom-deb-images-14170644209-1
    mkdir: created directory '/srv/gh-runners/quic-yocto/builds/qualcomm-linux/qcom-deb-images-14170644209-1'
    + cp -v disk.img /srv/gh-runners/quic-yocto/builds/qualcomm-linux/qcom-deb-images-14170644209-1
    'disk.img' -> '/srv/gh-runners/quic-yocto/builds/qualcomm-linux/qcom-deb-images-14170644209-1/disk.img'
    + url=https://quic-yocto-fileserver-1029608027416.us-central1.run.app/qualcomm-linux/qcom-deb-images- 
    14170644209-1/
    + curl -X POST -H 'Accept: text/event-stream' https://quic-yocto-fileserver-1029608027416.us- 
    central1.run.app/qualcomm-linux/qcom-deb-images-14170644209-1/

.github/workflows/debos.yml
doanac: ugh. i think this is going to wind up forcing a change in backend logic. the code is treating it as unique per github org and not repository. I think we have move to a format:
  `${FILESERVER_URL}/${GITHUB_REPOSITORY}/${GITHUB_RUN_ID}/`

I'll need to deploy something that's temporarily backward compatible while I work through all the places we do uploads.

doanac: n/m - this works well. i like it.

doanac (approved):

.github/workflows/debos.yml
lool: (See Andy's note [above](https://github.com/qualcomm-linux/qcom-deb-images/pull/4#discussion_r2023853797))

PR #8, opened 2025-04-08
workflows: Fix typo in fileserver copy

lool:
Rename /builds to fileserver-builds for clarity and in case we need to expose
others directories.


lool:
This is a regression from https://github.com/qualcomm-linux/qcom-deb-images/pull/7

doanac (commented):
good catch. thanks

PR #17, opened 2025-04-15
Lava testing

mattface:
Building on PR #15 
This fixes the LAVA auto-login user/pass and the file paths to the LAVA job template files.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -102,9 +102,9 @@ jobs:
           BUILD_FILE_NAME=flash.tar.gz
           BUILD_DOWNLOAD_URL="${url}${BUILD_FILE_NAME}"
 
-          sed -i "s|{{BUILD_FILE_NAME}}|${BUILD_FILE_NAME}|g" "ci/lava/rb3gen2-vision-kit-boot.yaml"
-          sed -i "s|{{BUILD_DOWNLOAD_URL}}|${BUILD_DOWNLOAD_URL}|g" "ci/lava/rb3gen2-vision-kit-boot.yaml"
-          cat "ci/lava/rb3gen2-vision-kit-boot.yaml"
+          sed -i "s|{{BUILD_FILE_NAME}}|${BUILD_FILE_NAME}|g" "../ci/lava/rb3gen2-vision-kit-boot.yaml"
+          sed -i "s|{{BUILD_DOWNLOAD_URL}}|${BUILD_DOWNLOAD_URL}|g" "../ci/lava/rb3gen2-vision-kit-boot.yaml"
+          cat "../ci/lava/rb3gen2-vision-kit-boot.yaml"
 
       - name: Submit LAVA job
         timeout-minutes: 30
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -25,6 +25,9 @@ defaults:
 
 env:
   FILESERVER_URL: https://quic-yocto-fileserver-1029608027416.us-central1.run.app
+  # github runs are only unique per repository and may also be re-run;
+  # create an unique id with repository, run id, and run attempt
+  BUILD_ID: ${{ github.repository }}-${{ github.run_id }}-${{ github.run_attempt }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/debos.yml
```diff
@@ -95,4 +95,35 @@ jobs:
...
           curl -X POST -H 'Accept: text/event-stream' "${url}"
+      - name: Prepare LAVA test job
+        run: |
+          id="${GITHUB_REPOSITORY}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          url="${FILESERVER_URL}/${id}/"
+          BUILD_FILE_NAME=flash.tar.gz
+          BUILD_DOWNLOAD_URL="${url}${BUILD_FILE_NAME}"
+          LAVA_JOB_FILE="../ci/lava/rb3gen2-vision-kit-boot.yaml"
+
+          sed -i "s|{{BUILD_FILE_NAME}}|${BUILD_FILE_NAME}|g" ${LAVA_JOB_FILE}
```
lool, line 106: I find it weird to in-place sed source files, I would make a copy first, but I understand this is copy-pasted from another instance

I'd also quote `"${LAVA_JOB_FILE}"`

.github/workflows/debos.yml
```diff
@@ -95,4 +95,35 @@ jobs:
...
+
+      - name: Submit LAVA job
+        timeout-minutes: 30
+        uses: foundriesio/lava-action@v6
+        with:
+          lava_token: ${{ secrets.LAVATOKEN }}
+          lava_url: 'lava.infra.foundries.io'
+          job_definition: ${LAVA_JOB_FILE}
+          wait_for_job: true
+          fail_action_on_failure: false
```
lool, line 121: Mixed feelings on this; I guess it sucks to fail the whole build because we haven't managed to submit a test job to LAVA, but it's also synchronous (wait_for_job: true); makes me wonder if we should have this as a separate job as to be able to give just that part back

ci/lava/rb3gen2-vision-kit-boot.yaml
```diff
@@ -0,0 +1,71 @@
...
+        url: "{{BUILD_DOWNLOAD_URL}}"
+    postprocess:
+      docker:
+        image: ghcr.io/foundriesio/lava-lmp-sign:main
+        steps:
+        - export IMAGE_PATH=$PWD
+        - cp overlay*.tar.gz overlay.tar.gz
+        - echo "OVERLAY=overlay.tar.gz" >> $IMAGE_PATH/flash.settings
+        - echo "OVERLAY_PATH=/home/" >> $IMAGE_PATH/flash.settings
+        - echo "ROOTFS_IMAGE=disk-ufs.img2" >> $IMAGE_PATH/flash.settings
```
lool, line 16: I guess this is the whole rootfs amending logic; this goes a bit counter to the ambition to test unmodified images, but we don't have a replacement to inject the overlay right now, so we can as well start with this.

ci/lava/rb3gen2-vision-kit-boot.yaml
```diff
@@ -0,0 +1,71 @@
...
+      repository: https://github.com/linaro/test-definitions.git
+      parameters:
+        SKIP_INSTALL: "True"
+        TESTS: "pwd, uname -a, ip a"
+- command:
+    name: network_turn_on
+context:
+  lava_test_results_dir: /home/lava-%s
+  test_character_delay: 10
+device_type: qcs6490
```
lool, line 61: I guess this is where we select target devices in the lab; I believe we have one vision kit and multiple core kits; could we specifically point at the vision kit? right now, vision kit images should boot on core kit and vice-versa, but we will want to specialize this progressively, notably as the DT for vision kit lands and we can start testing one of the two camera sensors with the vision kit

lool (commented):
Woohoo! Here's a first review pass; let's also wait until we've figured the secrets/token issue and we get a first build passing a boot test :)

lool:
Moved to https://github.com/qualcomm-linux/qcom-deb-images/pull/36

Issue #24, opened 2025-04-24
Add adbd


lool:
Hi,

adbd has been requested in the reference image as to help with automation tasks. This would be used when plugging RBx boards to a developer workstation. Adb is seen as a quite portable (macOS, Windows, Linux) way to implement things like "give me a shell on that device" or to push/pull files from devices.

PR #26, opened 2025-04-25
debos: rootfs: Install adbd

lool:
After installation, adbd just comes up automatically on boot and can be
used to open shells or push/pull files.

Fixes: #24

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -28,6 +28,8 @@ actions:
     description: Install foundational packages
     recommends: true
     packages:
+      # Android Debug Bridge (daemon)
+      - adbd
       # bluetooth
       - bluez
       # vfat tools, notably fsck.fat for the ESP


lool:
Tested by installing adbd on an RB1 provisioned with a recent flat image from CI and rebooting; I can:
- open a shell with `adb shell`
- pull files with `adb pull /etc/passwd`

Note that this only works on RB1 if you set DIP switch 6 on "DIP_SW_1" to ON, which routes SoM's USB 3.0 to the type-C port instead of the built-in USB hub.

koenkooi (approved):

ricardosalveti:
> Note that this only works on RB1 if you set DIP switch 6 on "DIP_SW_1" to ON, which routes SoM's USB 3.0 to the type-C port instead of the built-in USB hub.

Do we need to have this documented somewhere?

ricardosalveti (approved):

lool:
> > Note that this only works on RB1 if you set DIP switch 6 on "DIP_SW_1" to ON, which routes SoM's USB 3.0 to the type-C port instead of the built-in USB hub.
> 
> Do we need to have this documented somewhere?

Yes, perhaps in something like a wiki :D

PR #28, opened 2025-04-28
workflows: debos: Generate SBOM of rootfs with syft

lool:
Depends on cleanup work in https://github.com/qualcomm-linux/qcom-deb-images/pull/29

Add a new job to generate SBOMs with Syft

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -90,19 +90,19 @@ jobs:
         run: |
           set -ux
           # create a directory for the current run
-          BUILD_DIR="./uploads"
-          mkdir -vp "${BUILD_DIR}"
+          dir="debos-artifacts"
+          mkdir -v "${dir}"
           # copy output files
-          cp -av rootfs.tar.gz "${BUILD_DIR}"
-          cp -av dtbs.tar.gz "${BUILD_DIR}"
-          cp -av disk-ufs.img.gz "${BUILD_DIR}"
-          cp -av disk-sdcard.img.gz "${BUILD_DIR}"
+          cp -av rootfs.tar.gz "${dir}"
+          cp -av dtbs.tar.gz "${dir}"
+          cp -av disk-ufs.img.gz "${dir}"
+          cp -av disk-sdcard.img.gz "${dir}"
           # TODO: separate flash_* directories between UFS and eMMC
-          tar -cvf "${BUILD_DIR}"/flash-ufs.tar.gz \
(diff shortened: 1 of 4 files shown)


lool:
Ok, this seems to do a good initial job; I'm not convinced by the value of the default set of catalogers, perhaps we should only keep the deb/dpkg catalogers active.

I checked the syft-json report, and it has:
```
   "licenses": [
    {
     "value": "Apache-2.0",
     "spdxExpression": "Apache-2.0",
     "type": "declared",
     "urls": [],
     "locations": [
      {
       "path": "/usr/share/doc/adbd/copyright",
       "accessPath": "/usr/share/doc/adbd/copyright"
[...]
   "purl": "pkg:deb/debian/adbd@34.0.5-12?arch=arm64&distro=debian&upstream=android-platform-tools",
   "metadataType": "dpkg-db-entry",
   "metadata": {
    "package": "adbd",
    "source": "android-platform-tools",
    "version": "34.0.5-12",
```

So I'm confident we could have a table of concluded licenses by source package, either using the syft csv/go templating or doing some light post-processing of the JSON.

lool:
@ricardosalveti I think we could start with landing this, and then improve from there; what do you think?

ricardosalveti:
> @ricardosalveti I think we could start with landing this, and then improve from there; what do you think?

yeah, we can probably use deb/dpkg catalogers and merge the initial sbom generation even if that is package based at this point.

probably better to split the cleanup work in another pr.

next step would indeed be to merge the package based sbom to become a source package sbom.

lool:
Rebasing on tip after the wave of CI changes

lool:
Using syft's `--select-catalogers debian` picks up the dpkg and deb catalogers, which seems to be what we want here.

.github/workflows/debos.yml
```diff
@@ -112,3 +117,48 @@ jobs:
...
+        run: |
+          set -ux
+          apt -y install curl
+          curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh
+
+      - name: Generate SBOMs with Syft
+        run: |
+          set -ux
+          # TODO should probably restrict catalogers a bit as the rootfs is
+          # built entirely from deb packages
```
ricardosalveti, line 104: We can probably drop this comment since we're using the debian cataloger now.

lool:
Noticed that the sha256sum of the copyrights weren't properly computed; adding a flag to the script for the path to the rootfs

PR #30, opened 2025-04-29
workflows: linux: Change dir after reading configs

lool:
Config fragments might be passed as relative pathnames, so defer changing
directory until these have been processed.


lool:
See https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/14703944641/job/41259442725 for a failed weekly build

PR #32, opened 2025-04-30
debos: flash: v75 of boot binaries for QCM6490

lool:
From Qualcomm Linux 1.4

Known changes:
- Boot KPI optimizations
- Update memmap holes dynamically as conventional memory
- Minor bug fixes
- Supporting 32-bit OEM ID

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -24,7 +24,7 @@ actions:
 
   - action: download
     description: Download QCM6490 boot binaries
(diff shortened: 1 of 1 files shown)


lool:
I've tested the CI images with the updated firmware, and these work mostly as well as the previous ones.

Differences I've noticed:
- the boot didn't seem to happen automatically after flashing
- there's some DDR training on boot making the boot slightly slower (not sure if it's every boot)
- different output from early bootloaders around DDR init, seems as designed

There's currently a lot of UFS debug spew in Debian images on RB3, so it's hard to state how well the hw is enabled at the moment.

I think this boot binaries are ok to update though.

ricardosalveti (approved):

PR #34, opened 2025-04-30
Linux deb: Fix config fragment usage

lool:
Fix a typo testing whether there is a local config fragment. This is
nasty in that it won't fail even under `set -e` as it's under an if
statement.


lool:
I'm going to self-approve this to reroll a kernel and unbreak the image builds – currently all images fail to boot because the kernel is built without CONFIG_EFI_ZBOOT.

PR #44, opened 2025-05-14
ci: Move to new artifact upload action

doanac:


ricardosalveti (approved):
Nice

.github/workflows/debos.yml
```diff
@@ -35,6 +35,12 @@ jobs:
         - /efs/qli/metaqcom/gh-runners/quic-yocto/downloads:/fileserver-downloads
       options: --privileged
     steps:
+      - name: TMP-DEBUG
+        uses: qualcomm-linux/upload-private-artifact-action@v1
+        with:
+          path: /tmp/${BUILD_ID}
+          build_id: "${{ github.repository }}-${{ github.run_id }}-${{ github.run_attempt }}"
+          github_token: "${{ secrets.GITHUB_TOKEN }}"
```
lool, line 43: I'm surprised you need to pass this specifically in the yaml; isn't GITHUB_TOKEN provided to actions by default?

.github/workflows/debos.yml
```diff
@@ -35,6 +35,12 @@ jobs:
         - /efs/qli/metaqcom/gh-runners/quic-yocto/downloads:/fileserver-downloads
       options: --privileged
     steps:
+      - name: TMP-DEBUG
+        uses: qualcomm-linux/upload-private-artifact-action@v1
+        with:
+          path: /tmp/${BUILD_ID}
+          build_id: "${{ github.repository }}-${{ github.run_id }}-${{ github.run_attempt }}"
```
lool, line 42: This is more some thoughts on the API of this action, but I wonder if we should keep the concept of a build_id local to the repositories that needs one.

As an user of such an action, I would expect it to provide similar behavior as actions/upload-artifact. In particular, I expect that the artifacts uploaded from one repo are stored in a repo specific collection of artifacts.

When using upload-artifact, most parameters are optional. Perhaps our upload-private-artifact could provide similar behavior and there would be a download action to easily retrieve a particular artifact or the latest artifact with some name (like we're doing in the debos recipe to fetch u-boot/linux builds prior to building an image).

.github/workflows/debos.yml
```diff
@@ -35,6 +35,12 @@ jobs:
         - /efs/qli/metaqcom/gh-runners/quic-yocto/downloads:/fileserver-downloads
       options: --privileged
     steps:
+      - name: TMP-DEBUG
```
lool, line 38: I'm guessing this is going away in the final version?

lool (commented):
I guess this is a draft PR as I see some DEBUG entries in it still?

.github/workflows/debos.yml
doanac: Its a security thing. GitHub doesn't want you to accidentally share the token with a workflow you haven't explicitly granted permissions to.

lool: Is is possible to implement an action like the github upload-artifact action, without passing the token? Does that require some lengthy certification process with github or something?

.github/workflows/build-daily.yml
```diff
@@ -9,7 +9,7 @@ on:
 
 permissions:
   contents: read
-  security-events: read  # This is required to handle authentication to our artifact publishing API
+  id-token: write  # This is required to handle authentication to our artifact publishing API
```
lool, line 12: I read https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect and IIUC, the `id-token: write` is always required when we want to generate OIDC JWT tokens.

These tokens allow us to manage trust/permissions between GitHub and our public cloud infrastructure (the fileserver) and express things like "this upload-private-artifact request comes with a token that github guarantees as coming from a workflow run for this org/repo".

I guess it's hard to review the security architecture from the context of this PR, there's probably a design doc on the public cloud infra side where we'd keep track of that. The way I think about it, this particular permission is letting a github workflow join in on doing more advanced OIDC based things and working with public cloud permissions.

Since this is for the context of this particular repository, I would propose:
- that we add a comment that `id-token: write` is "needed for the upload-private-artifact action"
- (if you're tempted, we could add a comment that `contents: read` is only needed for `actions/checkout@v4` :-))

I guess there are rules on the public cloud side for all workflow runes or from the ones triggered from third party repository to not be allowed to do certain things? I wonder if we could capture them somewhere, perhaps next to the upload-private-artifacts action?

.github/workflows/build-on-pr.yml
```diff
@@ -5,7 +5,7 @@ on:
 
 permissions:
   contents: read
-  security-events: read  # This is required to handle authentication to our artifact publishing API
+  id-token: write # This is required to handle authentication to our artifact publishing API
```
lool, line 8: This is the only tricky one to review, but here the triggering event is a pull_request, and the workflow definition being used is the one in the main branch.

.github/workflows/debos.yml
```diff
@@ -118,7 +113,9 @@ jobs:
               disk-sdcard.img1 \
               disk-sdcard.img2 \
               flash_rb1*
-          # instruct fileserver to publish this directory
-          export URL="${FILESERVER_URL}/${BUILD_ID}/"
-          .github/workflows/publish_artifacts.py
-          echo Image available at: ${URL}
+
+      - name: Upload artifacts
```
lool, line 117: Maybe say private somewhere

lool (commented):
I think this is really looking really nice now, from the perspective of this repository!

.github/workflows/build-daily.yml
doanac: This has turned into a rabbit hole of a task. This approach was only working when you weren't using forked repos. I've figured out another way that doesn't require any "permission" stuff. It requires a bit more work, but I think I can have that sorted out by Monday to make this work pretty cleanly

doanac:
Okay - i've cleaned up the action and backend code in a way to make this work fairly cleanly

lool:
I liked all the removals in the latest version :)

doanac:
I think this is ready if you are satisfied. It turned into way more backend work than I was ever wanting to do for this but it should make the transition work for other repos go more smoothly as we get them on the new runners.

lool (approved):

PR #49, opened 2025-05-23
workflows: fix dependency name in build-on-push

mwasilew:
Test job depends on build, but there was a mistake in the dependency name. This patches fixes the dependency to proper name.


lool (approved):
Looks good

PR #57, opened 2025-05-26
ci: Fix Debian user/password in LAVA template

mwasilew:
Update auto_login section in LAVA template to match debian image settings.


lool:
(Tests ran but failed because template definition is in main branch)

PR #63, opened 2025-05-28
workflows: fix name of build tarball for RB1

mwasilew:
RB1 uses flash-emmc.tar.gz. So far test.yml had flash-ufs.tar.gz hardcoded. This patch adds support for chaning the tarball name depending on the board type.


lool:
Looks good to me, thanks!

PR #68, opened 2025-06-05
workflows: build-overlay-deb: Use workflow input

lool:
CONFIG_PATH was used during development of this workflow, but we want to
use the workflow_dispatch input.


lool:
@basak-qcom If you don't mind reviewing this followup :)

lool:
I'm going to go ahead and merge this as we need to move forward with the overlay

PR #70, opened 2025-06-10
ci: switch to AWS container registry

sampra2025:
As per AWS environment its recommended to use AWS Hosted docker hub instead of Public docker hub such as "docker.io", so updating the container registry.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -27,7 +27,7 @@ jobs:
       url: ${{ steps.upload_artifacts.outputs.url }}
     runs-on: [self-hosted, qcom-u2404, arm64]
     container:
-      image: debian:trixie
+      image: public.ecr.aws/debian/debian:trixie
       volumes:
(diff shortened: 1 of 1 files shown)


lool:
@sampra2025 Not sure you saw my comment above?

lool:
@sampra2025 and I connected out of band; he noted: "Yes I think we can close this PR. I think Andy has already implement such changes in some other PR, so its no more required. " (Cc: @doanac)

doanac:
I haven't implemented anything that would negate this PR. This PR is just a small performance optimization - The workflow is running in AWS, so pulling from an AWS container registry will be a little faster (and probably cheaper).

lool:
Hi @sampra2025 and @doanac, thanks for your input, I've sent a new pull request https://github.com/qualcomm-linux/qcom-deb-images/pull/121 implementing this idea, using the rationale provided in your conversation, and updated to cover the other workflows that use docker images. Thanks for proposing this!

PR #71, opened 2025-06-11
debos: rootfs: install OpenCL and Vulkan tools

lumag:
Install OpenCL info package, Vulkan tools and Mesa OpenCL ICD (driver).

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -112,9 +112,11 @@ actions:
     recommends: true
     packages:
       - alsa-utils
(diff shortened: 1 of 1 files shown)


debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -112,10 +112,13 @@ actions:
     recommends: true
     packages:
       - alsa-utils
+      - clinfo
       - device-tree-compiler
       - i2c-tools
       - locales
+      - mesa-opencl-icd
       - mesa-utils-bin
+      - vulkan-tools
```
ndechesne, line 121: sorted list?

PR #72, opened 2025-06-12
Add mesa snapshot overlay

lumag:
In order to evaluate upstream Mesa on the RB1 board, package the snapshot of the Mesa development tree. This package will go away once Mesa 25.2 gets released and enters Debian experimental or unstable.

### DIFF
--- a/overlay-debs/mesa-snapshot/mesa-snapshot.sh
+++ b/overlay-debs/mesa-snapshot/mesa-snapshot.sh
@@ -39,6 +39,8 @@ mesa (${version}-0qcom1) testing; urgency=medium
     - debian/patches/path_max.diff: refresh
     - d/rules: dropped removal of mme_{fermi,tu104}_sim_hw_test
     - pull in NVK dependency, librust-rustc-hash-2-dev
+    - opt-in and enable several OpenCL drivers by default: asahi, freedreno and
+      radeonsi
   * Drop support for features removed upstream:
     - XA tracker, dropped packages: libxatracker2, libxatracker-dev.
     - D3D9 tracker, dropped packages libd3dadapter9-mesa,
--- a/overlay-debs/mesa-snapshot/mesa-snapshot.sh
+++ b/overlay-debs/mesa-snapshot/mesa-snapshot.sh
@@ -7,6 +7,7 @@ COMMIT=${COMMIT:-origin/main}
 
 set -e
 
+# shellcheck disable=SC2153 # DSC_FILE is guaranteed to be defined by build-deb.py
 DEB_FILE=${DSC_FILE%%.dsc}.debian.tar.xz
 ORIG_FILE=$(echo ${DSC_FILE} | sed -e 's/-.*/.orig.tar.xz/')
 
--- a/overlay-debs/mesa-snapshot/mesa-snapshot.sh
+++ b/overlay-debs/mesa-snapshot/mesa-snapshot.sh
@@ -47,6 +47,9 @@ mesa (${version}-0qcom1) testing; urgency=medium
   * Backport to trixie:
     - Build with LLVM 19 since LLVM 20 is not available in trixie.
     - d/control: regenerate
+  * Feature patches:
+    - d/p/35316.patch: freedreno: Add sampling support for RGB/BGR
+      24-bit component texture formats.
 
  -- Dmitry Baryshkov <dmitry.baryshkov@oss.qualcomm.com>  Wed, 11 Jun 2025 14:58:50 +0300
 
--- a/overlay-debs/mesa-snapshot/mesa-snapshot.sh
+++ b/overlay-debs/mesa-snapshot/mesa-snapshot.sh
@@ -1,0 +1,66 @@
+#! /bin/sh
+#
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+
+COMMIT=${COMMIT:-origin/main}
+
+set -e
+
+DEB_FILE=${DSC_FILE%%.dsc}.debian.tar.xz
+ORIG_FILE=$(echo ${DSC_FILE} | sed -e 's/-.*/.orig.tar.xz/')
+
+rm -rf mesa
+git clone --depth 1 https://gitlab.freedesktop.org/mesa/mesa
+
+cd mesa
+git fetch --depth 1 origin ${COMMIT##origin/}
+
+date=$(git log -1 --format=%cd --date=format:%Y%m%d ${COMMIT})
+subject=$(git log -1 --format="%h (\"%s\")" ${COMMIT})
+version=25.2.0~git${date}
+
+rm -rf ../mesa-${version}
+mkdir ../mesa-${version}
+git archive --format=tar HEAD | tar x -C ../mesa-${version}
+
+cd ../mesa-${version}
+
+rm -rf debian
+tar xJf ${DEB_FILE}
+
+cat >> debian/changelog.tmp << EOF
+mesa (${version}-0qcom1) testing; urgency=medium
+
+  * Build git version from ${date}, commit ${subject}
+    - d/libegl-mesa0.symbols: include GL interop symbols into, exported by
+      libEGL_mesa.so.0
+    - d/p/etnaviv-add-support-for-texelfetch.patch: drop, applied upstream
+    - debian/patches/path_max.diff: refresh
+    - d/rules: dropped removal of mme_{fermi,tu104}_sim_hw_test
+    - pull in NVK dependency, librust-rustc-hash-2-dev
+  * Drop support for features removed upstream:
+    - XA tracker, dropped packages: libxatracker2, libxatracker-dev.
+    - D3D9 tracker, dropped packages libd3dadapter9-mesa,
+      libd3dadapter9-mesa-dev.
+    - Clover, removed clover files from d/mesa-opencl-icd.install.
+  * Backport to trixie:
+    - Build with LLVM 19 since LLVM 20 is not available in trixie.
+    - d/control: regenerate
+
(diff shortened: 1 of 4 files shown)


basak-qcom (changes requested):
[Comment only] Ideally we'd have a better mechanism to do all of this, such as packaging repositories in git that can derive from upstream and Debian packaging commits. But in the meantime to make progress we already have overlay-debs debdiffs as a compromise. Adding a generic script mechanism seems like a clean way of approaching this. Even though it's more creep towards the less ideal, I can't think of a better general approach.

[No change needed] We have a service dependency on gitlab.freedesktop.org being introduced here, whereas before we depended on Debian apt archives only I think. Again, given this is a stop-gap solution, I guess that's OK. Perhaps we should start noting these somewhere though.

[For discussion before landing] The `debian/` directory and packaging inside gives me some concern. It's being copied from Debian experimental I think, and then being modified? Since our goal is to minimise any "delta" to upstream over time, ideally we'd store the diff rather than a copy. Notably this approach is the one we take with the overlay-debs debdiff mechanism, which is what this is building upon. I wouldn't do that for the entire upstream snapshot, but perhaps we can do that for the packaging, supplying a debdiff to apply over the source package once it has been constructed from the upstream git commit and the debian directory from Debian experimental? Otherwise, it took me some effort to fetch and generate a diff against Debian experimental to review, and then I'm finding it difficult to understand why you've changed some aspects of the packaging against Debian experimental. For someone else to understand that is key to allowing that difference to be eliminated in the future. What do you think about storing a diff against experimental instead, and then generating that dynamically as part of the script?

[Changes requested] Please could you link directly to the upstream PRs or commits that you are cherry-picking from the patch files themselves? For example there's [dep3](https://dep-team.pages.debian.net/deps/dep3/) which describes what metadata should go into a patch in `debian/patches`. In this case of squashing multiple commits together, I think that's fine in which case maybe just dump the headers at the top? I don't know of any tool that reads these; it's just really useful when rebasing in the future, so I'd like to make it a standard that the information is available. I think this should be in the source tree itself rather than just in commits.

scripts/build-deb.py
```diff
@@ -37,30 +37,46 @@
...
-    print("✅ Checksum of original source package matched.")
+    script = config.get('script')
+    if script:
+        if not os.path.isabs(script):
+            config_dir = os.path.dirname(args.config)
+            script = os.path.abspath(os.path.join(config_dir, script))
+
+        if 'env' in config:
+            for k, v in config['env'].items():
+                os.putenv(k, v)
```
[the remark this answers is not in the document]
lumag: No, that's not possible. This way you will loose other important environment variables like `PATH`. Not to mention that the current build-with-debdiff option doesn't sanitise the environment, passing all variables from the caller to sbuild.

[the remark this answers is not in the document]
basak-qcom: > No, that's not possible. This way you will loose other important environment variables like `PATH`.

My sample code specifically preserves all other environment variables. It just ensures that the additional variables are supplied only to the script, and not to other subsequent calls.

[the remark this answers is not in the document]
lumag: Ack, I see.

lumag:
@lool @basak-qcom reworked the PR

Corresponding packages were uploaded to [OBS](https://build.opensuse.org/package/show/home:lumag_lumag/mesa-git)

scripts/build-deb.py
```diff
@@ -59,8 +59,24 @@
...
-    # Unpack the source package
-    subprocess.run(['dpkg-source', '-x', dsc_file], cwd=temp_dir, check=True)
+    script = config.get('script')
+    if script:
+        if not os.path.isabs(script):
+            config_dir = os.path.dirname(args.config)
+            script = os.path.abspath(os.path.join(config_dir, script))
+
+        env = os.environ.copy()
+        env['DSC_FILE'] = dsc_file
```
lool, line 69: I'd prefer if we passed the whole yaml file to the script, but I know it's not fun to parse yaml from shell; that's actual actually why I wrote build-deb in python, but now we're extending it to call a shell script :)

I guess this is a temporary solution and we need to get rid of this debdiff juggling logic ASAP, but this will do for now.

overlay-debs/mesa-snapshot/mesa-snapshot.sh
```diff
@@ -0,0 +1,52 @@
...
+mkdir ../mesa-${version}
+git archive --format=tar HEAD | tar x -C ../mesa-${version}
+
+cd ../mesa-${version}
+
+rm -rf debian
+tar xJf ${DEB_FILE}
+
+cat >> debian/changelog.tmp << EOF
+mesa (${version}-0) experimental; urgency=medium
```
lool, line 33: would you mind changing the version to 0qcom1 so that all packages with qcom changes have a qcom version?

lool (changes requested):
Hey, thanks for implementing this, and sorry for the delay in reviewing on my side; I've left mostly minor comments but please could you:
- change the Debian revision to 0qcom1
- add some notes on how you created the debdiff, either in the commit message or in debian/changelog inside the debdiff
- pass this new script through the shellcheck workflow (.github/workflows/static-checks.yml only runs against ci/*.sh).

Thanks!

scripts/build-deb.py
```diff
@@ -59,8 +59,24 @@
         )
     print("✅ Checksum of original source package matched.")
 
-    # Unpack the source package
-    subprocess.run(['dpkg-source', '-x', dsc_file], cwd=temp_dir, check=True)
+    script = config.get('script')
```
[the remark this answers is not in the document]
basak: Maybe `source_unpack_script`? Since it replaces the `dpkg-source -x` call, not the `.dsc` file download.

lumag:
@basak-qcom too many words for a package that is going to be obsoleted in July or August. Could you please make your comments easier to comprehend?

basak-qcom:
@lumag I pushed 76162c86d92842b09c0762637997ad1dc8a36186 directly to your branch to fix some shellcheck warnings. I'm confident that's unobjectionable. I trust that's appropriate process/etiquette here?

Three more changes to come: one remaining shellcheck issue, I'll add a README, and I'll rebase onto main in order to pick up the shellcheck CI changes. Please let me know if I should do this differently.

overlay-debs/mesa-snapshot/mesa-snapshot.sh
```diff
@@ -7,6 +7,11 @@ COMMIT=${COMMIT:-origin/main}
 
 set -e
 
+# Reported by shellcheck: SC2153 (info): Possible misspelling: DSC_FILE may not
+# be assigned. Did you mean DEB_FILE?
+#
+# DSC_FILE is guaranteed to be defined as part of the call API from build-deb.py
+# shellcheck disable=SC2153
```
lumag, line 14: NAK, don't add extra noise.

basak-qcom: Just removing it will cause it to fail CI. Please clarify how you want me to fix it instead.

overlay-debs/mesa-snapshot/README.md
```diff
@@ -0,0 +1,25 @@
...
+      24-bit component texture formats.
+    - d/p/etnaviv-add-support-for-texelfetch.patch: drop.
+    - debian/patches/path_max.diff: refresh for context only.
+    - debian/rules:
+      + enable relevant OpenCL drivers by default. Opt-in and use the
+        gallium-rusticl-enable-drivers in order to enable RustiCL drivers which
+        can be enabled by default: asahi, freedreno and radeonsi.
+      + Disable gallium-xa and gallium-opencl=icd.
+      + No longer need to remove mme_{fermi,tu104}_sim_hw_test
+    - debian/control regenerated from debian/control.in.
```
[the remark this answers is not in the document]
basak-qcom: I agree that would be better, but you are generating the changelog - parameterised - from your script. I didn't put this directly in there since it'll be wrong as soon as the parameters change. I can either keep it in the README.md as-is, or remove the changelog generation from the script and put a changelog entry via the debdiff instead. Which do you prefer?

overlay-debs/mesa-snapshot/mesa-snapshot.sh
```diff
@@ -8,26 +8,26 @@ COMMIT=${COMMIT:-origin/main}
...
 DEB_FILE=${DSC_FILE%%.dsc}.debian.tar.xz
-ORIG_FILE=$(echo ${DSC_FILE} | sed -e 's/-.*/.orig.tar.xz/')
+ORIG_FILE=$(echo "${DSC_FILE}" | sed -e 's/-.*/.orig.tar.xz/')
 
 rm -rf mesa
 git clone --depth 1 https://gitlab.freedesktop.org/mesa/mesa
 
 cd mesa
-git fetch --depth 1 origin ${COMMIT##origin/}
+git fetch --depth 1 origin "${COMMIT##origin/}"
```
lumag, line 22: We don't need to quote all the variables. Only the variables that might have whitespaces or might be empty need to be quoted.

lool: Yup, shellcheck doesn't know that, so it asks to quote everything by default, but we can add an override to tell it to ignore this particular line; I think it's cheaper to add the double quotes personally, even if it's a bit more verbose

overlay-debs/mesa-snapshot/mesa-snapshot.sh
lool: we can probably do this on one line, something like:
```
# shellcheck disable=SC2153 # DSC_FILE is guaranteed to be defined by build-deb.py
```

or we could set DSC_FILE if unset at the top of the file, with a comment saying it should be set by build-dep.py

lool:
@lumag I kicked a build, but it failed due to git not being installed:
https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/15966677844/job/45028450825

I'd suggest either adding a sudo apt install git to the shell script, or to add an entry in the yaml with list of dependencies to install prior to running the script.

PR #75, opened 2025-06-16
ci: fix LAVA job templates with password change prompt

mwasilew:
Update LAVA job templates for the case when password has to be changed at the first login attempt. This patch should only be merged after #74


lool (approved):
Looks good to me; thanks! I presume it's safe to land ahead of the image changes

mwasilew:
Unfortunately if you merge it before #74 all test jobs will fail.

basak-qcom:
@mwasilew thank you for the fix! I've picked your commit into my branch for #74 so that they can be tested together. Otherwise I can't demonstrate CI passing before landing that one. So I'll close this PR - consider it merged via #74 instead when that lands.

PR #77, opened 2025-06-19
Build and test images for various kernels

lool:
This PR will properly test our regular (currently weekly) mainline
kernel debs, and after switching the regular image builds back to Debian
kernels (from experimental as the trixie one has too many issues on RB1
and RB3Gen2) it uses a new APT repository with a linux-qcom kernel.

Since we currently lack a kernel metapackage, this is hardcoding a versioned kernel package name.


lool:
I can confirm this is now booting on RB1!

PR #83, opened 2025-06-25
workflows: test: Fix BUILD_URL env

lool:
BUILD_URL was not properly defined for a single step, and BUILD_URL is
used across multiple steps of that job, so define it properly and at the
job level.


mwasilew (approved):

mynameistechno:
👍🏼

PR #85, opened 2025-06-29
debos: flash: follow qcom-ptool changes

lumag:
qcom-ptool has renamed platform subdirs for several platforms, which broke one of the stages in this repo. Fix the script by following those changes.


lool:
Thanks! Will merge this now

After seeing the daily builds failing, I independently created https://github.com/qualcomm-linux/qcom-deb-images/pull/86 with more updates notably to comments etc. to be more consistent between RB1 and RB3G2.

PR #86, opened 2025-06-30
debos: flash: Update to use platform names

lool:
As qcom-ptool renamed partition files to follow platform / machine
names, follow suit and update code and comments to use consistent names
for silicon family, platform and board.


lumag:
I'd suggest squashing those two committs

PR #88, opened 2025-06-30
ci: lava: fix flashing path for tarballs

mwasilew:
After change in platform names caused by introduction of ptool, LAVA jobs fail because of wrong path in the tarball. This patch adjusts the name of the directory so the flashing can now succeed.


mwasilew:
This PR fixes the issue found in #86

PR #98, opened 2025-07-11
Misc default packages/groups improvements

lool:
Changes to default groups and packages

- **debos: rootfs: Add debian to audio and render**
- **debos: rootfs: Add Docker**
- **debos: rootfs: Install v4l-utils for media-ctl**

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -86,9 +86,11 @@ actions:
       set -eux
       # pre-emptively create a sudo group if sudo isn't installed
       getent group sudo >/dev/null 2>&1 || groupadd --system sudo
-      # some of these groups are only needed for desktop scenarios
+      # some useful groups for desktop scenarios, but also to run payloads
+      # from the serial console, over SSH, or in containers - where the desktop
+      # session has not updated ACLs to the device nodes
       useradd --create-home --shell /bin/bash --user-group \
-          --groups adm,video,users,sudo debian
+          --groups adm,audio,render,sudo,users,video debian
       # password must be changed on first login; set it to "debian"
       chage --lastday 0 debian
       echo debian:debian | chpasswd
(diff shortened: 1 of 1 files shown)


lool:
I was expecting to see the test results in the PR test run, but they won't: tests are always taken from the main branch, not from PR.

So perhaps I should really separate the LAVA CI updates from the rest, as they only make sense separately anyway!

Also, RB1 tests are currently broken.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -143,6 +144,13 @@ actions:
       - vulkan-tools
       - wget
 
+  - action: run
+    description: Add default user to docker group
+    chroot: true
+    command: |
+      set -eux
+      usermod -a -G docker debian
```
basak-qcom, line 154: Could you move the new user creation to after all the apt installations, and then you could add `docker` to the `--groups` option against `useradd`? Then all the group memberships could be one place rather than having to different ways of doing it.

From a quick glance at the recipe I think this should work? I don't see any dependencies in our recipe of the `debian` user existing, and generally `apt` shouldn't need the default user to exist yet, I don't think?

basak-qcom (changes requested):
Looks good but one question inline.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: I considered this, but I did in a separate section to keep the Docker related things together: there's a dependency on the docker package being installed before that group exists and can be used with useradd. The other groups from the useradd invocation are standard groups from base-files / standard Debian groups, so they are guaranteed to be there. So with the above, it's easy to spot where docker is being added/removed. That said, this block assumes the name of the debian user.

So I guess it's a tradeoff, I don't mind either way. With the above clarification, do you still prefer a single useradd call with all groups?

basak-qcom (approved):

PR #100, opened 2025-07-18
debos: rootfs: Change Firefox to Chromium

lool:
This is for two main reasons; Firefox is pulling libvpx while Chromium
is pulling openh264, the latter is somehow less problematic; and
Chromium is generally more common and less risky in terms of website
support.


lool:
> NB: there's a report of a Chromium regression in ANGLE on some Qualcomm targets in trixie, but we already identified a mesa patch to workaround/address the issue, so I'm going forward and landing this.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1109711

ANGLE bug: https://issues.angleproject.org/issues/431097618
(via Rob Clark and Dmitry Baryshkov – thanks!)

PR #102, opened 2025-07-21
Add more smoke tests from lava-definitions

lool:
ci: lava: Run wlan-smoke test on all targets

### DIFF
--- a/ci/lava/qcs6490-rb3gen2-core-kit/boot.yaml
+++ b/ci/lava/qcs6490-rb3gen2-core-kit/boot.yaml
@@ -53,6 +53,12 @@ actions:
       parameters:
         SKIP_INSTALL: "True"
         TESTS: "pwd, uname -a, ip a"
+    - from: git
+      name: "wlan-smoke-test"
+      path: automated/linux/wlan-smoke/wlan-smoke.yaml
+      repository: https://github.com/linaro/test-definitions.git
+      parameters:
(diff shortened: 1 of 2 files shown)


lool:
On RB1:
https://lava.infra.foundries.io/scheduler/job/87743

- same unhappy media-ctl issue
```INFO: Running media-ctl test...
Failed to enumerate /dev/media0 (-2)
media-ctl fail
```
- 1x usb smoke test failure
- 2x alsa smoke test failures
- wlan tests pass

So I'll enable wlan tests for both rb3 and rb1

lool:
Updated to only add wlan smoke tests; can be used on both targets; other smoke tests will require fixes to the enablement and/or to the tests first

ci/lava/qcs6490-rb3gen2-core-kit/boot.yaml
```diff
@@ -53,6 +53,12 @@ actions:
       parameters:
         SKIP_INSTALL: "True"
         TESTS: "pwd, uname -a, ip a"
+    - from: git
+      name: "wlan-smoke-test"
+      path: automated/linux/wlan-smoke/wlan-smoke.yaml
+      repository: https://github.com/linaro/test-definitions.git
+      parameters:
```
mattface, line 60: Does the wifi device get enumerated as 'wlan0' ?
If not I believe we would need to override it with the DEVICE parameter here

Issue #106, opened 2025-07-23
Update to QLI 95 boot binaries


lool:
We need to update our flat image recipe to the latest boot binaries v95.

This is similar to:
https://github.com/qualcomm-linux/meta-qcom/pull/886

This should bring at least a bug fix to QCM6490 (RB3Gen2).

PR #107, opened 2025-07-24
Update to QLI 95 boot binaries

basak-qcom:
This is similar to qualcomm-linux/meta-qcom#886 and should bring at least one bug fix for QCM6490 (RB3Gen2).

Fixes: #106


basak-qcom:
Thank you for the tip! This is ready to land now I think, subject to review.

lool:
Test failures were from unrelated initial rb1 run

PR #110, opened 2025-07-28
workflows: linux: Allow writing test results

lool:
test.yml requires checks write, so provide that.

GitHub is now smarter about not even starting the workflow is permissions aren't met.


lool:
Seems to make GH's checks happy at least :)
https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/16564966685/job/46842757588

PR #111, opened 2025-07-28
workflows: debos: Drop concurrency

lool:
This is duplicating concurrency already defined in workflows including
debos.yml, resulting in GitHub warnings such as "Canceling since a
deadlock was detected for concurrency group: 'Build Linux kernel deb and
debos image-refs/heads/main' between a top level workflow and
'debos-mainline-linux'".


basak-qcom (approved):
I don't see any issue with making this change.

PR #115, opened 2025-07-29
overlay-debs: Drop rmtfs

lool:
Fix is in trixie.


basak-qcom (approved):
I was also going to propose this this morning :-)

PR #119, opened 2025-08-04
kernel-configs: Add config fragment for IMSDK

lool:
Many GStreamer plugins in the Qualcomm IMSDK use DMABUF userspace heaps,
so provide a kernel config fragment to turn this on and use it in README
and GitHub workflows.


kernel-configs/qcom-imsdk.config
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
```
ricardosalveti, line 2: Don't think we need copyright statement for the config fragments, but fine to have it.

lool: Dropped from this new config, and also added a commit to drop it from the other config

PR #122, opened 2025-08-07
debos: flash: Add contents.xml for Axiom on RB3

lool:


lool:
Inspired by https://github.com/qualcomm-linux/meta-qcom/pull/992 but still missing a way to deploy this to Axiom

PR #126, opened 2025-08-27
ci: Escape kernel package name for APT

lool:


basak-qcom (approved):
As discussed I think we should fix this properly by building kernel debs without a `+` suffix, but I appreciate we're limited on time so this will do for now. It is a bit of a weird hack though, since the variable is now no longer a package name but an instruction to apt.

PR #127, opened 2025-08-27
Local APT repository

lool:
WIP – don't merge yet

- **feat: debos: Option for local APT repository**
- **ci: debos: Switch to local APT repository**
- **WIP: Force local testing**


lool:
This run is representative of how this would be used for the mainline kernel flow:
https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/17275458248/job/49030900617

lool:
This is likely to help with the current failures we're seeing with mainline kernel debs:
https://github.com/qualcomm-linux/qcom-deb-images/actions/workflows/linux.yml

It seems these are due to lack of space in the image, and this is a situation where we install two kernels (Debian one and mainline one) in the rootfs; with this approach, only one kernel should be pulled.

PR #131, opened 2025-08-28
mesa-unstable: fix the debdiff

lumag:
I've updated the debdiff manually, but I didn't update the patch header, which results in the failure to apply the patch.

Fix the patch header in the debdiff.


lool (approved):

lool:
Published to trixie-overlay-staging

lool:
Tested on RB1 and promoted to trixie-overlay

PR #133, opened 2025-09-02
[DO NOT MERGE] Validate Debian workflow on QLI-AMI

sampra2025:
AWS runner will be setup with newly crated QLI-AMI and GH workflow needs validation for end to end testing, before we deploy this AMI in production.

For validation, I have disable the CRON runs and as well upload artifacts to GCP.


lool:
FYI, I believe draft PRs will run your CI just the same, but won't get a chance to be merged

PR #142, opened 2025-09-10
Generate contents.xml flash files for Axiom

lool:
Generate contents.xml flash files for Axiom:
- **refactor(debos/flash): Use _ptool as workdir**
- **refactor(debos): Patch partitions.conf instead of .xml**
- **feat(debos)!: Drop now unused disk_image_type**
- **feat(debos): Generate contents.xml**


lool:
The tests actually pass, it's just that I had to give back the RB1 boot test due to https://github.com/foundriesio/linux-qcom/issues/2

Issue #37, opened 2025-04-30
Images should be versioned


lool:
It should be possible to pass some version string to the debos recipes to embed in the resulting images. This should be embedded in a file on the root filesystem like /.disk/info or similar and in files distributed with the artifacts.

PR #146, opened 2025-09-17
Add image build info

lool:
Add image build information when running debos. Specifically, add build id and variant/flavor to /etc/buildinfo.
- **feat(debos): Add image build information to /etc**
- **feat(ci): Pass our build id to debos builds**

Fixes: #37

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -94,6 +94,7 @@ jobs:
               -t xfcedesktop:true \
               -t aptlocalrepo:${PWD}/local-apt-repo \
               -t kernelpackage:'${{ inputs.kernelpackage }}' \
+              -t "buildid:${BUILD_ID}" \
               --print-recipe \
               debos-recipes/qualcomm-linux-debian-rootfs.yaml
 
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -2,6 +2,7 @@
 {{- $localdebs := or .localdebs "none" }}
 {{- $aptlocalrepo := or .aptlocalrepo "none" }}
 {{- $kernelpackage := or .kernelpackage "linux-image-arm64" }}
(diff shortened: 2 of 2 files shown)


basak-qcom (approved):
> I was ok with the conffile patching as I thought that a dist upgrade would replace the contents, and that seemed proper, but I don't like patching a regular file which has an expected debsum etc.

FWIW, I wouldn't be happy with that either, since it would mean that the user will receive a conffile prompt asking them to merge the changes on every upgrade, which they won't understand. But it's moot if you use /etc/buildinfo instead.

As we've discussed extensively elsewhere, I'm unhappy with using /etc for this now that it's not a defined field from the /etc/os-release spec (that seems more suitable for an image-based distribution like Yocto or other immutable distros). I think /var/log should be used instead. But it's more important that we have a decision to make progress and you've made it, so fine :)

As you've probably noticed I have a strong preference to add tests for everything that we change where possible, but pragmatically have been asking to validate changes PRs manually instead in cases where we don't have an appropriately straightforward way to add a test yet.

But maybe in this we can add a test? How hard would this be? I think we should check that the fields exist in the expected place in a test here, for example, and are of the form that we expect (eg. a regexp for each expected value). If our test infrastructure doesn't make this easy yet, I think it's OK to skip for now if we just validate the result manually instead.

lool:
I like the idea of a test; it would be nice to have a qemu test framework in place, we could also test this in LAVA, it's not very device specific though, so not sure there is good value in running a generic test like this on every device

Could we (could you :-p) resurrect this PR https://github.com/qualcomm-linux/qcom-deb-images/pull/74? I can help with workflows if needed, and add a test for the buildinfo stuff

lool:
I downloaded rootfs.tar.gz from CI and unpacked etc/buildinfo:
```
% tar xvf rootfs.tar.gz etc/buildinfo
x ./etc/buildinfo

% cat etc/buildinfo
BUILD_ID=17800097442-1
VARIANT_ID=xfce
```

PR #149, opened 2025-09-18
fix(ci): Bump download and overall timeouts by 180mn

lool:
Currently timing out at 20-25% when downloading, so multiplying the 20mn by 5 and and doubling that to give us a buffer.


EmbeddedAndroid:
> @EmbeddedAndroid please take a look. This is getting out of hand.

I've bumped the thread again. I agree, this is miserable.

PR #161, opened 2025-09-26
Add script to download a linux-image package and extract its config

lool:
Add script to download a linux-image package and extract its config, plus some
sample APT sources to use along the script
- **feat(scripts): Add get-debian-linux-config**
- **feat: Add some reference apt-sources**


gagath:
I do not understand why this is needed and is draft since some time. Closing.

lool:
This was research to derive our kernel config from the Debian one (Cc: @obbardc)

PR #163, opened 2025-09-29
fix(ci/mainline): Bump to 6.17.0

lool:


lool:
The LAVA tests are run on the default linux-qcom kernel in PRs, and don't reflect on the state of mainline unfortunately :-/

PR #165, opened 2025-09-29
fix(ci): Sanitize workflow inputs with an env var

lool:
Use an intermediate env var, sanitized by GitHub, to prevent
exploitation from untrusted user-provided input.

Reported-By: Alessandro Braccio <a.braccio@arduino.cc>
Signed-off-by: Loïc Minier <loic.minier@oss.qualcomm.com>


lool:
@rhpco Would appreciate your review!

PR #169, opened 2025-09-30
Qemu test expect failure output

basak-qcom:


lool:
@basak-qcom I don't see any console output in the build log; I was expecting to see the qemu boot log there?

PR #175, opened 2025-10-04
mesa: update to 25.2.4-1

lumag:
Update mesa-unstable to the freshly uploaded Mesa 25.2.4-1. Use the version with tilde in it to make sure that it's lower than the version in unstable.


lool:
I've built mesa for arm64 and amd64 here:
https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/18259652044
but I can't publish it to QArtifactory as the JF token expired once again, so waiting for an update from EngIT

lool:
> I've built mesa for arm64 and amd64 here: https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/18259652044 but I can't publish it to QArtifactory as the JF token expired once again, so waiting for an update from EngIT

Published to trixie-overlay-staging for testing

lumag:
> (Perhaps we should start looking at contributing to a backport in Debian so that we don't it just here?)

It's being handled by the X team: 
https://lists.debian.org/debian-x/2025/10/msg00017.html

daissi:
> (Perhaps we should start looking at contributing to a backport in Debian so that we don't it just here?)

That would be great! :relaxed: There is strong demand, the work is trivial but the frequency of releases is quite high.

PR #176, opened 2025-10-06
fix(ci): Rename to lava-test.yml in linux workflow

lool:
test.yml was renamed to lava-test.yml but the linux deb workflow was
missed.


basak-qcom (approved):
Sorry, I'm not sure how I missed this. The failing check seems unrelated and I think it's fine to merge this without trying to fix that for this PR.

Issue #66, opened 2025-06-05
QEMU support


lool:
We currently don't have a QEMU image, which would be useful as a fast, reliable and generally available platform. We should support building and testing of an image for QEMU.

PR #184, opened 2025-10-15
run-qemu.py

lool:
Add a new run-qemu.py helper script and update docs.
- **feat(scripts): Add run-qemu helper**
- **refactor: Update README to use new run-qemu.py**

Fixes: #66

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,75 +182,44 @@ NB: It's also possible to run qdl from the host while the baord is not connected
 
 Want to join in the development? Changes welcome! See [CONTRIBUTING.md file](CONTRIBUTING.md) for step by step instructions.
 
-### Test an image locally with qemu
+### Boot an image locally with QEMU (helper script)
 
-You can boot an image locally with qemu as follows:
+Use the `scripts/run-qemu.py` helper to boot generated disk images under QEMU. It automatically:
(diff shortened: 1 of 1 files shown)


scripts/run-qemu.py
```diff
@@ -0,0 +1,294 @@
...
+                "qcow2",
+                "-F",
+                "raw",
+                overlay_path,
+            ]
+            print("Running:", " ".join(cmd))
+            subprocess.run(cmd, check=True)
+        except subprocess.CalledProcessError as e:
+            sys.stderr.write(f"Failed to create COW overlay: {e}\n")
+            if temp_dir:
```
basak-qcom: I think this method of cleaning up temp_dir is a bit error prone. Could you instead either:

1) Create the TemporaryDirectory unconditionally using `with`

2) Or if you don't like doing it unconditionally, you could use `with contextlib.ExitStack() as exit_stack` and push the TemporaryDirectory to the exit_stack if you create it.

I think I prefer the former for simplicity but I don't mind either way.

basak-qcom (changes requested):
This looks good!

I wondered if we should try and have some common code between this and the implementation in ci/qemu_test.py but given the current level of maturity of this project I don't think it's worth going that far right now. For similar reasons I think the direct `sys.exit()` calls are OK for the time being.

A couple of Python pattern comments inline which I think are worth changing. No need to ask for re-review before merging though.

PR #185, opened 2025-10-15
APT fixes

lool:
Collection of important APT fixes:
* fix(debos/rootfs): Add missing trixie suites
* fix(debos/rootfs): Update debootstrap components
* fix(debos/rootfs): Install ca-certificates
* feat(debos/rootfs): Add Debian backports
* feat(debos/rootfs): APT pinning for pipewire
* fix(debos): Bump imagesize to 4.5GiB
* fix(debos/rootfs): Local APT repo needs no upgrade

The most crucial fix is the addition of security (and stable updates) repositories to our default APT sources; the rest is tweaking the APT setup to be in line with what we want to do (use mesa from backports, avoid libfreeaptx for compliance for now by using pipewire from our repo) and fixes.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -29,7 +29,7 @@ actions:
       mount --bind {{$aptlocalrepo}} "${ROOTDIR}/media/apt-local-repo"
 
   - action: run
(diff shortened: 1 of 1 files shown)


lool:
In the v1, I also added an apt update and upgrade, and indeed I can see that while our daily has a vulnerable libssl (3.5.1-1), the v1 rootfs picked up 3.5.1-1+deb13u1.

PR #187, opened 2025-10-17
feat(debos/flash): Updates to boot fw

lool:
Update all boards to recent boot firmware, aligned with meta-qcom. RB1 and CDTs haven't changed.
- feat(debos/flash): Update to r1.0_00106.0 boot fw
- feat(debos/flash): Update to r1.0_00108.0 boot fw
- feat(debos/flash): Update to r1.0_00110.0 boot fw
- feat(debos/flash): Update to r1.0_00112.0 boot fw
- feat(debos/flash): Update to r1.0_00114.0 boot fw

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -28,10 +28,10 @@ actions:
     "ptool_platforms" (list "qcs615-ride/ufs")
     "boot_binaries_download" (dict
         "description" "QCS615 boot binaries"
-        "url" "https://softwarecenter.qualcomm.com/download/software/chip/qualcomm_linux-spf-1-0/qualcomm-linux-spf-1-0_test_device_public/r1.0_00112.0/qcs615-le-1-0/common/build/common/bin/QCS615_bootbinaries.zip"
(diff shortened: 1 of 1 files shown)


basak-qcom:
IIUC meta-qcom is on 00103.0: https://github.com/qualcomm-linux/meta-qcom/tree/master/recipes-bsp/firmware-boot

Do you think it would make sense to stay aligned, for less painful development and debugging of related issues on both sides? For bonus points, maybe we can automatically have one follow the other :-)

Otherwise, this looks fine.

basak-qcom:
From https://github.com/qualcomm-linux/meta-qcom/pull/1345 I think this needs bumping to 0.114.0 now (thanks Ricardo!)

ricardosalveti (approved):

ricardosalveti:
> A few notes:
> 
> * I can't realistically retest all boot firmware manually, it will have to be automated tests only, and waiting for bug reports; I think it's ok as meta-qcom is already testing those
> * there's a new boot firmware, 116, but we haven't received a request to merge it in meta-qcom, and I didn't see changelog for it, so I'll wait for that before updating here
> * we don't have hamoa-iot (iq-x) support in qcom-deb-images, so obviously can't update that boot firmware which is hosted on a weird other place

Just to add a bit more,  in 0.114.0 we also support fit-dtb (which we are using on a few targets in meta-qcom), and I believe it might be good to add support for it here as well at some point.

PR #189, opened 2025-10-20
docs(README): Use shields.io to fix broken badges

lool:
GitHub badges require authentication for projects that one is a member
of. Members of the project who should really see these badges will have
to login with GitHub before seeing these. Use shields.io as a badge
proxying and caching service.

Use this opportunity to modernize the badge and list the mainline
builds.


lool:
> Will shields.io have access to actions? I think this is the tricky bit. Github asks me to authenticate if I want to see actions.

Yeah, you can try the URLs with img shield, they work in e.g. an incognito window :)

PR #198, opened 2025-11-14
https by default

### LINKED ISSUE #290: Switch to https mirrors by default
Consider switching to `https://deb.debian.org/` instead of using http. I think this would match the defaults in apt and makes more sense nowadays.

lool:
Review http vs https usage in the project and use https where it makes sense.
- **fix!(debos): Use https for APT by default**
- **fix(ci): Bootstrap chroots with https**
- **feat(Makefile): Also set https_proxy**

This is particularly important in the context of compliance as http:// might be seen as insecure, or at least lacking confidentiality.

Fixes: #290

### DIFF
--- a/.github/workflows/build-overlay-deb.yml
+++ b/.github/workflows/build-overlay-deb.yml
@@ -61,7 +61,7 @@ jobs:
           if [ "${suite}" = experimental ]; then
               # special extra args for experimental
               debootstrap_suite='unstable'
-              extra_repo='--extra-repository=deb   http://deb.debian.org/debian  experimental main'
+              extra_repo='--extra-repository=deb   https://deb.debian.org/debian  experimental main'
               chroot_prefix='--chroot-prefix=experimental'
           fi
           sudo sbuild-createchroot --include=eatmydata,ccache \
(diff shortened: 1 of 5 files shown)


basak-qcom:
I'm OK with shipping sources.list with https, but can we continue to use plain http for development and testing please? Otherwise caching becomes impossible.

lool:
After feedback from upstream, I'm not sure I'll pursue this branch, but I should at least move it to draft.

lool:
I'm using apt-cacher-ng, and https support is not fun to use

basak-qcom:
AIUI, apt-cacher-ng only supports HTTPS passthrough, ie. not caching? The only way to cache it would be to provide an MITM certificate for use on the client.

I also bumped into there being no HTTP support for https://qartifactory-edge.qualcomm.com/artifactory/qsc-deb-releases. That is shipping hexagon-dsp-binaries currently, so this hurts caching :-(

gagath:
Closing this as it makes using apt-cacher-ng painful for building images. Feel free open another PR that does this only for the final image and not for the building of the image eventually.

Issue #193, opened 2025-11-04
Sudoers should disable FQDN lookups


lool:
The default sudoers config in Debian will do FQDN lookups by default. This operation may block/timeout if the DNS/network is misconfigured.

This is quite common in typical usage scenarios, so we should disable this config by default. This is how Apertis does it:
https://gitlab.apertis.org/infrastructure/apertis-image-recipes/-/blob/apertis/v2026pre/overlays/sudo-fqdn/etc/sudoers.d/disable-fqdn?ref_type=heads

PR #199, opened 2025-11-14
Do not use sudo FQDN lookups

basak-qcom:
Details are in the individual commit messages.

### DIFF
--- a/ci/qemu_test.py
+++ b/ci/qemu_test.py
@@ -13,6 +13,12 @@ import types
 import pexpect
 import pytest
 
+# Since the first test checks for the mandatory password reset functionality
+# that also prepares the VM for shell-based access, we make the additional
+# optimisation that the fixture for a logged in VM re-uses that VM, so the
+# ordering of [plain VM fixture, password reset test, logged-in VM fixture]
+# matters here.
+
 
 @pytest.fixture(scope="module")
 def vm():
@@ -58,10 +64,14 @@ def vm():
                 "-nographic",
                 "-bios",
                 "/usr/share/AAVMF/AAVMF_CODE.fd",
+                "-fsdev",
+                f"local,id=fsdev0,path={os.getcwd()},security_model=none",
+                "-device",
+                "virtio-9p-pci,fsdev=fsdev0,mount_tag=qcom-deb-images",
             ],
         )
         spawn.logfile = sys.stdout.buffer
-        yield types.SimpleNamespace(spawn=spawn)
+        yield types.SimpleNamespace(spawn=spawn, logged_in=False)
 
         # No need to be nice; that would take time
         spawn.kill(signal.SIGKILL)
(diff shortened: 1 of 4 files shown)


basak-qcom:
```
Author: Robie Basak <robie.basak@oss.qualcomm.com>
Date:   Fri Nov 14 14:28:01 2025 +0000

    ci: add and integrate qemu_guest_test.py
    
    This allows us to write tests that run natively on the guest, and
    qemu_test.py will invoke them using qemu.
    
    Signed-off-by: Robie Basak <robie.basak@oss.qualcomm.com>
---
 ci/qemu_guest_test.py | 22 ++++++++++++++++++++++
 ci/qemu_test.py       | 51 ++++++++++++++++++++++++++++++++++++++++++++++++++-
 pyproject.toml        |  7 +++++++
 3 files changed, 79 insertions(+), 1 deletion(-)
 ```

80 lines :-)

File: ci/qemu_test.py [hunk 1]
```diff
@@ -89,3 +99,42 @@ def test_password_reset_required(vm):
...
+def test_using_guest_tests(logged_in_vm):
+    """Run the tests in qemu_guest_test.py inside the qemu guest"""
+    # Statement of test success and failure that are unlikely to appear by
+    # accident
+    SUCCESS_NOTICE = "All ci/qemu_guest_test.py tests passed"
+    FAILURE_NOTICE = "Some ci/qemu_guest_test.py tests failed"
+    # We use apt-get -U here and the apt_dependencies fixture in
+    # qemu_guest_test.py relies on this.
+    SCRIPT = f"""sudo -i sh <<EOT
+apt-get install -Uy --no-install-recommends python3-pytest
```
lool: It's quite unfortunate to introduce a network dependency :-(

I also don't like that we're changing the rootfs, but I guess we need to bring the tests somehow.

Should we just add this test runner to the image?

lool (approved):
So I have some doubts about the architecture of this approach:
- using expect was necessary for the login interaction, but seems generally brittle and to be avoided; I guess that's only for the bootstraping phase
- the dependencies on downloading packages and changing the rootfs are kind of heavy
- I would personally prefer if the tests were driven from the host towards the DUT rather than installing an on device framework and I'm not a big fan of the gdb usage here, but there's a limit to what we can do there; in this case, I guess we could set the DNS and check if there are packets coming its way, this would also work with sudo-rs

Anyway, if this works, it is understandable enough that I feel I could debug it if needed and the python and testing code is clean; I've left a few comments to your discretion :)

gagath:
@obbardc do you think you can take over this PR? I wanted to rebase it eventually but never got the bandwidth, and I think being able to run tests from a pytest file instead of expect would be nice!

obbardc:
> Christopher Obbard (@obbardc) do you think you can take over this PR? I wanted to rebase it eventually but never got the bandwidth, and I think being able to run tests from a pytest file instead of expect would be nice!

Yes; please assign this to me.

Also this may Close: #193 ; need to make sure that's somewhere in a commit message.

lool:
Closing this superseded PR

PR #206, opened 2025-12-02
Ptool update to bring Lemans EVK support

lool:
Update ptool to the latest git commit to get support for the latest boards. Take advantage of the new flag to set build-id.

Add support for Lemans EVK, also known as IQ-9075 EVK (was RB8). Confusingly, this QCS9075 EVK is using the QCS9100 boot binaries.


lool:
This needs https://github.com/qualcomm-linux/qcom-deb-images/pull/186 first

lool:
actually lemans-evk support only landed after 6.17, so switching to the backports kernel is not enough to get this in the flat build.

PR #226, opened 2026-01-29
UNO Q support

lool:
Changes to add support for UNO Q:
- **feat(debos/flash): Update to ptool d46b04b**
- **feat(debos/flash): Use boot.img from boot binaries**
- **feat(debos/flash): Add support for UNO Q (16GB)**

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -171,6 +171,19 @@ actions:
     "dtb" "qcom/qrb2210-rb1.dtb"
 )}}
 {{- end }}
+{{- $boards = append $boards (dict
+    "name" "qrb2210-arduino-imola"
+    "silicon_family" "qcm2290"
+    "ptool_platforms" (list "qrb2210-unoq/emmc-16GB")
+    "boot_binaries_download" (dict
+        "description" "UNO Q boot binaries"
+        "url" "https://downloads.arduino.cc/debian-im/unoq-bootloader-emmc-linux-251020.zip"
+        "name" "qrb2210-arduino-imola_boot-binaries"
+        "filename" "qrb2210-arduino-imola_boot-binaries.zip"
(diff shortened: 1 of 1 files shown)


lool:
I got this booting after changing the flash recipe to copy *.img files as well. This copies the u-boot binary from Arduino's boot binaries zip into the flash_qrb2210-arduino-imola, as well as boot-erase.img. It has the side effect of doing the same thing for RB1 (boot.img and boot-erase.img are present there too).

Initially I was hoping we wouldn't need an u-boot build for Imola in this project, but I guess I prefer us building GPL U-Boot explicitly rather than having it sneak by in the Imola case.

Funnily, flash_qrb2210-rb1_emmc/boot.img is actually broken, it contains some HTML fragments from somewhere; I'm not sure why. And boot-erase.img files are zeroes in both cases.

b49020:
@lool 

For RB1, I would suggest we just migrate to U-Boot qcom-next branch. Regarding capsule update patches, I don't see a clear path to upstream them without EFI runtime variables support.

For UNO-Q, let's stick with Arduino's branch as it has much more delta related to Wifi MAC address DT fixups and other features. Not sure if Arduino has any plans for pushing U-Boot changes upstream.

lool:
> @lool
> 
> For RB1, I would suggest we just migrate to U-Boot qcom-next branch. Regarding capsule update patches, I don't see a clear path to upstream them without EFI runtime variables support.

Then I think we should drop these patches; I don't want to carry long-lived patches in our mainline-centric reference builds/images.

Would you agree it makes sense to do this CI directly in u-boot:qcom-next? I could pull the binaries from there and drop the crazy build I'm doing in qcom-deb-images :)

> For UNO-Q, let's stick with Arduino's branch as it has much more delta related to Wifi MAC address DT fixups and other features. Not sure if Arduino has any plans for pushing U-Boot changes upstream.

Let's check-in with their team on these next Thursday?

PR #230, opened 2026-02-08
feat(scripts): New check-boot-firmwares helper

lool:
This helper is mostly for maintenance of the flash recipe, helping check
whether new boot binaries or CDTs have been released, and computing
their SHA256.


gagath:
@lool Tech teams are [already able to ask us to update boot firmware](https://github.com/qualcomm-linux/qcom-deb-images/pull/413) without the help of this script, so I doubt it is that useful. Closing.

PR #244, opened 2026-02-17
feat: add gnome-desktop variant

gagath:
This reuses the existing work for the XFCE variant, the difference is that task-gnome-desktop was not manually pruned down as I cannot think of specific packages to remove from it yet. GNOME is not known for being small on disk in the first place.

Due to this, the image size and the debos `--scratchsize` need to be bumped to 7GiB. The resulting root partition in disk-sdcard.img has 529MiB free space left out of 5.4 GiB.

The generated image was successfully boot-tested on qemu and a Thinkpad T14s. GDM asks for a password change and then the GNOME desktop can be used as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
   debos expects relative pathnames)
 - `xfcedesktop`: install an Xfce desktop environment; default: console only
   environment
+- `gnomedesktop`: install a GNOME desktop environment; default: console only environment
 - `overlays`: a `,`-separated list of rootfs overlays to add from
   `debos-recipes/overlays/`. See the *Supported overlays* section below.
 - `kernelpackage`: name of the kernel package to install from apt; defaults to
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -1,4 +1,5 @@
 {{- $xfcedesktop := or .xfcedesktop "false" }}
+{{- $gnomedesktop := or .gnomedesktop "false" }}
 {{- $localdebs := or .localdebs "none" }}
 {{- $aptlocalrepo := or .aptlocalrepo "none" }}
 {{- $kernelpackage := or .kernelpackage "linux-image-arm64" }}
@@ -9,6 +10,8 @@
 {{- $variantid := "console" }}
 {{- if eq $xfcedesktop "true" }}
 {{- $variantid = "xfce" }}
+{{- else if eq $gnomedesktop "true" }}
+{{- $variantid = "gnome" }}
 {{- end }}
 
 # Rename sid to unstable to reduce if/else complexity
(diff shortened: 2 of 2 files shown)


debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -252,6 +253,40 @@ actions:
...
 {{- end }}
 
+{{- if eq $gnomedesktop "true" }}
+  - action: apt
+    description: Install Gnome desktop
+    recommends: false
+    packages:
+      - plymouth-label- # we probably don't need a boot splash
+      # from task-gnome-desktop
+      - task-gnome-desktop
```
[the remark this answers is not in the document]
gagath: Makes sense.

I see that `task-gnome-desktop` depends on `gnome-core` and only recommends `gnome`. The package even says that it is to allow it to fit on a CD-ROM. So I will retry by replacing this section with `gnome-core` instead of the full `task-gnome-desktop`.

gagath:
Pushed the `gnome-core` WIP changes for the CI to have a go at it, image size is now 5.5 GiB.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -1,5 +1,5 @@
 {{- $dtb := or .dtb "firmware" }}
-{{- $imagesize := or .imagesize "4.5GiB" }}
+{{- $imagesize := or .imagesize "7.0GiB" }}
```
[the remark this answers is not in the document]
lool: I'm proposing a bump here https://github.com/qualcomm-linux/qcom-deb-images/pull/286, hopefully enough to fit GNOME too

gagath:
@lool On retrospective having the problematic codecs out of the image is a long-run task. If we want to wait for the image to be perfect, we are delaying GNOME image support for users of qcom-deb-images whom would like to try it soon to exercise the GPU accel and/or Wayland on hardware.

As long as we do not distribute the produced binaries, should we consider merging this as a first step since "depolluting" the image appears to be a long term effort? This PR already produced (undistributed) binaries, merging it would allow users to grab the images from the daily builds (abeilt some CI changes) rather than from this burried PR.

gagath:
CI is happy, marking as ready for review (see previous comment for rationale).

.github/workflows/debos.yml
```diff
@@ -118,7 +118,7 @@ jobs:
 
           debos \
               -t overlays:'${{ inputs.overlays }}' \
-              -t xfcedesktop:true \
+              -t gnomedesktop:true \
```
lool: I think we need more mileage and wider discussion before switching the default

However, we could add an extra GNOME build, perhaps only a daily forky one?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -363,6 +366,95 @@ actions:
...
+      - tecla
+      - adwaita-icon-theme
+      - glib-networking
+      - gsettings-desktop-schemas
+      - baobab
+      - evince
+      - gnome-calculator
+      - gnome-calendar
+      - gnome-characters
+      - gnome-clocks
```
lool, line 415: a few are missing here when compared to forky's gnome-core; perhaps the comment should say it's based on the trixie or forky one?
- gnome-connections
- ptyxis
- loupe
- papers
- showtime
- gnome-snapshot
- cups
- evolution-data-server
- fonts-adwaita
and a few others

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -363,6 +366,95 @@ actions:
...
+      - wireplumber
+      # camera: mostly supported through pipewire, but pull the qcam app
+      # and cam CLI tool from libcamera-tools; also install libcamera
+      # gstreamer plugins for for gstreamer based apps; ideally, cam and
+      # qcam would be packaged separately as to allow pulling the CLI
+      # version outside of the desktop list. libcamera-v4l2 provides a
+      # LD_PRELOAD library for compatiblity with apps relying on v4l2
+      # (similar to pipewire and pulse situation)
+      - gstreamer1.0-libcamera
+      - libcamera-tools
```
lool, line 470: arguably, these should be moved up a layer, but unfortunately I think they depend on Xorg or similar

lool (changes requested):
See review comments, please:
- don't change the default at this point (or add a new build)
- let's clarify which suite this is based on / supported on

.github/workflows/debos.yml
gagath: Yes, sorry, this is a WIP commit to trigger the CI on the PR. Will drop it before merge.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -274,6 +309,8 @@ actions:
       echo "BUILD_ID={{$buildid}}" >>/etc/buildinfo
 {{- if eq $xfcedesktop "true" }}
       echo "VARIANT_ID=xfce" >>/etc/buildinfo
+{{- else if eq $gnomedesktop "true" }}
+      echo "VARIANT_ID=gnome" >>/etc/buildinfo
```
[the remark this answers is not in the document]
gagath: I deliberately choose to not fix this for this PR. I would like to introduce a new selection system for console/xfce/gnome, but in a later PR.

.github/workflows/debos.yml
gagath: Dropped.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: Good catch, created #416 to keep track and added a comment referring to it.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: apt has a dependency solver? I just copied from xfce here, and am not sure if you want me to try to move them up for xfce and gnome (would be a separate PR I guess).

gagath:
Addressed comments. The WIP commit to have the PR CI use $gnomedesktop has been dropped, so do not expect too much from the next PR CI run.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: Yes; this could be a separate PR, agreed

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -368,6 +366,99 @@ actions:
       - libcamera-v4l2
 {{- end }}
 
+{{- if eq $gnomedesktop "true" }}
+# The following list of package is a manual selection from the gnome-core
+# metapackage from trixie with some pruning. The list of packages is longer in
+# forky, and needs to be revisited.
```
lool: perhaps best to error out if not building for trixie then

README.md
```diff
@@ -145,6 +145,7 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
   debos expects relative pathnames)
 - `xfcedesktop`: install an Xfce desktop environment; default: console only
   environment
+- `gnomedesktop`: install a Gnome desktop environment; default: console only environment
```
lool: GNOME

lool (changes requested):
s/Gnome/GNOME/ and I would error out if not building for trixie / if building for forky, but otherwise LGTM

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: agreed, done.

gagath:
Comments addressed. Rebased on top of main to get rid of the Monaco/LAVA failure.

lool (approved):

Issue #246, opened 2026-02-18
Lots of compression/decompression cycles when building an image


gagath:
I have the impression that there are compression cycles at the end of one step, and a decompression cycle at the start of the next. This could be optimized out by only compressing the last step, but I am bothering why the images are compressed in the first place. If one wants to then flash the produced image to an amovible media, then the image needs to be decompressed again.

If this is to reduce artifact size in CI, I suggest that the compression is performed in the CI steps instead of in the user steps.

PR #253, opened 2026-02-20
feat: remove rootfs.tar compression/decompression step

gagath:
gzip compression only uses a single core, and we are decompressing the rootfs right after when building an image taking the rootfs as its input.

On my build machine, here are the timings saved:

- compression: 128 s
- decompression: 48 s

The Makefile rules are updated to reflect the change.

The CI workflow is updated to compress the rootfs before staging it as an artifact for upload, saving some bandwidth and storage space only where it seems necessary.

This commit is the first work regarding trying to reduce the number of compression/decompression cycles described in #246.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -140,6 +140,8 @@ jobs:
           # create a directory for the current run
           dir="debos-artifacts"
           mkdir -v "${dir}"
+          # compress output files before staging them
+          gzip --keep rootfs.tar rootfs.tar.gz
           # copy output files
           cp -av rootfs.tar.gz "${dir}"
           cp -av dtbs.tar.gz "${dir}"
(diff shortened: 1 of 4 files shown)


gagath:
> FWIW, `pigz` is a drop-in replacement and does support multithreading. Another way to save time is to look at the compression level versus size and time. The default level (6) is not always the best tradeoff between size and time, especially when we have a lot of space taken up by /lib/firmware.

We would still be compressing to decompress it right away. I even considered lz4 to go super fast, but what’s the point? The only use of compressing it seems to be to reduce artifacts size in CI.

gagath:
Removing draft status, the CI failure is a false positive now documented in #254.

basak-qcom:
Looks good. Thanks!

Issue #254, opened 2026-02-20
Flaky wlan-smoke-test


gagath:
The wlan-smoke-test seems to fail on its own with the following error:

```
Device "wlan0" does not exist.
```

Example faulty run: https://lava.infra.foundries.io/scheduler/job/146693#bottom

This seems to be because the test do not have enough time to detect the wlan interface in the case that the ath11k driver takes more time than expected to probe. The test should either have an increased timeout, or be disabled.

PR #261, opened 2026-02-25
CI: Use new WAIT_TIME support for wlan-smoke-test for rb3gen2/rb1

mattface:
The wlan interface on rb3gen2 is slow to appear after boot, so use
the new support in test-definitions to wait 60 seconds until it
does. Use same wait for RB1 just in case.

Also add a timeout to the test action so that it does not use a default,
as our new wait time could exceed that default.

Fixes: #254


lool (changes requested):
I looked at the new WAIT_TIME implementation in test-definitions, that looked fine; could you:
- add a word about the addition of a test timeout in your commit message
- make the same change for rb1

Issue #256, opened 2026-02-23
qcom-next fails to boot on RB1


koenkooi:
It fails with:

[2](https://lava.infra.foundries.io/scheduler/job/148691#L722)Failed to load EFI variables
[L723](https://lava.infra.foundries.io/scheduler/job/148691#L723)FAT sector size mismatch (fs=4096, dev=512)
[L724](https://lava.infra.foundries.io/scheduler/job/148691#L724)FAT sector size mismatch (fs=4096, dev=512)
[L725](https://lava.infra.foundries.io/scheduler/job/148691#L725)FAT sector size mismatch (fs=4096, dev=512)
[L726](https://lava.infra.foundries.io/scheduler/job/148691#L726)FAT sector size mismatch (fs=4096, dev=512)

Which is a sector size issue.

gagath:
@koenkooi which image variant did you flash? The UFS image uses a sector size of 4096, while the sd/emmc variant uses 512:

```
sectorsize: {{if eq $imagetype "ufs"}}4096{{else}}512{{end}}
```

koenkooi:
I did use the emmc one, I'll retry today to verify.

koenkooi:
Looking at the [LAVA job](https://lava.infra.foundries.io/scheduler/job/148691#L723), LAVA flashed it using https://quic-yocto-fileserver-1029608027416.us-central1.run.app/qualcomm-linux/qcom-deb-images/22295854451-1//flash-emmc.tar.gz

gagath:
This issue seems to be specific to the MMC storage of the RB1 and RB2 using a 4096 GPT sector size [according to this patch proposal from Linaro](https://patchwork.ozlabs.org/project/uboot/patch/20230730111516.v2.1.Ia13846500fab3d5a1d5573db11a040d233994fa6@changeid/). Seems kernel related and specific to the RB1/RB2. We should track the upstream progress to see which patch lands and is applied to qcom-next. Latest patch I could find seems to be [this v5 patch from a Qualcomm OSS contributor](https://lore.kernel.org/u-boot/20260217035643.2383088-1-varadarajan.narayanan@oss.qualcomm.com/).

idlethread:
We're working to get RB1 and UnoQ into the HW lab. In the meanwhile, @b49020 can you check on this with your local RB1.

b49020:
[...]

PR #274, opened 2026-02-27
[Do not merge] For qcom-next testing purposes on RB1

b49020:


gagath:
qcom-next works on RB1 as stated in #256 so this is probably not needed anymore. Closing.

PR #277, opened 2026-02-27
Set workflow default kernel to linux-image-arm64

basak-qcom:
Drop the default kernel specification from debos.yml. This causes the default in the rootfs debos recipe to be used, which is linux-image-arm64, to be fulfilled from debian-backports.

The previous kernel of linux-image-6.16.7-qcom1+ fails fastrpc_test, but it does work from linux-image-arm64. Further, that kernel has fallen behind because it required manual updates.

### DIFF
--- a/.github/workflows/build-on-pr.yml
+++ b/.github/workflows/build-on-pr.yml
@@ -26,15 +26,11 @@ jobs:
       fail-fast: false
       matrix:
         suite: [trixie, forky]
-        kernelpackage: [default-ci, linux-image-arm64]
         exclude:
           - suite: trixie
-            kernelpackage: linux-image-arm64
           - suite: forky
-            kernelpackage: default-ci
     uses: ./.github/workflows/debos.yml
     with:
       suite: ${{ matrix.suite }}
-      kernelpackage: ${{ matrix.kernelpackage }}
   schema-check:
     uses: ./.github/workflows/lava-schema-check.yml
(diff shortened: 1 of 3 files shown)


gagath:
The RB1 changes have made it to the latest linux-image-arm64 in experimental. When this lands in unstable and then testing (forky) we can at least do the switch for forky. And then for trixie when it arrives in backports eventually.

gagath:
Taking over this.

The following changes have made it into the trixie-backports kernel:


    [ Loïc Minier ]
    * [arm64] Enable Qualcomm SA8775P platforms: add SA_GCC_8775P,
      PINCTRL_SA8775P, INTERCONNECT_QCOM_SA8775P as builtins and
      SA_GPUCC_8775P, SA_CAMCC_8775P, SA_DISPCC_8775P, SA_VIDEOCC_8775P as
      modules.

Let’s see if the CI is happy with it and can run the LAVA tests with this kernel.

gagath:
Well, it seems I was a little bit too enthusiastic as the RB1 changes are not the ones that I quoted, and landed in 7.1~rc, which implies more waiting for them to arrive in trixie-backports.

gagath:
RB1 is still failing to boot the Debian kernel because of crypto self test. This will have to wait a bit more.

obbardc:
@basak-qcom I believe this has been superseded by the new [debian.yml workflow](https://github.com/qualcomm-linux/qcom-deb-images/blob/main/.github/workflows/build-debian.yml) and this can PR be closed, what do you think?

lool:
While RB1 still fails to boot with Debian kernel, we went ahead and switched to different kernels entirely now, so I think we can close this

PR #278, opened 2026-02-27
feat: add main .gitignore file

gagath:
Running debos produces a lot of different files that pollutes git status. These files should never be versioned. Create a main .gitignore file to filter out:

- produced images
- produced rootfs
- produced debian artifacts


lool:
Looked good, thanks!

PR #280, opened 2026-02-27
Install FastRPC by default

basak-qcom:
Now that we have hexagon-dsp-binaries and fastrpc packaging in trixie-overlay, install them by default.

Once https://github.com/qualcomm-linux/qcom-deb-images/pull/277 is landed, this should mean that `fastrpc_test` passes by default.


lool (approved):
Looks good; could you add a PR description?

Would you mind adding a line documenting overlays in README? (I realize it was there before your changes, it's slightly more visible now)

lool:
This bumped rootfs size significantly, but this wasn't visible in the build-on-push or daily builds, probably due to differences with the kernel. I've submitted https://github.com/qualcomm-linux/qcom-deb-images/pull/286 to bump the rootfs size.

PR #286, opened 2026-02-28
fix: Bump image size to 6GiB

lool:
Addition of hexagon-dsp-binaries as part of AI stack makes the rootfs
significantly larger. Bump the image size to accomodate this and a
future GNOME flavor. Update scratchsize correspondingly.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -124,11 +124,11 @@ jobs:
           # qemu backend also requires to set scratchsize, otherwise the
           # whole build is done from memory and the out of memory killer
           # gets triggered
-          debos -b qemu --scratchsize 4GiB -t imagetype:ufs \
+          debos -b qemu --scratchsize 6GiB -t imagetype:ufs \
               ${DEBOS_EXTRA_ARGS} \
               --print-recipe \
               debos-recipes/qualcomm-linux-debian-image.yaml
-          debos -b qemu --scratchsize 4GiB -t imagetype:sdcard \
+          debos -b qemu --scratchsize 5GiB -t imagetype:sdcard \
               ${DEBOS_EXTRA_ARGS} \
               --print-recipe \
               debos-recipes/qualcomm-linux-debian-image.yaml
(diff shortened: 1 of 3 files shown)


lool:
So these daily builds failed today:
- monza
- linux-next

But these daily builds succeeded today:
- qcom-next
- daily

I compared the linux-next and qcom-next builds as these are closest; the linux-next build fails while building the first image recipe with debos:
```
2026-02-28T10:18:23.1781937Z 2026/02/28 10:18:22 ==== Unpack root filesystem ====
2026-02-28T10:22:23.6701179Z 2026/02/28 10:22:23 unpack | tar: ./boot/System.map-7.0.0-rc1-next-20260227: Wrote only 6144 of 10240 bytes
2026-02-28T10:22:23.6729805Z 2026/02/28 10:22:23 unpack | tar: ./boot/vmlinuz-7.0.0-rc1-next-20260227: Cannot write: No space left on device
2026-02-28T10:22:23.6782361Z 2026/02/28 10:22:23 unpack | tar: ./boot/config-7.0.0-rc1-next-20260227: Cannot write: No space left on device
2026-02-28T10:22:23.6821035Z 2026/02/28 10:22:23 unpack | tar: ./root: Cannot mkdir: No space left on device
```

This is happening during unpack, but I was curious as to why qcom-next would pass and not linux-next. I spotted two important differences:
- qcom-next is built with a special prune.config which disables a bunch of non-qcom platforms supported by default in defconfig; this cuts down the list drivers and of DTBs
- while generating an initramfs, there are a number of warnings about missing firmware (` apt | W: Possible missing firmware /lib/firmware/microchip/mscc_vsc8574_revb_int8051_29e8.bin for built-in driver mscc`) which are likely caused by trying to build support for these platforms while not having the corresponding firmware available (as we only install qcom firmware in the rootfs)

This makes me think we should aim to have something like prune.config in qcom-deb-images even for mainline or linux-next kernels, to cut down on warnings for end-users and to make the build faster, and the kernel and initramfs smaller.

lool:
Seeing this built and tested successfully, I'll go ahead and merge this to hopefully unbreak other builds.

basak-qcom:
Does the Makefile need tweaking as well? Interestingly I didn't hit this when testing locally.

PR #294, opened 2026-03-02
feat(Makefile): add clean target

gagath:
After the introduction of a .gitignore, one might need to reset their
repository in a clean state and `git clean -fd` does not work anymore.

Add a `clean` target to the Makefile to remove most likely produced
artifacts that users want to remove to reset the repository in a mostly
clean target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,3 +29,9 @@ test: disk-ufs.img
 	# rootfs/ is a build artifact, so should not be scanned for tests
 	py.test-3 --ignore=rootfs
 
+.PHONY: clean
+clean:
+	rm -f disk-ufs.img*
+	rm -f disk-sdcard.img*
+	rm -f rootfs.tar*
+	rm -f dtbs.tar.gz


basak-qcom:
FWIW, I use `git clean -ffxd` out of habit to ignore everything and "JFDI".

I have a mild aversion to Makefile clean targets. I think that VCS supersedes the need for these, with the trouble [with clean targets] being that they have to be maintained and therefore can miss things. But it's just some code (that I wouldn't use) so I don't have a huge objection to doing it if you insist.

Makefile
```diff
@@ -23,8 +24,14 @@ disk-ufs.img disk-ufs.img.gz: debos-recipes/qualcomm-linux-debian-image.yaml roo
...
 
+.PHONY: test
 test: disk-ufs.img
 	# rootfs/ is a build artifact, so should not be scanned for tests
 	py.test-3 --ignore=rootfs
 
-.PHONY: all test
+.PHONY: clean
+clean:
+	rm -f disk-ufs.img*
```
lool: Not a fan of these wildcards; I sometimes keep a renamed copy of previous builds to do local comparisons...

FWIW, I would have preferred being specific about the artifacts to clean.

gagath: Makes sense. Created #303 to fix this.

PR #295, opened 2026-03-02
feat!: compress image only for CI (5.5x speedup)

gagath:
Compressing the image feels like the most time-consuming task from the qualcomm-linux-debian-image.yaml recipe. Then users need to uncompress the image before flashing it to a device or mass storage media.

Move the compression step to the CI workflow to reduce build-flash-test iteration cycle duration for developers.

Measured image build time indicates a ~5.5x speed-up, which is very nice for local builds and testing:

- before: 143.55 s
- after: 26.09 s

Marked as potentially breaking change if external scripts are relying on the .gz file to be immediately present after running the recipe. Should not have much impact in practise.

Closes: #246.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -155,7 +155,7 @@ jobs:
           dir="debos-artifacts"
           mkdir -v "${dir}"
           # compress output files before staging them
-          gzip --keep rootfs.tar rootfs.tar.gz
+          gzip --keep rootfs.tar
           gzip --keep disk-ufs.img
           gzip --keep disk-sdcard.img
           # copy output files


.github/workflows/debos.yml
```diff
@@ -156,6 +156,8 @@ jobs:
           mkdir -v "${dir}"
           # compress output files before staging them
           gzip --keep rootfs.tar rootfs.tar.gz
+          gzip --keep disk-ufs.img disk-ufs.img.gz
```
lool: I didn't spot this in the previous PR, but the last arg to gzip is superfluous; should just be:
```
gzip --keep disk-ufs.img
```

Having both is redundant and triggers a warning:
```
lool@trixie:/tmp$ gzip --keep disk-ufs.img disk-ufs.img.gz
gzip: disk-ufs.img.gz already has .gz suffix -- unchanged
```

lool (changes requested):
Looks good to me, except superfluous gzip arg

gagath:
> @gagath can you rebase on tip?

Rebased on top of main, which changed `DEBOS` to `DEBOS_CMD`.

PR #311, opened 2026-03-05
workflows: only upload private artifacts to S3

mwasilew:
Stop using GCP for storing private artifacts. Move all uploads to S3.


basak-qcom:
The code changes all seem reasonable to me. I don't have the wider context around the change but I trust that Milosz does.

PR #314, opened 2026-03-08
Multi suite, multi distro

lool:
This is still exploratory and in the short-term meant to help compare
SBOMs across distros and suites
- **feat(rootfs): Add suite and distro variables**
- **feat(rootfs): Generate APT sources dynamically**
- **feat(rootfs): Generate Debian backports APT config**
- **feat(rootfs): Generate qsc-deb-releases APT config**
- **feat(rootfs): Add hostname and username parameters**
- **feat(rootfs): Per distro mirrors and components**
- **feat(Makefile): Pass suite and distro to debos**
- **ci(debos): Add suite and distro inputs, matrix build on PR**
- **WIP: Add matrix build to build-on-push too**


.github/workflows/debos.yml
```diff
@@ -13,10 +21,7 @@ on:
           linux-image-arm64) or name of a fileserver kernel package directory
           (e.g. fileserver/mainline)
         type: string
-        # the package name will be passed to APT which interprets the trailing
-        # plus sign in the package name as a request to install the package, so
-        # use two plus signs
-        default: linux-image-6.16.7-qcom1++
+        default: linux-image-arm64
```
gagath: Note that change is already pending in #277

gagath:
@lool I think this can be closed, I salvaged most of it in #350 which is now merged.

lool:
> @lool I think this can be closed, I salvaged most of it in #350 which is now merged.

Thanks!

lool:
I've looked at other commits in here, and went to salvage just the locale-gen fix.

Other interesting bits:
1) Ubuntu support: in general, we are unlikely to use debos to build official Ubuntu images, but these might make sense for some CI scenarios; however, I don't think we should land all fixes, it's too invasive

2) the mirror handling bits should probably be revisited; we will want the ability to build against defined snapshots; we could start with a snapshot.d.o PoC, and then switch to Debusine snapshots

3) I almost took over the DTB path handling, but I kind of don't like us having knowledge of this in the first place :D

Other random improvements have been cherry-picked in other PRs.

PR #316, opened 2026-03-09
feat!: Change default user and hostname to qcom

gagath:
For consistency in docs and CI, use qcom as default user, password, and hostname, no matter the distro.

Cherry-picked from #314, and added missing update to README with the new user/password.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -236,6 +236,7 @@ jobs:
         env:
           build_url: ${{ steps.upload_artifacts_s3.outputs.url }}
         run: |
+          echo '⚠️ Username and password are now "qcom"'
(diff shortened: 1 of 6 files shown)


lool (changes requested):
We need to update at least README, perhaps other places

This is a fairly invasive change; I'm not sure how we broadcast this more widely

lool (approved):

gagath:
If we take RaspberryPi as a reference there is the `pi` user with `sudo` like we have now. Same for Debian cloud images and Ubuntu.I would prefer to keep the separate user with sudo. I will try to see where I can put this message next to the download link.

EDIT: if you meant to add a qcom user to meta-qcom, sure. I am not sure about shipping sudo though.

lool:
a nice message

PR #331, opened 2026-03-18
feat(Makefile): add flash_glymur/ target

gagath:
Add a new target that downloads the Glymur firmware and extracts it in a flash_glymur/ directory. This allows for autocompletion thanks to Makefile targets and allows to do the following and get what one would expect:

	$ make flash_glymur/bootfw1.bin
	$ # or simply:
	$ make flash_glymur/


gagath:
Superseded by #332.

Issue #273, opened 2026-02-27
Add support for NVMe flashing


lool:
Hi

Some devices we're aiming to support, such as compute CRDs, boot from NVMe.

We should:
- create "generic" or "flash_generic" assets in our flash-emmc.tgz and flash-ufs.tgz tarballs
- update README to mention the NVMe option; this is shows how to flash: https://github.com/qualcomm-linux/qcom-deb-images/issues/264#issuecomment-3969643698

Perhaps we can contribute a generic platform to ptool, or create a trivial partitions.conf and call ptool with it.

PR #332, opened 2026-03-19
feat(flash): add glymur NVME support

gagath:
Add support for generating the flash_glymur-crd/ directory. Note that
the "spi-nor" ptool platform, while supported, is not generated as we do
not have spi-nor support yet.

Closes: #273


debos-recipes/qualcomm-linux-debian-flash.yaml
abickett: Flashing for Glymur (and hamoa) requires device programmer "xbl_s_devprg_ns.melf" - need to add *.melf here?

lool (approved):

PR #348, opened 2026-04-03
debos: install and enable openssh-server for SSH access

vemana1niranjan:
Install openssh-server and enable the ssh service during image build so SSH access is available immediately after first boot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,8 @@ Once the image has booted, you can log in as the `debian` user, with the
 default `debian` password. The image should then ask you to change this default
 password to a safe one.
 
+Note: "openssh server is installed but disabled by default; enable with "systemctl enable --now ssh" as root user".
+
 ## Development
 
 Want to join in the development? Changes welcome! See [CONTRIBUTING.md file](CONTRIBUTING.md) for step by step instructions.
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -302,13 +302,25 @@ actions:
       - wget
 
   - action: run
-    description: Add default user to docker group
+    description: Install openssh-server without generating host keys and report any existing keys
     chroot: true
     command: |
       set -eux
-      usermod -a -G docker debian
+      export DEBIAN_FRONTEND=noninteractive
+      install -d -m 700 /etc/ssh
+      touch /etc/ssh/ssh_host_rsa_key /etc/ssh/ssh_host_rsa_key.pub /etc/ssh/ssh_host_ecdsa_key /etc/ssh/ssh_host_ecdsa_key.pub /etc/ssh/ssh_host_ed25519_key /etc/ssh/ssh_host_ed25519_key.pub
+      chmod 600 /etc/ssh/ssh_host_rsa_key /etc/ssh/ssh_host_ecdsa_key /etc/ssh/ssh_host_ed25519_key
+      dpkg-divert --add --rename /usr/bin/ssh-keygen
+      ln -s /bin/true /usr/bin/ssh-keygen
+      apt-get -y install openssh-server
+      rm /usr/bin/ssh-keygen
+      dpkg-divert --remove /usr/bin/ssh-keygen
+      systemctl mask ssh.service 2>/dev/null || true
+      systemctl disable ssh.service 2>/dev/null || true
+      find /etc/ssh -maxdepth 1 -type f -name 'ssh_host_*_key' -size +0 | grep -q . && { echo "SSH host key files found"; exit 1; } || echo "No SSH host key files present"
 
 {{- if eq $xfcedesktop "true" }}
+
   # this should ideally be task-xfce-desktop, but this pulls too many apps;
   # instead, curate from the Depends and Recommends of task-xfce-desktop and
   # task-desktop recursively; also, disable Recommends


gagath (approved):
Makes sense to me, this avoids using the serial port to install a ssh server. This will also be useful when we move our qemu tests from `expect` to something a bit more robust.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -201,6 +203,16 @@ actions:
...
 
+  - action: run
+    description: Sanitize SSH setup
+    chroot: true
+    command: |
+      set -euo pipefail
+      rm -f /etc/ssh/ssh_host_* || true
+      rm -rf /root/.ssh /home/*/.ssh || true
+      systemctl disable ssh 2>/dev/null || true
+      find /etc/ssh /root /home -type f \( -name 'ssh_host_*' -o -name 'id_rsa*' -o -name 'id_ed25519*' -o -name 'authorized_keys' \) -exec rm -f {} + 2>/dev/null || true
```
lool: There are a few issues here; first, I think we need comments to clarify the intent and rationale:
"Disable SSH server on the installed system because the default password is weak" or something like that

On the ordering, I would also expect to disable + stop (but stop isn't needed) the service BEFORE removing its files.

For the first removals (rm calls) "Remove host keys generated on install"; I would change the rm to "rm -v", so that we see the actual removal in the log and so that it's not always successful (-f works even if files are present); I wouldn't guard with || true either: we expect these files to be present.

The removals are a bit heavy handed with /root and /home; I don't see why we'd actually have files there?

Lastly, the find makes sense as a sanity/safety check to not ship a rootfs without pre-generated identical keys on all devices, but not in the same block where we've just removed the files; right now, it's just a second rm. I think it should check and fail when these files are present. I'm not sure where the ideal place would be, but quite late in the build, perhaps next to the QEMU sanity tests we'd have a rootfs contents sanity test in a separate script?

lool (changes requested):
Left various proposed improvements; coud you also add a commit message and fixed your signed-off-by to use your full name?

lool (commented):
I'm not sure why my review wasn't fully sent

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -200,6 +200,8 @@ actions:
       # convenience networking commands (arp, ifconfig, route etc.)
       - net-tools
       - openssh-client
+      # OPENSSH is installed but disabled by default.
```
lool: this is a weird way to write that name; either OpenSSH or openssh

or just write "disabled by default" since that's the only interesting part

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -212,6 +214,15 @@ actions:
       - vulkan-tools
       - wget
 
+  - action: run
+    description: Sanitize SSH keys
+    chroot: true
+    command: |
+      set -eux
+      rm -f /etc/ssh/ssh_host_*
+      rm -rf /root/.ssh /home/*/.ssh
```
lool: Drop the root/.ssh part, there is no chance this will be generated

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -212,6 +214,15 @@ actions:
       - vulkan-tools
       - wget
 
+  - action: run
+    description: Sanitize SSH keys
```
lool: This action is about disabling SSH by default; update the description to match

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -212,6 +214,15 @@ actions:
...
+    chroot: true
+    command: |
+      set -eux
+      rm -f /etc/ssh/ssh_host_*
```
lool: Change this to check if the keys are present and fail if so; we don't want a silent removal, we want to escalate the issue if these keys were somehow generated.

Also, move this logic to AFTER disabling and potentially stopping the service.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -212,6 +214,15 @@ actions:
...
+      rm -rf /root/.ssh /home/*/.ssh
+      systemctl disable ssh 2>/dev/null || true
```
lool: Try to pass --now as well as that seems stricter.

Don't ignore failure, or explain why we want to run it if it's expected to fail.

README.md
```diff
@@ -170,6 +170,7 @@ For the flash recipe:
   `qcs9100-ride-r3`, `qrb2210-rb1`.
 
 Note: Boards whose required device tree (.dtb) is not present in `dtbs.tar.gz` are automatically skipped during flash asset generation.
+Note: "A SSH server is installed but disabled by default; enable with sudo systemctl enable ssh".
```
lool: This is the wrong place to add this change; look at the whole README and find the place which makes the most sense. I would place it near the login instructions since this is a note about usage.

lool (changes requested):

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vemana1niranjan: tried to add --now but
flag fails because systemd is not running in the chroot build environment

README.md
vemana1niranjan: adding at login instructions

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vemana1niranjan: made change logic starts after disabling if found printing error mesg in rootfs generation part and forcely exiting rootfs generation

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vemana1niranjan: changed description to "Disable ssh by default and sanitize ssh keys"

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vemana1niranjan: removed root/.ssh part kept the /home/*/.ssh

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vemana1niranjan: changed to "disabled by default"

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -212,6 +214,18 @@ actions:
...
+    description: Disable ssh by default and sanitize ssh keys
+    chroot: true
+    command: |
+      set -eux
+      systemctl disable ssh
+      if [ -f /etc/ssh/ssh_host_* ]; then
+        echo "ERROR: SSH host keys were unexpectedly generated" >&2
+        exit 1
+      fi
+      rm -rf /home/*/.ssh
```
lool: Please drop this rm, no chance of being created during build (otherwise it should be a check as above)

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -212,6 +214,18 @@ actions:
       - vulkan-tools
       - wget
 
+  - action: run
+    description: Disable ssh by default and sanitize ssh keys
+    chroot: true
+    command: |
+      set -eux
+      systemctl disable ssh
+      if [ -f /etc/ssh/ssh_host_* ]; then
```
lool: if this expands to multiple files, the test command will be incorrect

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -212,6 +214,18 @@ actions:
       - vulkan-tools
       - wget
 
+  - action: run
+    description: Disable ssh by default and sanitize ssh keys
+    chroot: true
+    command: |
+      set -eux
+      systemctl disable ssh
```
lool: please add `--now`

lool (changes requested):
Hi, some shell syntax issues

Please update the commit message to reflect the latest changes

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -301,6 +214,17 @@ actions:
       - vulkan-tools
       - wget
 
+  - action: run
+    description: Disable ssh by default and sanitize ssh keys
+    chroot: true
+    command: |
+      set -eux
+      systemctl disable ssh --now
```
lool: This might work, but it's an unusual position for flags; `disable --now ssh` instead

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -301,6 +214,17 @@ actions:
...
+      if [ -f /etc/ssh/ssh_host_rsa_key ] || [ -f /etc/ssh/ssh_host_ecdsa_key ] || [ -f /etc/ssh/ssh_host_ed25519_key ]; then
```
lool: This is not robust against addition of key types in the future...

lool (changes requested):

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vemana1niranjan: even this "systemctl disable --now ssh" local build failing with this error message
msg : --now cannot be used when systemd is not running.
 
There is interesting observation below in trixie build fails with same error and in forky build this error was not found

so shall i remove --now?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vemana1niranjan: tried to add a generic loop over ssh_host_*_key 

but observations are host keys are always copied in initial ssh-server package during debos install
so, removing them forcefully might be needed with (rm) 

else build will always fail
is there any alternative way?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: There seems to be conflicting situations in your attempts; you report commands fail because systemd is not running (and indeed, then --now makes no sense, I can see why systemd would reject it, and we should drop it), but some recent failures also show that --now is accepted, and that the SSH keys were generated, which suggest that we're in the opposite situation:
```
2026/04/29 10:53:34 ==== Disable ssh by default and sanitize ssh keys ====
2026/04/29 10:53:34 set -eux... | + systemctl disable --now ssh
2026/04/29 10:53:34 set -eux... | Removed '/etc/systemd/system/multi-user.target.wants/ssh.service'.
2026/04/29 10:53:34 set -eux... | Removed '/etc/systemd/system/sshd.service'.
2026/04/29 10:53:34 set -eux... | + [ -f /etc/ssh/ssh_host_ecdsa_key ]
2026/04/29 10:53:34 set -eux... | + basename /etc/ssh/ssh_host_ecdsa_key
2026/04/29 10:53:34 set -eux... | + echo ERROR: SSH host key (ssh_host_ecdsa_key) was unexpectedly generated
```

I'm not clear on when/how debos is preventing systemd from running, I would expect most image generation tools to work in chroots and disable system services entirely though some kind of policy files or replacing the binaries.

It's risky for us to run daemons such as SSH during image generation because we might end up with keys or ids that were generated once and are distributed everywhere, when they are supposed to be private or unique and generated on the first boot after flashing.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -301,6 +214,19 @@ actions:
...
       - wget
 
+  - action: run
+    description: Disable ssh by default and sanitize ssh keys
+    chroot: true
+    command: |
+      set -eux
+      systemctl disable --now ssh
+      for keyfile in /etc/ssh/ssh_host_*_key; do
+        if [ -f ${keyfile} ]; then
```
lool: This shell snippet is slightly broken, on some shells it will result in an error if the files are missing; on bash, it will expand to run the loop once with keyfile=/etc/ssh/ssh_host_*_key. Then the `[ -f` test call will try the same missing expansion again. It's very confusion and an unorthodox way of testing for the presence of a pattern of files.

Instead, I'd recommend using "find".

Add quotes around the keyfile use, it shouldn't be wildcard-expanded again.

lool (changes requested):

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: > tried to add a generic loop over ssh_host_*_key
> 
> but observations are host keys are always copied in initial ssh-server package during debos install so, removing them forcefully might be needed with (rm)
> 
> else build will always fail is there any alternative way?

this requires researching why in debos openssh is being started and generates keys; I believe it shouldn't

obbardc:
I was thinking about adding SSH support (currently I have a snippet manually that installs openssh-server) and found this PR.

I guess the current thinking is we install openssh-server but disable the daemon on first-boot; meaning the user has to manually enable it which can be quite an annoying thing, especially if the board is not in reach/headless?

In the past for these kinds of things where you want to manually add some useful development tools/enable development environment at first boot I have used a `$development` (or similar) variable which is false by default, in this case would just enable the SSH server at boot.

What do you think @lool ? Also, @vemana1niranjan what are your plans to land this PR ?

lool:
@obbardc we've been discussing ways to address, depending on context, both ambitions:
- I want a secure by default image that is not "debian / debian" on the Internet
- we want an easy way to get SSH for end-developers, or for automated testing (Axiom/LAVA)

Ideally we'd like this to also be possible in our Yocto images.

My current thinking is that it's a bit of a cloud-init or d-i pattern, albeit I don't to use cloud-init for that: we want for a particular deployment to tweak this or that setting. I believe systemd-creds would allow doing exactly this classs of tweaks: setting up a default password, enabling SSH, running some commands on first-boot etc.

I believe, but haven't researched, that we could generate an optional extra filesystem that would contain a systemd-creds payload and would be picked up automagically on boot when present. Users/infra could then decide whether they want to flash the version with or without this data, and we could provide a few different systemd-creds image builds: set the default password to debian + enable SSH, vs do nothing.

We could do an unsigned version of that as a first pass (currently, we don't implement any kind of disk encryption in these images, and you can e.g. replace the ESP with one that sets init=/bin/sh or something). Later, we could sign these with development keys to demonstrate how this could be done securely.

Issue #166, opened 2025-09-29
Fix issues reported by zizmor


lool:
Running zizmor 1.14.1 on current HEAD, I get:
90 findings (42 suppressed, 17 fixable): 0 informational, 0 low, 27 medium, 21 high

We should address or document these reported tags and add zizmor to make check and to our workflows.

PR #350, opened 2026-04-03
Multi-series Debian support

gagath:
This is a work in progress with multiple amended cherry-picks from PR #314. While we still target trixie as a final goal, having the possibility to use unstable allows us to verify which packages are in an eventual need of a trixie backport

Publishing as a draft to exercise the CI, first with existing trixie support, and then with a test matrix in additional commits to come..


gagath:
See this from a developer perspective. I agree that having unstable builds in CI might not be what we want. But as a DD, I want to be able to create an image based on unstable (or testing) in order to test the latest packages before we decide to backport some of them.

Regarding backports, I forgot to tackle the backports install for unstable as well as testing. Will amend the commit to only install backports when it makes sense, ie. not unstable nor testing.

gagath:
- rebased on top of #351 
- use tee instead of cat to see what is generated
- fixed a whitespace issue breaking the trixie apt source

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -230,8 +230,13 @@ actions:
       # some useful groups for desktop scenarios, but also to run payloads
       # from the serial console, over SSH, or in containers - where the desktop
       # session has not updated ACLs to the device nodes
+      groups="adm,audio,render,sudo,users,video"
+      # fastrpc group only exists when fastrpc-support is installed (Debian)
+      if getent group fastrpc >/dev/null 2>&1; then
+        groups="${groups},fastrpc"
+      fi
       useradd --create-home --shell /bin/bash --user-group \
-          --groups adm,audio,fastrpc,render,sudo,users,video debian
```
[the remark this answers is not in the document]
gagath: Not only Ubuntu but also Debian forky since fastrpc-tests is not available there, and the group will not be created. I am moving this cherry-pick to #351 as it will make more sense, and will rebase this branch on top of it next.

gagath:
Rebased on top of #351. I can add a CI commit to build forky as it seems it will exercise this code and not keep it dormant in CI.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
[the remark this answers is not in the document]
lool: fastrpc-support would create the group in Debian images too; fastrpc-tests is just the test binaries, this is the only part missing in Debian

lool:
> Rebased on top of #351. I can add a CI commit to build forky as it seems it will exercise this code and not keep it dormant in CI.

I'd love that! Thanks for breaking by stupidly large draft PR into this more manageable approach! :)

gagath:
Rebased on top of main.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -50,11 +50,38 @@ actions:
...
+      set -eux
+      tee /etc/apt/sources.list.d/debian.sources <<EOF
+      # Debian archive
+      Types: deb
+      URIs: http://deb.debian.org/debian/
+      Suites: {{ $suite }}
+      Components: main contrib non-free non-free-firmware
+      Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
+
+{{- if and (ne $suite "unstable") (ne $suite "testing") }}
```
lool: testing is likely ok to to allow, there is testing-updates and testing-security

gagath:
Cherry-picked and edited additional commits from #314 for supporting the new $series option in CI.

gagath:
Fakemachine seems to fail when building with the version from unstable that now Depends ([for an unknown reason](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1116451)) instead of Recommends systemd-resolved, which seem to be involved. Will retry the job, but maybe I will also cherry-pick the fix that always use trixie for builders.

gagath:
Got the CI to work after having it use Debian’s linux-image-arm64 when building for forky, I think this is ready for review!

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -4,6 +4,7 @@
 {{- $kernelpackage := or .kernelpackage "linux-image-arm64" }}
 {{- $overlays := or .overlays "qsc-deb-releases" }}
 {{- $buildid := or .buildid "" }}
+{{- $suite := or .suite "trixie" }}
```
lool, line 7: New option => let's add it to README

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: perhaps you want to exclude sid as well

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -83,9 +83,30 @@ actions:
...
+      Components: main contrib non-free non-free-firmware
+      Enabled: yes
+      Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
+      EOF
+      tee /etc/apt/preferences.d/debian-backports.pref <<EOF
+      # for binary packages built from these source packages, score the version from
+      # Debian backports higher as to get hardware enabled or better hardware support
+
+      Package: src:alsa-ucm-conf:any src:firmware-free:any src:firmware-nonfree:any src:linux:any src:linux-signed-arm64:any src:mesa:any src:u-boot-efi-dtb
+      Pin: release n={{ $suite }}-backports
```
lool, line 106: This is not something to change in this PR, but I am making a mental note that templated files is something that debos should support out of the box.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -203,6 +195,22 @@ actions:
...
+{{- if ne $overlays "none" }}
+{{- range $overlay := split "," $overlays }}
+{{- if eq $overlay "qsc-deb-releases" }}
+  # fastrpc-tests is only available in trixie-overlay
+  - action: apt
+    description: Install fastrpc-tests (trixie only)
+    recommends: true
+    packages:
+      # fastrpc support. fastrpc-tests pulls in the required stack
+      - fastrpc-tests
```
lool, line 212: Let's install Debian's fastrpc-support instead when fastrpc-tests is not an option; this will give us the mandatory daemons like the adsp service you need on your hamoa laptop :)

.github/workflows/debos.yml
```diff
@@ -40,16 +44,16 @@ env:
   # image build id; used for SBOM generation
   BUILD_ID: ${{ github.run_id }}-${{ github.run_attempt }}
   KERNEL_PACKAGE: ${{ inputs.kernelpackage }}
-  DEBOS_EXTRA_ARGS: ${{ inputs.debos_extra_args }}
+  DEBOS_EXTRA_ARGS: ${{ inputs.debos_extra_args }} -t suite:${{ inputs.suite }}
```
lool: Re-reading that, I'd flip this around to allow overriding the suite

.github/workflows/debos.yml
```diff
@@ -40,16 +44,16 @@ env:
...
 jobs:
   build-debos:
-    name: Build and upload debos recipes
+    name: Build and upload debos recipes (${{ inputs.suite }})
     outputs:
       url: ${{ steps.upload_artifacts_s3.outputs.url }}
     runs-on: [self-hosted, qcom-u2404, arm64]
     container:
-      image: public.ecr.aws/debian/debian:trixie
+      image: public.ecr.aws/debian/debian:${{ inputs.suite }}
```
lool: Maybe that should be its own commit; it's quite orthogonal to the other changes

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -143,6 +146,13 @@ actions:
       Pin: release n={{ $suite }}-overlay
       Pin-Priority: 950
       EOF
+        # verify the overlay suite exists before enabling it
```
lool: Perhaps this was slightly insane, we should just make sure an empty suite exists; I feel bad about the wget dependency

.github/workflows/debos.yml
```diff
@@ -142,16 +142,17 @@ jobs:
       - name: Stage debos artifacts for publishing
         run: |
           set -ux
+          prefix="${{ inputs.suite }}"
```
lool: this will probably trigger a security warning from zizmor; we should set this in an GH workflow env var before using

.github/workflows/debos.yml
```diff
@@ -203,10 +204,11 @@ jobs:
       - name: Stage SBOMs for publishing
         run: |
           set -ux
+          prefix="${{ inputs.suite }}"
```
lool: ditto, and since we use it multiple times, it's another motivation to define the artifact name prefix a level up

.github/workflows/debos.yml
```diff
@@ -53,7 +53,7 @@ jobs:
       url: ${{ steps.upload_artifacts_s3.outputs.url }}
     runs-on: [self-hosted, qcom-u2404, arm64]
     container:
-      image: public.ecr.aws/debian/debian:${{ inputs.suite }}
```
lool: Let's log this as an issue somewhere, and add a link to the issue here; we should squash this with the other commit adding inputs.suite

lool (changes requested):
I left a broad set of comments; mostly minor

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: I agree, Ansible does support it. Maybe something that we can propose to upstream Debos, or at least report the need there in an issue as a first step.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: Yes, I was thinking of it but wanted to delay it to later. I will add a commit for it. :)

.github/workflows/debos.yml
gagath: Squashed, and reported in #379.

.github/workflows/debos.yml
gagath: Reverted to trixie only.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: done

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: good catch, done

lool (commented):
It was nice to have the "ap:" in the commit messages, but I assume you'll drop them before merging?

"Generate APT sources dynamically": commit message says -updates/-security is only for stable, but it's also for testing; it should say "except for unstable/sid"

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -83,9 +83,30 @@ actions:
 {{- end }}
       EOF
 
-  - action: overlay
-    description: Add Debian backports APT configuration
-    source: overlays/backports
+{{- if and (ne $suite "unstable") (ne $suite "sid") (ne $suite "testing") }}
```
lool, line 86: It's a bit weird to me too, but testing-backports exists!

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -116,6 +116,35 @@ actions:
...
 
+  # The qsc-deb-releases overlay only ships the keyring (.asc); generate the
+  # .sources and .pref files dynamically to use the correct $suite-overlay
+  - action: run
+    description: Create qsc-deb-releases APT sources for {{ $suite }}-overlay
+    chroot: true
+    command: |
+      set -eux
+      # only generate if the keyring was installed by the overlay
+      if [ -f /etc/apt/keyrings/qsc-deb-releases.asc ]; then
```
lool: How about flipping this around, if the file is missing, exit 0? avoids the extra code indentation for the two tee calls

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -203,6 +195,22 @@ actions:
       # WPA / WPA2 / WPA3 client support
       - wpasupplicant
 
+{{- if eq $suite "trixie" }}
+{{- if ne $overlays "none" }}
+{{- range $overlay := split "," $overlays }}
+{{- if eq $overlay "qsc-deb-releases" }}
+  # fastrpc-tests is only available in trixie-overlay
```
lool, line 206: I wonder if it would be simpler to just check with apt if the fastrpc-tests package is available, and install it, otherwise install fastrpc-support

.github/workflows/build-on-pr.yml
```diff
@@ -26,8 +26,15 @@ jobs:
       fail-fast: false
       matrix:
         suite: [trixie, forky]
+        kernelpackage: [linux-image-6.16.7-qcom1++, linux-image-arm64]
```
lool: This makes us hardcode this idea in two workflows and the linux-image-6.16.7 name in 3; how about we change debos to use "default" as default kernelname, and then add some logic in the debos recipe or debos workflow?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: I think we should not enable it then if weird. Less surprise is better. Keeping the current condition.

.github/workflows/debos.yml
gagath: after running zizmor on this file there is already the following security error on main:

```
error[template-injection]: code injection via template expansion
   --> .github/workflows/debos.yml:116:32
    |
105 |         run: |
    |         --- this run block
...
116 |               -t overlays:'${{ inputs.overlays }}' \
    |                                ^^^^^^^^^^^^^^^ may expand into attacker-controllable code
    |
    = note: audit confidence → High
```

If we want to use zizmor to validate our workflows, I would expect that we fix them first before asking maintainers to not introduce new issues. Do you run zizmor manually? Did you plan to create a workflow around it somehow? I would have expected this message to come from a CI pipeline, not manually from you. :)

gagath:
Removed all the [ap] tags. Pushed to close some conversations and have the CI run, still needs additional fixes before merging.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: done

.github/workflows/debos.yml
gagath: done

.github/workflows/build-on-pr.yml
gagath: I tried in a separate branch and to preserve current behavior we would have to introduce the following mappings:

- default -> linux-image-arm64 to preserve local behavior
- default-ci -> linux-image-6.16.7-qcom1++ to preserve ci behavior

I am not happy at all about this, I would prefer to have the linux-image-6.16.7-qcom1++ duplicated rather than having a "default-ci" match in the rootfs recipe. As long as we use a different kernel in CI than in local builds, I think this is acceptable (the duplication, not the fact of using a different kernel in CI).

gagath:
Rebased on top of main after #380 was merged.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: Agreed; for the record, I think people that use "forky" should get the backports because that will eventually be a concrete release track, but people that use "testing" should not get the backports because that's an always rolling suite. I can see why some tools might want to refer to testing-backports to not hardcode the name of testing in their logic, but I don't think we should enable backports for testing.

.github/workflows/build-on-pr.yml
lool: I didn't think about the fact we have different defaults in our recipes and our CI, good point.

I think our global default should be linux-image-arm64 right now, and linux-image-qcom or -qcom-next in the future (see pkg-linux-qcom work).

In order to land this PR, let's just add some default-ci handling in the debos github workflow (this is where I thought we'd initially handle that, but then I thought the debos recipe would be better).

.github/workflows/debos.yml
lool: I did run it manually in the past and wanted us to add automated runs: https://github.com/qualcomm-linux/qcom-deb-images/issues/166

I don't want to hold you to a zizmor-clean codebase, indeed it's not, I spotted this new line as something that was obviously making things worse from that perspective (I agree you couldn't have naturally thought about it), and I think we should not introduce new variables that are prone to injection when we can.

Defining an env would also avoid some duplication (see below).

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: Wanted to provide that in a separate branch on main, but it seems that fastrpc-support is not available in trixie nor trixie-backports: https://tracker.debian.org/pkg/fastrpc

> mandatory daemons like the adsp service you need on your hamoa laptop :)

As my Hamoa laptop is currently running trixie, I have no fastrpc package installed nor "cdsprpcd.service" or "adsprpcd.service" running. Is "mandatory" the correct word here? Maybe I am missing things out.

lool: I am pretty sure it's mandatory, perhaps for battery management or audio, not sure. We should discuss with team :)

Indeed, we don't have a backport, so I guess we must skip this for now; Robie wanted clarification on whether the daemons were required or not before landing a backport, we discussed this here:
https://github.com/qualcomm/fastrpc/issues/321

.github/workflows/debos.yml
gagath: Thanks for pointing me to the issue, was not aware of it. Pushed a new version to resolve this conversation and see how the CI goes with using the env variable. zizmor do not report any additional errors/warnings than what is already in main.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: done

.github/workflows/build-on-pr.yml
gagath: Introduced a "default-ci" special value that is handled in workflows for now. Will resolve this conversation after checking the CI run.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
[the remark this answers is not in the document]
gagath: I have shown that fastrpc-support cannot be installed on trixie when the qsc-deb-releases overlay is disabled, because of a missing backport. Hence the dynamic check on the group to see if it exists. Marking as resolved.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: Dropped the commit for a static if/else using templating. Note that our qartifactory answers 404 for "sid" so it does not consider it as an alias for "unstable". I think we should get rid of the "sid" support and ask users to use unstable instead, that would simplify some conditions in the recipe as well.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: Acknowledged, and skipping for now. Leaving the conversation open.

gagath:
I believe I have addressed all comments. I am still not sure if we should support "sid" as an alias for "unstable" as our qartifactory setup does not seem to support that.

lool (commented):
Looked great! Nice work!!

There's a build failure with forky, but we seem really close!

gagath:
@lool CI now passes. Since you did not provide an explicit approval I will let you review and merge at your convenience.

lool (approved):
All good; indeed I was waiting for the final fixes for the build issues, and I see you've addressed the missing parenthesis

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lumag: No, battery management and generic audio will work without those daemons.

PR #351, opened 2026-04-03
fix: support "none" value for $overlays

gagath:
The $overlays variable cannot easily be set to an empty string from the
command line or when calling the makefile. At least the following tries
do not work:

        make rootfs.tar EXTRA_DEBOS_OPTS="-t overlays:"
        make rootfs.tar EXTRA_DEBOS_OPTS="-t overlays:''"

Allow to set the overlays to "none" to be able to disable the otherwise
always enabled qsc-deb-releases overlay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,8 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
 - `overlays`: a `,`-separated list of rootfs overlays to add from
   `debos-recipes/overlays/`. Defaults to `qsc-deb-releases` to add our overlay
   apt repository that contains some package delta that isn't fully upstreamed
-  and backported to trixie in Debian yet.
+  and backported to trixie in Debian yet. Can be set to "none" to not include
+  any overlay.
 - `kernelpackage`: name of the kernel package to install from apt; defaults to
   `Debian’s linux-image-arm64`. Can (and should) be set to `none` if you are
   providing local kernel package instead.
--- a/README.md
+++ b/README.md
@@ -140,10 +140,7 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
 - `xfcedesktop`: install an Xfce desktop environment; default: console only
   environment
 - `overlays`: a `,`-separated list of rootfs overlays to add from
-  `debos-recipes/overlays/`. Defaults to `qsc-deb-releases` to add our overlay
-  apt repository that contains some package delta that isn't fully upstreamed
-  and backported to trixie in Debian yet. Can be set to "none" to not include
-  any overlay.
+  `debos-recipes/overlays/`. See the *Supported overlays* section below.
 - `kernelpackage`: name of the kernel package to install from apt; defaults to
   `Debian’s linux-image-arm64`. Can (and should) be set to `none` if you are
   providing local kernel package instead.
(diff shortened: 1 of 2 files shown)


debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -92,7 +92,13 @@ actions:
       # ext4 tools, notably e2fsck for the root filesystem
       - e2fsprogs
       # fastrpc support. fastrpc-tests pulls in the required stack
+{{- if ne $overlays "none" }}
+{{- range $overlay := split "," $overlays }}
+{{- if eq $overlay "qsc-deb-releases" }}
```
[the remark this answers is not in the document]
gagath: I think I prefer the ability to activate a selected amount of overlays rather than choose all-in or all-out regarding QLI customization. Its allows for more possibilities. However I agree that this may be a little bit confusing for users.

I propose we move on and revisit this later eventually. I pushed an extra commit that describes the overlays I know. The workaround overlays can be described later if needed.

gagath:
Updated with an additional commit to document common overlays in README. The fastrpc group cherry-pick was added because the group does not exist when fastrpc-tests is not installed.

gagath:
Rebased on top of main.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -92,7 +92,13 @@ actions:
...
       - fastrpc-tests
```
lool, line 98: can you add a placeholder that we should install fastrpc-support when there's a backport?

README.md
```diff
@@ -200,6 +198,34 @@ extra options to debos invocations, use `EXTRA_DEBOS_OPTS`, e.g.:
...
+Here is the list of supported overlays:
+
+<dl>
+    <dt>none</dt>
+    <dd>
+        Special value to disable all overlays.
+    </dd>
+    <dt>qsc-deb-releases</dt>
+    <dd>
+        Enable our overlay apt repository that contains some package delta that
```
lool, line 220: "Including this overlay will also select the fastrpc-test package"

README.md
```diff
@@ -200,6 +198,34 @@ extra options to debos invocations, use `EXTRA_DEBOS_OPTS`, e.g.:
...
+        isn't fully upstreamed and backported to trixie in Debian yet.
+    </dd>
+    <dt>backports</dt>
```
lool: Hmm we include this one unconditionally; should we list it in overlays instead? (I don't have a strong opinion.)

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: I am not sure what you want. I do not see a backport yet in Debian, so I say YAGNI and let’s introduce an additional change when the backport exists.

README.md
gagath: My bad, I do not think mandatory overlays should be described. Will remove it.

README.md
gagath: done

README.md
gagath: removed

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: I just wanted a comment but sure

lool (approved):

PR #354, opened 2026-04-06
fix(scripts/bundle): Pack NVMe-SPINOR tarballs separately

vishwasudupa:
Place SPINOR content under the NVME folder to align with the current content.xml layout.

Use _UFS and _NVME directories to distinguish storage types instead of inferring from partition conf.


lool (changes requested):
Hey, thanks for this!

Your changes look clean but they kind of go counter the current idea: we currently generate two tarballs, grouping platforms by the sector size they use. To do this, we're scanning generated files to see if the 512B or 4k sector size was used (disk-sdcard or -ufs).

Right now, nvme platforms end up in the emmc tarball because they use 512B sectors, and spinor ends up with them due to the default case – a bit of a hack.


What I would propose instead is to:
1) stop including spinor in emmc by default
2) for each _spinor dir, copy it to each of the corresponding _emmc, _ufs and _nvme directories with the same platform name

This avoids special casing NVMe and special cases spinor as a sign of a NHLOS storage.


I don't think you need to do it in this PR, but I'd like us to rename disk-sdcard/disk-ufs and flash-ufs/flash-emmc to better names related to the sector size.

lool:
We discussed this out of band with Vishwas:
- I pointed out I want to keep the two tarballs coming out of CI (one per disk sector size)
- perhaps we can add a yaml metadata entry to point out to the NHLOS ptool config to use (spinor for glymur-crd)
- Vishwas pointed out ptool has the information in partitions.conf about the sector size; this is great news, and we can use it to route files to one or the other (probably testing rawprogram0.xml to see which sector size it refers to)
- I haven't checked how the contents.xml.in in ptool currently expect things to be laid out, but I expect that it could look like:
  - flash_glymur-crd_nvme/
    - files to flash on the NVMe (disk-media.img1, disk-media.img2, rawprogram0.xml, patch0.xml...)
    - spinor/
      - files to flash on spinor (cdt.bin, uefi.bin, rawprogram0.xml, patch0.xml...)
    - contents.xml listing both
- and since it's likely going to use 512B sectors, I expect it will be zipped into flash-emmc.tgz for now

vishwasudupa:
> Hey, thanks for this!
> 
> Your changes look clean but they kind of go counter the current idea: we currently generate two tarballs, grouping platforms by the sector size they use. To do this, we're scanning generated files to see if the 512B or 4k sector size was used (disk-sdcard or -ufs).
> 
> Right now, nvme platforms end up in the emmc tarball because they use 512B sectors, and spinor ends up with them due to the default case – a bit of a hack.
> 
> What I would propose instead is to:
> 
> 1. stop including spinor in emmc by default
> 2. for each _spinor dir, copy it to each of the corresponding _emmc, _ufs and _nvme directories with the same platform name
> 
> This avoids special casing NVMe and special cases spinor as a sign of a NHLOS storage.
> 
> I don't think you need to do it in this PR, but I'd like us to rename disk-sdcard/disk-ufs and flash-ufs/flash-emmc to better names related to the sector size.

I have enabled spinor to be going to its respective directory in latest patch

scripts/bundle-flash-dirs.sh
```diff
@@ -43,6 +43,16 @@ done
 echo "emmc_dirs: $emmc_dirs"
 echo "ufs_dirs: $ufs_dirs"
 
+# Organize spinor directories under their base directories
+for dir in $emmc_dirs $ufs_dirs; do
+    [ -d "$dir" ] || continue
```
lool: I don't think this line is useful? we just assembled a list of directories, and you're checking if the directory is present?

scripts/bundle-flash-dirs.sh
```diff
@@ -43,6 +43,16 @@ done
...
+    echo "$dir" | grep -q "_spinor$" || continue
```
lool: So we loop through all dirs that mix emmc, ufs, nvme and spinor, just to spot the spinor ones; then we have to do another loop to see where to copy them to. That seems needlessly complicated.

You could either:
- do this earlier
- or check if the directory is **not** suffixed _spinor, and then see if there's a corresponding _spinor dir to copy into it

scripts/bundle-flash-dirs.sh
```diff
@@ -43,6 +43,16 @@ done
...
+    prefix=$(echo "$dir" | sed 's/_spinor$//')
+    for folder in $emmc_dirs $ufs_dirs; do
+        echo "$folder" | grep -q "^${prefix}_" && [ "$folder" != "$dir" ] && [ -d "$folder" ] && cp -r "${dir}"/* "${folder}/" && break
```
lool: The && chained commands are hard to read; also, you seem to stop the loop after the first cp (I guess that's the point of break), but what if we support nvme and ufs, and we want spinor in both?

lool (changes requested):
Hey, this seems needlessly complicated, there's already a top loop on all directories earlier and a simpler way to write this

My main issue though is that I think we could do that in the flash recipe itself, so that this behavior is the result of image builds, not just the result in our CI artifacts (this script is used from the github workflow).

I would think the same logic should work from the main flash recipe; **after** the $platforms loop; something like:
```
for dir in flash_*; do
    # skip directories that are named _spinor
    if [ -z "${d%%*_spinor}" ]; then
        continue
    fi
    # strip _storage suffix, and check for a _spinor directory
    spinor_dir="${d%_*}_spinor"
    if [ -d "$spinor_dir" ]; then
        # embed a copy of spinor contents
        cp ...preserve args... "$spinor_dir" "$d"
    fi
done
```

scripts/bundle-flash-dirs.sh
vishwasudupa: ok will remove it and simplify

scripts/bundle-flash-dirs.sh
vishwasudupa: yes in those cases need to maintain metadata somewhere to see which HLOS once get spinor, i am thinking id that info can be got via any xml/conf from ptool 

let me know if any place if this mapping can be found

vishwasudupa: I found a way to get the content.xml in spinor works with which other storage

I have used that logic in latest patch, this should resolve for case where there is both nvme/ufs

scripts/bundle-flash-dirs.sh
vishwasudupa: I have added in start and simplified code without && for better visibility

vishwasudupa:
> Hey, this seems needlessly complicated, there's already a top loop on all directories earlier and a simpler way to write this
> 
> My main issue though is that I think we could do that in the flash recipe itself, so that this behavior is the result of image builds, not just the result in our CI artifacts (this script is used from the github workflow).
> 
> I would think the same logic should work from the main flash recipe; **after** the $platforms loop; something like:
> 
> ```
> for dir in flash_*; do
>     # skip directories that are named _spinor
>     if [ -z "${d%%*_spinor}" ]; then
>         continue
>     fi
>     # strip _storage suffix, and check for a _spinor directory
>     spinor_dir="${d%_*}_spinor"
>     if [ -d "$spinor_dir" ]; then
>         # embed a copy of spinor contentsd
>         cp ...preserve args... "$spinor_dir" "$d"
>     fi
> done
> ```

I taught since this is bundling decision ,it was better to keep in bundle-flash-dirs.
alterations if needed, can it be done in future PR?

lool:
> > Hey, this seems needlessly complicated, there's already a top loop on all directories earlier and a simpler way to write this
> > My main issue though is that I think we could do that in the flash recipe itself, so that this behavior is the result of image builds, not just the result in our CI artifacts (this script is used from the github workflow).
> > I would think the same logic should work from the main flash recipe; **after** the $platforms loop; something like:
> > ```
> > for dir in flash_*; do
> >     # skip directories that are named _spinor
> >     if [ -z "${d%%*_spinor}" ]; then
> >         continue
> >     fi
> >     # strip _storage suffix, and check for a _spinor directory
> >     spinor_dir="${d%_*}_spinor"
> >     if [ -d "$spinor_dir" ]; then
> >         # embed a copy of spinor contentsd
> >         cp ...preserve args... "$spinor_dir" "$d"
> >     fi
> > done
> > ```
> 
> I taught since this is bundling decision ,it was better to keep in bundle-flash-dirs. alterations if needed, can it be done in future PR?

Let's fix it now; I think I went the extra mile to share how this could look like?

scripts/bundle-flash-dirs.sh
lool: We discussed this with Srinivas tonight and the challenge with partition names and directories which might be wrong in contents.xml; what I suggested is to fix is as follows:
- add logic at the time where we create a spinor/ subdir under flash_nvme to adjust the contents.xml
- this logic can scan for storage type = NVMe elements in the .xml, and fix their directory and filenames based on the nvme rawprogram.xml

you could do this with a python snippet (import xml from stdlib) or with xmlstarlet.

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -439,6 +439,40 @@ actions:
...
 {{- end }}
 
+      # copy each spinor flash dir into its parent storage flash dirs;
+      # storage types are read from product_info/chipid in contents.xml
+      for spinor_dir in "${ARTIFACTDIR}"/flash_*_spinor; do
+          contents_xml="${spinor_dir}/contents.xml"
+          [ -f "${contents_xml}" ] || continue
+          spinor_dir_name="$(basename "${spinor_dir}")"
+          board="${spinor_dir_name#flash_}"; board="${board%_spinor}"
+          xmlstarlet sel -t -m "//product_info/chipid" -v "@storage_type" -n "${contents_xml}" | \
```
lool, line 449: Interesting approach, I like the idea of reading the storage from contents.xml!

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -439,6 +439,40 @@ actions:
...
+              tr '[:upper:]' '[:lower:]' | sort -u | \
+              while read -r storage; do
+                  [ "${storage}" = "spinor" ] && continue
+                  target_dir="${ARTIFACTDIR}/flash_${board}_${storage}"
+                  [ -d "${target_dir}" ] || continue
+                  cp --preserve=all --no-dereference -av "${spinor_dir}" "${target_dir}/"
```
lool: I thought we wanted this to be flash_glymur-crd_nvme/spinor/contents.xml, as in the XML file, won't that make it flash_glymur-crd_nvme/flash_glymur-crd_spinor?

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -439,6 +439,40 @@ actions:
...
+                  # ufs -> disk-ufs, else -> disk-sdcard
```
lool: I don't want a second place where we maintain this map; I'd suggest extracting these from rawprogram0.xml instead:
```
foo.py:
import sys, xml.etree.ElementTree as ET
e = ET.parse(sys.argv[2]).getroot().find(f".//program[@label='{sys.argv[1]}']")
print(e.get("filename"))

% python3 foo.py rootfs rawprogram0.xml
disk-sdcard.img2
```

debos-recipes/qualcomm-linux-debian-flash.yaml
lool, line 449: So I'm surprised this built without us explicitly installing xmlstarlet somewhere, not sure why; could you apt install it in the github workflow ("Install debos and dependencies of the recipes and local tests")?

lool (changes requested):
Thanks for adding the logic in the right place; looks pretty good

I would have had a slight preference for a dedicated python script using stdlib XML parsing (no extra special dependency needed) over xmlstarlet, but I'm ok with xmlstarlet if that's what you prefer.

Could you
1) extract filenames (will give you file_path and file_name) from rawprogram0.xml? (see review comment for a python example)
2) install the xmlstarlet dependency explicitly

NB: we don't seem to have a section in the README for dependencies, perhaps we should though

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: main requirement would be content and rawprogram .xml to be able to access mentioned contents. since paths are ../ and ../../ even current aproach would work

i would also incline towards flash_glymur-crd_spinor because it tells it is flashing directory for which board and target type.

this might have issue if "spinor" is folder name mentioned somewhere in xml. and it is none that i know of, hence it should not be issue

lool: I find it a bit odd to repeat the board name under a subdirectory, and this flash_* top level dir convention is a qcom-deb-images one, not something widespread, so I'm keen that the contents of the folder look "normal" and familiar, similar to other images.

lool: I looked at other contents.xml.in files, and I see it's always relative pathnames for file_path, except for SAIL which typically uses:
```
<file_path flavor="sail_nor">./sail_nor</file_path>
```
So I think doing something similar for spinor makes sense.

vishwasudupa: updated to flash_glymur-crd_nvme/spinor in latest patch

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: interesting, yes, should not repeat coding, created a py file as suggested

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: I have updated in Readme,
i made change in debos.yaml i dont have permission to to update github workflows. so reverted it locally
please help to add in debos.yaml

scripts/get-rawprogram-filename.py
```diff
@@ -0,0 +1,29 @@
...
+import sys
+import xml.etree.ElementTree as ET
+
+if len(sys.argv) != 3:
+    print(f"Usage: {sys.argv[0]} <label> <rawprogram0.xml>", file=sys.stderr)
+    sys.exit(1)
+
+label, xml_file = sys.argv[1], sys.argv[2]
+
+root = ET.parse(xml_file).getroot()
```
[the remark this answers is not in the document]
vishwasudupa: pylint --errors-only ./ci/schemacheck.py ./scripts/get-disktype.py ./scripts/syft-license-summary.py ./scripts/build-deb.py ./scripts/run-qemu.py ./scripts/build-linux-deb.py
************* Module get-disktype
scripts/get-disktype.py:15:0: E0401: Unable to import 'defusedxml.ElementTree' (import-error)
Error: Process completed with exit code 123.

how to accommodate both ?

scripts/bundle-flash-dirs.sh
```diff
@@ -15,11 +15,12 @@ for d in flash_*
 do
     rawprogram0="$d/rawprogram0.xml"
     echo "examining $rawprogram0"
-    if grep disk-sdcard "$rawprogram0"
+    rootfs_img=$(python3 "$(dirname "$0")/get-disktype.py" rootfs "$rawprogram0")
```
lool: Can you drop python3 from the call? (Shebang should be enough)

lool (changes requested):
This is looking really good; extra kudos for updating bundle-flash-dirs to use the new script!

A few minor polish asks:
- can we call the new get-disktype script "get-rawprogram-filename.py"
- please drop python3 from the invocations (let's just make this script +x)
- please tweak the commit message to start sentences with a capital or use some kind of list, it's hard to parse the individual items

Then I think we will be able to do a final check on the resulting artifacts and merge

scripts/bundle-flash-dirs.sh
vishwasudupa: updated

vishwasudupa:
> This is looking really good; extra kudos for updating bundle-flash-dirs to use the new script!
> 
> A few minor polish asks:
> 
> * can we call the new get-disktype script "get-rawprogram-filename.py"
> * please drop python3 from the invocations (let's just make this script +x)
> * please tweak the commit message to start sentences with a capital or use some kind of list, it's hard to parse the individual items
> 
> Then I think we will be able to do a final check on the resulting artifacts and merge

I have updated
please help to add xmlscarlet and python3-defusedxml to debos.yml, i dont have permissions to do changes for that

scripts/get-rawprogram-filename.py
```diff
@@ -0,0 +1,30 @@
...
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# Print the filename attribute of a <program> element by label from a
+# rawprogram XML file generated by qcom-ptool.
+#
+# Usage: get-disktype.py <label> <rawprogram0.xml>
+#
+# Example:
+#   $ python3 get-disktype.py rootfs rawprogram0.xml
```
nivasnaik: get-disktype.py typo here?

vishwasudupa: yes, had missed it. resolved in latest patch

vishwasudupa:
> > This is looking really good; extra kudos for updating bundle-flash-dirs to use the new script!
> > A few minor polish asks:
> > 
> > * can we call the new get-disktype script "get-rawprogram-filename.py"
> > * please drop python3 from the invocations (let's just make this script +x)
> > * please tweak the commit message to start sentences with a capital or use some kind of list, it's hard to parse the individual items
> > 
> > Then I think we will be able to do a final check on the resulting artifacts and merge
> 
> I have updated please help to add xmlscarlet and python3-defusedxml to debos.yml, i dont have permissions to do changes for that

I was able to locally verify with inputs from [PR368 ](https://github.com/qualcomm-linux/qcom-deb-images/pull/368)

was able to locally install xmlscarlet and python3-defusedxml and verify 

folders and contents of content.xml is modified as expeted

@lool , need your help to enable xmlscarlet and python3-defusedxml as part of debos.yml, else help me get permissions to be able to contribute to debos.yml

lool:
> @lool , need your help to enable xmlscarlet and python3-defusedxml as part of debos.yml, else help me get permissions to be able to contribute to debos.yml

I'm not sure what permissions you need; do feel free to add that to your commit; is GitHub preventing you somehow?

vishwasudupa:
> > @lool , need your help to enable xmlscarlet and python3-defusedxml as part of debos.yml, else help me get permissions to be able to contribute to debos.yml
> 
> I'm not sure what permissions you need; do feel free to add that to your commit; is GitHub preventing you somehow?

digged deep down, it was my github token acess issue, fixed it and updated in latest

vishwasudupa:
> > > @lool , need your help to enable xmlscarlet and python3-defusedxml as part of debos.yml, else help me get permissions to be able to contribute to debos.yml
> > 
> > 
> > I'm not sure what permissions you need; do feel free to add that to your commit; is GitHub preventing you somehow?
> 
> digged deep down, it was my github token acess issue, fixed it and updated in latest

Build has also passed in CI now, all comments are addressed. @lool , please help to get it merged

lool:
https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/24491484065/job/71577228135?pr=354

lool:
@vishwasudupa if you check build status, you'll see one item is red, it's the static analysis for all scripts; I believe it's a matter of installing the new dep for that workflow run too; could you do that? You only need the python-* one.

vishwasudupa:
> @vishwasudupa if you check build status, you'll see one item is red, it's the static analysis for all scripts; I believe it's a matter of installing the new dep for that workflow run too; could you do that? You only need the python-* one.

thanks for pointing out, made change, the check is passing now

lool (approved):
Looks good! We will sort the new deps on the apt install lines alphabetically on the next opportunity

PR #362, opened 2026-04-08
Basic UKI and stubble support

gagath:
This is a proposal for a first basic support of UKIs with stubble and DeviceTreeAuto support.


debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,15 +77,82 @@ actions:
...
+          else
+            dt_auto="$dt_auto $dtb"
+          fi
+        done
+      done
+
+      tee /etc/kernel/uki.conf <<EOF
+      [UKI]
+      HWIDs=/usr/share/stubble/hwids/
+      DeviceTreeAuto=$dt_auto
```
lool: this will get out of date, how about we add a kernel install script that updates this line in the file with a comment that the line is automatically updated? these two could be provided by some stubble package in Debian?

lool (commented):
Cool! Left a few comments

I don't think we're in a rush to land this until we have figured where to do things correctly (otherwise we'll be distracted and will keep the technical debt); we could develop changes to systemd of ukify or kernel-install if needed

gagath:
Replaced the debos script with a kernel hook to generate a proper `/etc/kernel/uki.conf`. Still considered as a workaround since I think this will not handle multiple kernels quite well, but better than a one-time generation during recipe creation.

Still need to boot-test it on T14s.

debos-recipes/qualcomm-linux-debian-image.yaml
gagath: done as a kernel hook for now, can still be improved. will think of an upstreaming path once I get it to boot on real hardware.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,15 +77,82 @@ actions:
...
+    command: |
+      set -eux
+
+      # TODO: investigate glob support in systemd/src/ukify/ukify.py
+      # instead of this workaround that only works for a single image
+
+      dt_auto=""
+      for image in /lib/linux-image-*
+      do
+        for dtb in "${image}"/qcom/x1e78100*
```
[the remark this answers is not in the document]
gagath: Included "qcom/*" but that is not what you asked since many of these DTBs do not have a stubble entry. Will refine the script to only include ones that have a corresponding entry.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,15 +77,82 @@ actions:
...
       # TODO investigate why systemd-boot Recommends: shim-signed which
       # Depends on grub packages
       - shim-signed-
 
+  - action: apt
+    description: Install binutils to verify UKI generation
+    recommends: true
+    packages:
+      # provides objdump
+      - binutils
```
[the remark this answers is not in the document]
gagath: I think we could move this as an "image integration test" to check our image content maybe, or move it in the qemu test but seems overkill. Or I can plain remove it.

gagath:
After bricking my T14s, turns out that we cannot enable `qcom/*.dtb` because both `x1e78100-lenovo-thinkpad-t14s.dtb` and `x1e78100-lenovo-thinkpad-t14s-el2.dtb` match the compatible string. In my case the el2 variant was selected as it appeared first, leading to a boot failure. So we need to be careful which DT we want to include as there are multiple candidates with corresponding "compatible" string, but with different behavior.

I will do a match to remove "*el2.dtb" for now, on top on only including DTs compatible with the stubble DB.

EDIT: a similar fix [seems to be already present in stubble’s find_dtb script](https://github.com/ubuntu/stubble/blob/main/hwids/finddtbs.py#L26).

gagath:
New version with a kernel hook that I am now using on T14s. Generated image was boot-tested OK on T14s. Now to validate on Glymur CRD and have a look at the CI run.

For platforms that do not exist in stubble, `/etc/kernel/devicetree` can still be used to create a `.dtb` section in the UKI. Then the stub will replace this DT with another one only when an HWIDs match is found.

The stub reported by bootctl on T14s is systemd-stub, so concerns about not using stubble.efi can be ruled out.

gagath:
For the long term solution I think the best place for this hook would be in an additional binary package in stubble. We could also split the stubble source package to have the database and stub be shipped into different binary packages too. To be discussed with upstream once this hook is validated here.

gagath:
Fixed the flake8 line length issues. Still need to try this on Glymur.

gagath:
Tested OK on glymur:

 - on an existing install by installing the hook
 - as well as with the UKI produced by the image build

I pushed an extra commit to not pass the dtb option in the glymur ci job, but we probably can just use linux-daily-qcom-next.yml and drop the dedicated glymur job altogether. This would imply notifying users of the daily Glymur job to switch to the daily qcom-next build however, so maybe this is fine as a first step that causes minimal disruption.

Marking as ready for review.

lool:
> Tested OK on glymur:
> 
> * on an existing install by installing the hook
> * as well as with the UKI produced by the image build
> 
> I pushed an extra commit to not pass the dtb option in the glymur ci job, but we probably can just use linux-daily-qcom-next.yml and drop the dedicated glymur job altogether. This would imply notifying users of the daily Glymur job to switch to the daily qcom-next build however, so maybe this is fine as a first step that causes minimal disruption.
> 
> Marking as ready for review.

Please submit the new UKI generation script to the relevant upstream; I don't want to keep it in qcom-deb-images. The local scripts to select the kernel installation mode to use seem appropriate to keep here though.

For Glymur, I agree we can drop the DT override thanks to the stubble approach; should we recycle that setup to build Abel's branch instead?

lool (changes requested):
The debos-recipes and GH workflow changes are +1 for me; the UKI generation script should be in a package

gagath:
Started discussion in [Bug#1134185](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1134185) on the Debian BTS as I think it would be the appropriate place. I agree that we should recycle the ci setup for the `glymur-next` branch, will have a look at it.

gagath:
@lool While I figure out how to lend this upstream, what do you think of another PR (or amending this one) that would enable UKIs without stubble first? This would allow us to investigate any potential issue regarding UKI support on our platforms, and will use the `.dtb` section.

obbardc:
> @lool While I figure out how to lend this upstream, what do you think of another PR (or amending this one) that would enable UKIs without stubble first? This would allow us to investigate any potential issue regarding UKI support on our platforms, and will use the `.dtb` section.

Sounds like reasonable approach to me; landing it in two parts, UKI support then stubble support for auto-loading DT.

gagath:
> Sounds like reasonable approach to me; landing it in two parts, UKI support then stubble support for auto-loading DT.

Thanks for the appreciated feedback.

I have introduced #377 as a first step, which cherry-pick and rework the first commit to add UKI support only. Rebased this PR on top of #377. Let me know what you think!

gagath:
Rebased on top the uki branch to use tee instead of cat.

debos-recipes/overlays/workaround-ukify-devicetreeauto/etc/kernel/install.d/59-gen-ukify-conf.install
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python3
```
obbardc, line 1: add link to Debian patch https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1134185

and also probably worth opening a PR on stubble upstream http://github.com/ubuntu/stubble/

PR #366, opened 2026-04-13
fix(scripts/gen-ptool): Drop Axiom compat symlinks

lool:
Windows requires non-default permissions to unpack symlinks, so this is
not a mechanism we can scale. Instead, Axiom will be fixed to follow
relative pathnames in *.xml files.

### DIFF
--- a/scripts/gen-ptool.sh
+++ b/scripts/gen-ptool.sh
@@ -24,12 +24,12 @@ PARTITIONS_CONF="${QCOM_PTOOL}/platforms/${PLATFORM}/partitions.conf"
 
 case "$DISK_TYPE" in
   emmc|nvme)
-    esp="disk-sdcard.img1"
-    rootfs="disk-sdcard.img2"
+    esp="../disk-sdcard.img1"
+    rootfs="../disk-sdcard.img2"
     ;;
   ufs)
-    esp="disk-ufs.img1"
-    rootfs="disk-ufs.img2"
+    esp="../disk-ufs.img1"
+    rootfs="../disk-ufs.img2"
     ;;
   spinor)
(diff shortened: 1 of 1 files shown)


gagath (approved):

scripts/gen-ptool.sh
```diff
@@ -24,12 +24,12 @@ PARTITIONS_CONF="${QCOM_PTOOL}/platforms/${PLATFORM}/partitions.conf"
...
+    rootfs="../disk-sdcard.img2"
     ;;
   ufs)
-    esp="disk-ufs.img1"
-    rootfs="disk-ufs.img2"
+    esp="../disk-ufs.img1"
+    rootfs="../disk-ufs.img2"
     ;;
   spinor)
     # spinor carries firmware only; no OS efi/rootfs partitions
```
nivasnaik, line 35: to update contents.xml paths inside spinor folder, please add these. This is to update the contents.xml with correct path and also correct file names.
   esp="../../disk-ufs.img1"
   rootfs="../../disk-ufs.img2"

lool: I've actually asked @vishwasudupa to move the logic to copy spinor/ under flash_nvme to the flash recipe, after having run all the gen-ptool calls; I think this is where I'd expect the copy and the contents.xml fixup logic. See https://github.com/qualcomm-linux/qcom-deb-images/pull/354#discussion_r3074455516 and https://github.com/qualcomm-linux/qcom-deb-images/pull/354#pullrequestreview-4088106323

lool:
@vemana1niranjan confirms that this was flashed successfully via Axiom on a RB3Gen2, marking this as non-draft

gagath (approved):
Renewing approval.

PR #367, opened 2026-04-13
feat(ci): generate sha256sums.txt file

### LINKED ISSUE #320: ci: produced artifacts should be listed in a sha256sums.txt file
The produced artifacts are quite big and download mishaps can happen. We should have a sha256sums.txt file generated along the artifacts so that users can verify what they downloaded is correct.

gagath:
This allows to verify the correctness of the image files that can be quite big and bitrot over bad connections.

Closes: #320

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -148,6 +148,8 @@ jobs:
           cp -av dtbs.tar.gz "${dir}"
           # create tarballs with support for all UFS and all eMMC boards
           scripts/bundle-flash-dirs.sh "${dir}"
+          # generate sha256sums for all artifacts
+          (cd "${dir}" && sha256sum * | tee sha256sums.txt)
 
       # upload to a cloud storage space accessible by Axiom
       - name: Upload private artifacts (S3)


gagath:
Generated file looks fine and the sums match the produced artifacts in the CI run:

```console
$ sha256sum -c sha256sums.txt
disk-sdcard.img.gz: OK
...
```

Marking as ready for review.

lool (commented):
Looks good! I've checked build time, and the stage debos step seems to be up by 45s for a 14mn23s step, so about 5,2%, which I guess is acceptable, especially in the wider context.

I usually go for SHA256SUM, but since we have Windows users, sha256sums.txt makes sense.

Any thoughts on how to handle the SBOM files?

gagath:
> Any thoughts on how to handle the SBOM files?

The angle which I am taking here with introducing sha256sums.txt is to verify big downloads against bitrot to be sure nothing was lost during download, which is more likely as files get bigger. I did not add the SBOM files to the list because I consider them small (biggest one is 5 MiB) and the impact of a bitflip is way less than "image not booting" regarding use.

While I think there is less value for doing it, I can add the SBOM files but they are so small I doubt I or other users will ever verify them against their sha256sums.txt so I feel it is not necessary in the first place.

lool (approved):

gagath:
Thanks.

PR #368, opened 2026-04-14
chore(debos): add mtools to debos container image

vemana1niranjan:
While runnign make flash on arm dev machines getting error mtools to be installed 
So following this change to get image with mtools in it 

# Problem
Devs facing issue in using "CONTAINER_IMAGE ?= ghcr.io/go-debos/debos:latest"
Error message says mcopy cant be used 

Solution Creating Container image with mtools installed
step1: create a Docker file
commandd:cat <<'EOF' > Dockerfile.debos-mtools
FROM ghcr.io/go-debos/debos:latest

RUN apt-get update && \
    apt-get install -y mtools && \
    apt-get clean && rm -rf /var/lib/apt/lists/*
EOF
step 2:build image 
command:sudo docker build -t debos-with-mtools -f Dockerfile.debos-mtools .
step 3:Change Make file line 17
old line:CONTAINER_IMAGE ?= ghcr.io/go-debos/debos:latest
new line :CONTAINER_IMAGE ?= debos-with-mtools

Now devs can use command "make flash" to get flash artifacts

### DIFF
--- a/Dockerfile.debos-mtools
+++ b/Dockerfile.debos-mtools
@@ -1,0 +1,5 @@
+FROM ghcr.io/go-debos/debos:latest
+
+RUN apt-get update && \
+    apt-get install -y mtools && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
--- a/README.md
+++ b/README.md
@@ -281,6 +281,27 @@ Notes:
 
 We'd love to hear if you run into issues or have ideas for improvements. [Report an Issue on GitHub](../../issues) to discuss, and try to include as much information as possible on your specific environment.
 
+# Problem
+Devs facing issue in using "CONTAINER_IMAGE ?= ghcr.io/go-debos/debos:latest"
+Error message says mcopy cant be used 
+
+Solution Creating Container image with mtools installed
+step1: create a Docker file
+commandd:cat <<'EOF' > Dockerfile.debos-mtools
+FROM ghcr.io/go-debos/debos:latest
+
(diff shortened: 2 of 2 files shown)


lool (changes requested):
Hi Niranjan,

First of all, thanks for the PR! I think this is a real issue and it will keep getting worse as we're adding more dependencies to the flash recipe in upcoming pull requests, and I think @abickett mentioned that issue to me too (he contributed support for running debos from a local container).

Now on your PR:
1) your git setup needs fixing, as evidenced by your Signed-off-by
2) the README changes are weird, is this what you would expect to find as an user, a description of a problem and its solution, out of the blue?
3) on the overall approach, I'd prefer doing this:
- create a small script that checks for and installs missing dependencies; you can find the list in the debos github workflow
- use it from the Makefile (perhaps `check-deps` or `doctor`?)
- use the make target from the debos github workflow
- arrange to use it on top of the standard debos docker image (e.g. run the script unconditionally every time)

Dockerfile.debos-mtools
```diff
@@ -0,0 +1,5 @@
+FROM ghcr.io/go-debos/debos:latest
+
+RUN apt-get update && \
+    apt-get install -y mtools && \
```
vishwasudupa, line 4: please add 

        xmlstarlet \
        python3-defusedxml \

might need it post merge of PR (354)

vemana1niranjan:
#371 is solving same so closing this

PR #369, opened 2026-04-14
workflows: upgrade lava-action to v11

mwasilew:
lava-action@v11 moves to node24 in place of deprecated node20. This will remove the warning when workflow is executed.


lool (approved):
Looks good

PR #370, opened 2026-04-17
fix(get-rawprogram-filename): Return none when missing

vishwasudupa:
Return none when the input XML does not contain a rootfs entry or
when parsing fails, instead of erroring out. This allows callers to
distinguish between valid matches and missing data gracefully.


vishwasudupa:
> Return None when the input XML does not contain a rootfs entry or when parsing fails, instead of erroring out. This allows script to exit gracefully.

this would help to fix regression caused in latest nightly build

lool:
@copilot review this

lool (approved):

nivasnaik:
> @copilot review this

Copilot did not highlight this tough :)

lool:
> > @copilot review this
> 
> Copilot did not highlight this tough :)

I keep trying from time to time, hoping we'll fix it :D

PR #373, opened 2026-04-17
Fix silent pipeline failures

lool:
Fix real and potential silent pipeline failures
- **debos: Fix silent failure with missing xmlstarlet**
- **fix(debos): Pre-emptively fix silent failures**

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -446,8 +446,9 @@ actions:
           [ -f "${contents_xml}" ] || continue
           spinor_dir_name="$(basename "${spinor_dir}")"
           board="${spinor_dir_name#flash_}"; board="${board%_spinor}"
-          xmlstarlet sel -t -m "//product_info/chipid" -v "@storage_type" -n "${contents_xml}" | \
-              tr '[:upper:]' '[:lower:]' | sort -u | \
+          # extract storage types to variable to avoid silent pipeline failure
+          storage_types=$(xmlstarlet sel -t -m "//product_info/chipid" -v "@storage_type" -n "${contents_xml}")
+          echo "$storage_types" | tr '[:upper:]' '[:lower:]' | sort -u | \
               while read -r storage; do
                   [ "${storage}" = "spinor" ] && continue
                   target_dir="${ARTIFACTDIR}/flash_${board}_${storage}"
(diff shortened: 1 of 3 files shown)


gagath (changes requested):
The correct term is "pipe", not "pipeline".

Moreover, this is why the -o pipefail option exist. Your changes do not check for other pipe failures but only the ones you extract, which is a shortcoming. EDIT: the commit mentions that this is because `/bin/sh` might be something else than `dash`, while `dash` does support the option. Nevermind. I guess this is why big scripts should probably be extracted at some point.

The commits messages are overly verbose IMO; the second commit does not respect the conventional commit specs that we seem to use in the project. EDIT: I see why I feel this: the second paragraph is an unperfect copy paste. I guess this is a good signal that this should be one commit, not two.

lool:
> The correct term is "pipe", not "pipeline".

From bash's manpage:
```
       A pipeline is a sequence of one or more commands separated by the
       character |.  The format for a pipeline is:

              [time [-p]] [ ! ] command [ | command2 ... ]

       The standard output of command is connected via a pipe to the standard
       input of command2.  This connection is performed before any
       redirections specified by the command (see REDIRECTION below).

       The return status of a pipeline is the exit status of the last command,
       unless the pipefail option is enabled.  If pipefail is enabled, the
       pipeline's return status is the value of the last (rightmost) command
       to exit with a non-zero status, or zero if all commands exit
       successfully.  If the reserved word !  precedes a pipeline, the exit
       status of that pipeline is the logical negation of the exit status as
       described above.  The shell waits for all commands in the pipeline to
       terminate before returning a value.
```

> Moreover, this is why the -o pipefail option exist. Your changes do not check for other pipe failures but only the ones you extract, which is a shortcoming. EDIT: the commit mentions that this is because `/bin/sh` might be something else than `dash`, while `dash` does support the option. Nevermind. I guess this is why big scripts should probably be extracted at some point
> 
> The commits messages are overly verbose IMO; the second commit does not respect the conventional commit specs that we seem to use in the project. EDIT: I see why I feel this: the second paragraph is an unperfect copy paste. I guess this is a good signal that this should be one commit, not two.

So squash in one commit?

lool:
Squashed and tweaked commit message!

gagath (approved):
> From bash's manpage:

TIL

> Squashed and tweaked commit message!

LGTM

PR #374, opened 2026-04-17
refactor(get-rawprogram-fix): Misc cleanups

lool:
Simplify code nesting/depth, update description and exit with an
error when called with missing/too many arguments (doesn't affect
current usage).
- **fix(get-rawprogram-filename)!: Fail on incorrects args**
- **refactor(get-rawprogram-filename): Simplify logic**
- **fix(get-rawprogram-filename): Update description**


scripts/get-rawprogram-filename.py
```diff
@@ -16,17 +19,18 @@
 
 if len(sys.argv) != 3:
     print(f"Usage: {sys.argv[0]} <label> <rawprogram0.xml>", file=sys.stderr)
+    sys.exit(1)
```
nivasnaik, line 22: A non-zero return value will cause failure in bundle-flash-dirs.sh?

lool: Only if the number of arguments is wrong, which it won't be?

gagath (commented):
Not a fan of the "misc cleanups" first line in commit message and grouped changes. I would have preferred one-liner desc, focused commits, one for exit, one for flow, one for description.

lool:
> Not a fan of the "misc cleanups" first line in commit message and grouped changes. I would have preferred one-liner desc, focused commits, one for exit, one for flow, one for description.

Fair enough, I've split in 3 commits as you suggested

gagath (approved):
Thanks, LGTM

PR #375, opened 2026-04-20
feat(build-linux-deb): refactor

gagath:
Refactor build-linux-deb.py to allow for future extension.

### DIFF
--- a/scripts/build-linux-deb.py
+++ b/scripts/build-linux-deb.py
@@ -13,14 +13,17 @@ GIT_UPSTREAM = {
     "linux": {
         "repo": "https://github.com/torvalds/linux",
         "ref": "master",
+        "ref_prefix": None,
     },
     "linux-next": {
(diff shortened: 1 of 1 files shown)


lool:
Hey, I don't know if it makes sense to list every more or less short-lived tree like that; linux-next and qcom-next are quite permanent, and also require custom logic to select the right tag. The glymur tree has been 3 different trees in the last month or so...

gagath:
Fair enough, we can pass the URL and ref to it as args in the CI. I have dropped the last commit and renamed the PR to keep the refactoring commits as I think they are still valuable.

lool (approved):
Commits looked good; could you update PR description before merging?

The dict makes sense, there's another dimension which is the ref_prefix, do you want to add that?

gagath:
> Commits looked good; could you update PR description before merging?

Done.

> The dict makes sense, there's another dimension which is the ref_prefix, do you want to add that?

Done. Introduced an `upstream_git_key` variable to simplify the `if/elif` block a bit.

Will merge after the CI run.

PR #377, opened 2026-04-21
feat!(image): use UKI instead of kernel+initramfs

gagath:
Use systemd-ukify to create an UKI .efi file instead of the classic "bls" approach, which splits kernel and initramfs into separate files.

It can be discovered and booted by systemd-boot directly without the need for any entry. However this means than editing the command line and/or device-tree takes more step as one as to regenerate the UKI after changing /etc/kernel/commandline and/or /etc/kernel/devicetree resp. and dpkg-reconfigure the kernel to call the ukify install hooks.

This change is needed as a first step for using stubble to have .dtbauto sections and ultimately be able to autoselect a devicetree for supported platforms.

cherry-picked and reworked from #362

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-image.yaml
+++ b/debos-recipes/qualcomm-linux-debian-image.yaml
@@ -77,11 +77,21 @@ actions:
     setup-fstab: true
     append-kernel-cmdline: clk_ignore_unused pd_ignore_unused audit=0 deferred_probe_timeout=30
 
+  - action: run
+    description: Create /etc/kernel/install.conf file
+    chroot: true
+    command: |
+      set -eux
+      tee /etc/kernel/install.conf <<EOF
+      layout=uki
+      EOF
+
   - action: apt
     description: Make system bootable with systemd-boot
     recommends: true
     packages:
       - systemd-boot
+      - systemd-ukify
       # TODO investigate why systemd-boot Recommends: shim-signed which
       # Depends on grub packages
       - shim-signed-
--- a/debos-recipes/qualcomm-linux-debian-image.yaml
+++ b/debos-recipes/qualcomm-linux-debian-image.yaml
@@ -88,11 +88,21 @@ actions:
       apt-get update
 {{- end }}
 
+  - action: run
+    description: Ask for UKI via /etc/kernel/install.conf
+    chroot: true
+    command: |
+      set -eux
+      tee /etc/kernel/install.conf <<EOF
+      layout=uki
+      EOF
+
   - action: apt
     description: Make system bootable with systemd-boot
     recommends: true
     packages:
       - systemd-boot
+      - systemd-ukify
       # TODO investigate why systemd-boot Recommends: shim-signed which
       # Depends on grub packages
       - shim-signed-


gagath:
Pushed a small fix that uses `tee` instead of `cat` so we can see the content of the file in CI logs.

lool (commented):
So generally ok with the implementation of the recipe changes if we want to move to UKIs; I'd like to take a step back to make an impact assessment. We should probably broadcast this a bit to get more eyes on the functional change too.

Now my own impact assessment on this:

First, I use UKIs to boot my snapdragon (hamoa) work laptop; that works fine (wee!).

We have or want to have support for the following boot architectures:
1) boot firmware loads a platform specific DT from a platform specific dtb.bin partition (current on almost all platforms)
2) per platform u-boot loads newer DT from ESP (RB1, UNOQ); this will disappear in favor of 1) once we complete the move to the newly released Agatti boot firmware
3) systemd-boot stub loads an HLOS provided DT override based on boot
4) boot firmware loads a platform specific DT and configurable overlays from a generic database in the dtb.bin partition

I don't think switching to UKIs hurts any of the above. For glymur builds, we also force the DT in the systemd-boot config, and I think that would be picked up in the UKI world as well.

Partition size: I think this is pretty much the same, we're having larger UKI EFI files instead of discrete kernel + initrd, but size should be similar.

Last but not least: secure boot: debian kernels are signed, but our daily builds aren't. We do want to support secure boot, and it's part of the functional requirements for pkg-linux-qcom, but for a somewhat distant future. If we build the UKI at image creation/package installation time, I don't see how we'll deliver secure boot.

I'm not sure if there are other angles to consider on UKIs vs no-UKIs. I expect Debian to remain on GRUB + kernel image + initrd for the foreseeable future, so we should also test that somehow, but I'm personally ok to go towards UKI in qcom-deb-images QLI builds.

gagath:
This will have to be reworked to use type 2 UKIs instead of type 3. Moving back to draft.

gagath:
I have rebased this and reworked the commit message a bit. Regarding secure boot, the UKI can be signed the same way kernel could be signed if we decide to do it at some point. We would loose the signature as soon as we update the kernel or initramfs from the installed OS, but there is a possible workaround by shipping updated kernel or initramfs via a .addon.efi PE file that can eventually be signed too.

I think this is still a bit far away to reason about. Since having UKIs and .dtbauto would allow us to reduce the number of specialized builds we do, let’s focus on this first and revisit the secure boot story when it present itself in a more concrete way.

lool (changes requested):
So I'm trying to think of the implications again; what do we gain in enabling full UKIs immediately? IIUC, we want to do light UKIs, where we essentially tack an optional DTB loader in front of the kernel image, not full UKIs with an initramfs.

There's also no DTB loading here, since it's not supported upstream and we don't have a stubble-data replacement.

Sorry, but this feels like a step in the wrong direction, pulling us further away from Debian.

lool:
I think we should either share arguments on why we should diverge from Debian on kernel image formats, or close this PR

gagath:
I agree, UKIs make sense for an installer but not really for installed systems at the moment; let’s close this and revisit when we have a proper need.

PR #378, opened 2026-04-22
workflows: enable LAVA testing on Glymur

mwasilew:
Add boot/smoke test template for Glymur and enable LAVA testing in daily Glymur build.

### DIFF
--- a/ci/lava/glymur-crd/boot.yaml
+++ b/ci/lava/glymur-crd/boot.yaml
@@ -1,0 +1,85 @@
+actions:
+- deploy:
+    images:
+      image:
+        headers:
+          Authorization: Q_S3_TOKEN
+        url: "{{BUILD_DOWNLOAD_URL}}/{{SUITE}}-flash-emmc.tar.gz"
+    postprocess:
+      docker:
+        image: ghcr.io/foundriesio/lava-lmp-sign:main
+        steps:
+        - export IMAGE_PATH=$PWD
+        - cp overlay*.tar.gz overlay.tar.gz
+        - echo "OVERLAY=overlay.tar.gz" >> $IMAGE_PATH/flash.settings
+        - echo "OVERLAY_PATH=/home/" >> $IMAGE_PATH/flash.settings
+        - echo "STORAGE=emmc" >> $IMAGE_PATH/flash.settings
+        - echo "ROOTFS_IMAGE=disk-sdcard.img2" >> $IMAGE_PATH/flash.settings
+        - echo "DEVICE_TYPE=debian-glymur-crd_nvme" >> $IMAGE_PATH/flash.settings
+        - cat $IMAGE_PATH/flash.settings
+    timeout:
+      minutes: 200
+    to: downloads
+- deploy:
+    images:
+      image:
+        url: downloads://{{SUITE}}-flash-emmc.tar.gz
+      settings:
+        url: downloads://flash.settings
+      overlay:
+        url: downloads://overlay.tar.gz
+    timeout:
+      minutes: 20
+    to: flasher
+- boot:
+    auto_login:
+      login_prompt: 'login:'
+      username: debian
+      password_prompt: 'Password'
+      password: debian
+      login_commands:
+      - "debian"
+      - "new password"
+      - "new password"
+      - sudo su
+    method: minimal
+    prompts:
+    - root@debian
+    - debian@debian
+    - "Current password"
+    - "New password"
+    - "Retype new password"
+    timeout:
+      minutes: 3
+- test:
+    timeout:
+      minutes: 5
+    definitions:
+    - from: git
+      name: "smoke-test"
+      path: automated/linux/smoke/smoke.yaml
+      repository: https://github.com/linaro/test-definitions.git
+      parameters:
+        SKIP_INSTALL: "True"
+        TESTS: "pwd, uname -a, ip a"
+    - from: git
+      name: "wlan-smoke-test"
+      path: automated/linux/wlan-smoke/wlan-smoke.yaml
+      repository: https://github.com/linaro/test-definitions.git
+      parameters:
+        SKIP_INSTALL: "True"
+        WAIT_TIME: 60
+- command:
+    name: network_turn_on
+context:
+  lava_test_results_dir: /home/lava-%s
+  test_character_delay: 10
+device_type: glymur-crd
+job_name: smoke tests {{GITHUB_REPOSITORY}} {{GITHUB_RUN_ID}}-{{GITHUB_RUN_ATTEMPT}}
+metadata:
+  build-commit: '{{GITHUB_SHA}}'
+priority: 50
+timeouts:
+  job:
+    minutes: 210
+visibility: public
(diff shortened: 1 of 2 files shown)


mwasilew:
I'll mark it as draft for the moment as there are no devices available in "production" LAVA. There is one in  staging. Example job: https://lava-staging.infra.foundries.io/scheduler/job/900

lool:
The glymur build currently can't boot on anything else than glymur (it hardcodes the DT name to load on boot), so if we send test jobs for that daily for RB1 and RB3Gen2, they will fail.

The other way around might work though: perhaps the qcom-next build does works on glymur-crd, it should, but Agathe is on a different boot firmware so I'm not sure.

gagath:
> The other way around might work though: perhaps the qcom-next build does works on glymur-crd

I do not see how it can work without setting the dtb to glymur-crd.dtb at some point? But maybe I am too unfamiliar with our CI and LAVA setup.

> Agathe is on a different boot firmware so I'm not sure.

Flashing the different boot firmware and try to boot the daily-glymur image on top of it is something I can do. But as discussed other people working on Glymur are not using this boot firmware at the moment. We need to make a decision on what to use in CI and be aware of the difference between CI and what people actually use.

lool:
> > The other way around might work though: perhaps the qcom-next build does works on glymur-crd
> 
> I do not see how it can work without setting the dtb to glymur-crd.dtb at some point? But maybe I am too unfamiliar with our CI and LAVA setup.

In the Linux version of the boot firmware, we support loading the DT from the dtb.bin partition.

> Flashing the different boot firmware and try to boot the daily-glymur image on top of it is something I can do. But as discussed other people working on Glymur are not using this boot firmware at the moment. We need to make a decision on what to use in CI and be aware of the difference between CI and what people actually use.

I think for upstream linux development and getting a kernel that works, it is the right decision to track the Windows version of the firmware and its quirks. However, it's not currently suitable for our CI because there are no public releases of it (AFAIK); also, the convergence plans are not clear.

I'll put another coin in our internal discussions on this, perhaps we need to have two CI jobs with both firmwares?

mwasilew:
> I think for upstream linux development and getting a kernel that works, it is the right decision to track the Windows version of the firmware and its quirks. However, it's not currently suitable for our CI because there are no public releases of it (AFAIK); also, the convergence plans are not clear.
> 
> I'll put another coin in our internal discussions on this, perhaps we need to have two CI jobs with both firmwares?

I only flashed spinor once with output from meta-qcom build. I assume this is the same as used in qcom-deb-images. Is there some other firmware available?

Note: hopefully this will change in a few weeks with LAVA support for multistage flashing. But it will also mean that flashing takes longer.

lool:
> I only flashed spinor once with output from meta-qcom build. I assume this is the same as used in qcom-deb-images. Is there some other firmware available?

meta-qcom and qcom-deb-images are enabling glymur-crd with the only boot firmware that is public, which is a Linux-enabled boot firmware, but it's quite different from the boot firmware for Windows on Arm laptops. Upstream kernel developers are currently using a Windows on Arm boot firmware for Glymur CRD.

> Note: hopefully this will change in a few weeks with LAVA support for multistage flashing. But it will also mean that flashing takes longer.

This is something I've been wanting to raise to you and to the Axiom team: the practice of flashing SPINOR on every CI run seems bad, it's very long and I suspect it's causing unacceptable wear to the device.

How would you feel about implementing caching to skip reflashing some of the partitions that have already been flashed? We'd have to be careful about that because booted devices could alter the content of this storage, so perhaps it's a carefully crafted whitelist of files and job types that would benefit from that.

mwasilew:
> > Note: hopefully this will change in a few weeks with LAVA support for multistage flashing. But it will also mean that flashing takes longer.
> 
> This is something I've been wanting to raise to you and to the Axiom team: the practice of flashing SPINOR on every CI run seems bad, it's very long and I suspect it's causing unacceptable wear to the device.

It's not done at the moment. My plan is to provide ability to do so, not mandate it.

> 
> How would you feel about implementing caching to skip reflashing some of the partitions that have already been flashed? We'd have to be careful about that because booted devices could alter the content of this storage, so perhaps it's a carefully crafted whitelist of files and job types that would benefit from that.

I don't like this idea from the reproducibility point of view. I would like to think that each test should start from a good, known state. You can't be sure what was flashed to the device at the beginning of the job. If you want to avoid flashing (I agree wear might become an issue) there should be a way to verify that partition contents are as expected.

lool:
(@mwasilew FYI while the title of the PR says draft, it's not in Draft state)

I see we have [4 devices in our prod LAVA](https://lava.infra.foundries.io/scheduler/device_type/glymur-crd), are we good to land this? :)

I guess you made changes to the flasher to also flash SPI-NOR, or is this only flashing NVMe? I'm fine either way, I'm just keen to know which mode we're in

mwasilew:
> (@mwasilew FYI while the title of the PR says draft, it's not in Draft state)
> 

I'm not great with github interface. I'll try to update it

> I see we have [4 devices in our prod LAVA](https://lava.infra.foundries.io/scheduler/device_type/glymur-crd), are we good to land this? :)

Naresh added them today. I'll take a look tomorrow.

> 
> I guess you made changes to the flasher to also flash SPI-NOR, or is this only flashing NVMe? I'm fine either way, I'm just keen to know which mode we're in

No. Only NVMe is flashed. SPI-NOR is kept at a fixed version. There are a few options here: 
 - flash spinor with every job (slow and may wear the memory)
 - keep it at a known fixed version (like it is now) and only update manually when firmware changes
 - find a different solution: https://github.com/linux-msm/qdl/issues/221

gagath:
See https://github.com/qualcomm-linux/qcom-deb-images/pull/385 for enabling forky testing in the linux.yml workflow and consequently the daily builds.

mwasilew:
@lool I rebased on top of `main`. I think this is ready from my pov.

ci/lava/glymur-crd/boot.yaml
```diff
@@ -0,0 +1,85 @@
+actions:
+- deploy:
+    images:
+      image:
+        headers:
+          Authorization: Q_S3_TOKEN
+        url: "{{BUILD_DOWNLOAD_URL}}/flash-emmc.tar.gz"
```
lool: `url: "{{BUILD_DOWNLOAD_URL}}/{{SUITE}}-flash-emmc.tar.gz"`

ci/lava/glymur-crd/boot.yaml
```diff
@@ -0,0 +1,85 @@
...
+        - echo "ROOTFS_IMAGE=disk-sdcard.img2" >> $IMAGE_PATH/flash.settings
+        - echo "DEVICE_TYPE=debian-glymur-crd_nvme" >> $IMAGE_PATH/flash.settings
+        - cat $IMAGE_PATH/flash.settings
+    timeout:
+      minutes: 200
+    to: downloads
+- deploy:
+    images:
+      image:
+        url: downloads://flash-emmc.tar.gz
```
lool: `url: downloads://{{SUITE}}-flash-emmc.tar.gz`

lool (changes requested):
@mwasilew We've renamed artifacts to have trixie and forky versions; please see review comments and other files

lool (approved):

PR #382, opened 2026-04-24
Debian unstable memory debug

gagath:
Do not merge, for testing CI runners behavior only.


gagath:
@lool this seems to be a runner OOM rather than a fakemachine OOM, so we have likely to bump the memory of the runners. I see three of them that can run the build-debos job:

```
    runs-on: [self-hosted, qcom-u2404, arm64]
```

Do you know the memory configuration of each runner, and if they can be upped somehow? Thanks.

lool:
> Do you know the memory configuration of each runner, and if they can be upped somehow? Thanks.

I don't but I'll loop you in a thread where I'll ask!

lool:
> > Do you know the memory configuration of each runner, and if they can be upped somehow? Thanks.
> 
> I don't but I'll loop you in a thread where I'll ask!

-  vCPUs 16
- Memory (GiB) 64

So probably not a limit of physical memory size :)

PR #390, opened 2026-05-01
LAVA boards per workflow

lool:
Include/exclude boards from LAVA per workflow:
- **ci: lava: add boards include/exclude options**
- **ci: linux: add boards include/exclude**
- **ci: include/exclude boards from LAVA per workflow**


mwasilew (approved):
This seems to work well.

PR #395, opened 2026-05-04
fix(debos.yml): prefix sha256sums.txt files

### LINKED ISSUE #388: Conflicting sha256sums.txt uploads
Hi

Since landing support for multi-suites builds (trixie + forky), a matrix workflow will arrange for trixie and forky artifacts to be uploaded to the same overarching output artifacts dir on S3.

The sha256sums.txt file there only lists files for one or the other series (I'm not sure if first uploaded or last uploaded files wins).

We need to either fix the sha256sums.txt to list all files, or rename them to not collide (probably easier given current design).

gagath:
Now that the debos.yml workflow is called multiple times to test both Debian trixie and forky, concatenate the generated sha256sums.txt files instead of erasing them so all files are listed instead of the ones from the last running workflow.

Closes: #388

Created as a draft to verify the CI results.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -157,7 +157,7 @@ jobs:
           # create tarballs with support for all UFS and all eMMC boards
           scripts/bundle-flash-dirs.sh "${dir}" "${PREFIX}"
           # generate sha256sums for all artifacts
-          (cd "${dir}" && sha256sum * | tee sha256sums.txt)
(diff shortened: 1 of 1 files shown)


lool:
The builds are happening concurrently on two different runners, so you won't have a local file to concatenate to

gagath:
The files are now prefixed as concatenation do not work as you explained. I checked the file contents and they look as expected. Marking as ready for review.

lool (approved):

PR #396, opened 2026-05-04
feat(kernel-configs): add qcom-laptops config

lool:
Fix T14s Gen6 not booting with upcoming M.2 changes which require a
new config not yet in defconfig.

### DIFF
--- a/kernel-configs/qcom-laptops.config
+++ b/kernel-configs/qcom-laptops.config
@@ -1,0 +1,3 @@
+# needed for M.2 pwrseq driver
+# https://lore.kernel.org/linux-pci/20260501164440.11788-1-manivannan.sadhasivam@oss.qualcomm.com/
+CONFIG_POWER_SEQUENCING_PCIE_M2=m


gagath:
@lool I checked the Debian kernel tree and the Kconfig to enable POWER_SEQUENCING_PCIE_M2 does not seem to exist yet. My understanding is that this will come in a later kernel version. How do we plan to enable this in the Debian kernel as soon as it land? Are T14s users expected to have their laptop not boot anymore when the driver appears in the Debian kernel and is not enabled by default?

The change is OK as-is, but my worry is that this will break the laptop of the Debian kernel on T14s when the driver lands.

lool:
> @lool I checked the Debian kernel tree and the Kconfig to enable POWER_SEQUENCING_PCIE_M2 does not seem to exist yet. My understanding is that this will come in a later kernel version. How do we plan to enable this in the Debian kernel as soon as it land? Are T14s users expected to have their laptop not boot anymore when the driver appears in the Debian kernel and is not enabled by default?

I might not be the best person to convey the upstream kernel development dynamics; I did ask Ben if there's any kind of defconfig review on new upstream kernel releases, and he said that's not really happening, but that Aurélien is updating the configs at least for RISC-V. My current idea is that we propose a merge request relaying new arm64 defconfig changes to the Debian config, or at least the Qualcomm related ones. I've heard that's how the Fedora kernel team does it.

> The change is OK as-is, but my worry is that this will break the laptop of the Debian kernel on T14s when the driver lands.

When the DT changes land, yeah, that was my experience indeed (see chat with Abel and Mani).

gagath (approved):
Thanks for the explanation. At least we are aware of the situation.

gagath:
Follow-up: I just opened this PR to enable the option in Debian’s 7.1~rc3 experimental kernel:
https://salsa.debian.org/kernel-team/linux/-/merge_requests/1946

PR #398, opened 2026-05-06
Update unique job name for linux-daily-glymur.yml to invoke coral Workflow with unique checkrun completion

phanendra-qc:
Coral workflow invoke based on checkrun passed from Github https://qswat.qualcomm.com/coral/qli-open-development/workflows/Debian_CRM_Nightly_Test_Hyd/edit/triggers

Since linux-daily-glymur.yml & linux-daily-monza.yml are having same check runs , which is effecting Glymur nightly coral workflow to be initiated even for monza nightly completion , 

To avoid conflict from other workflows, making glymur workflow checkrun name as build-glymur

### DIFF
--- a/.github/workflows/linux-daily-glymur.yml
+++ b/.github/workflows/linux-daily-glymur.yml
@@ -14,7 +14,7 @@ permissions:
   pull-requests: write # linux.yml
 
 jobs:
-  build:
+  build-glymur:
     # don't run cron from forks of the main repository or from other branches
     if: github.repository == 'qualcomm-linux/qcom-deb-images' && github.ref == 'refs/heads/main'
     uses: ./.github/workflows/linux.yml


basak-qcom:
This seems like a trivial change that's easy to make, but there are a couple of things that seem error prone here that might be worth addressing.

1. Can Coral instead make sure to key on the workflow as well (preferably filename or at least the name) so that it is more robust, rather than fixing this at the other end where the issue may repeat itself?
2. If we're effectively imposing a rule on the workflow on this side that some job names must be unique, please could you leave comments to that effect against every such job to ensure that we do not inadvertenly regress in the future?

shja-qcom:
Coral workflows triggered by GitHub Check Run events rely solely on the check run name for matching with configured triggers. They do not consider or validate the creator of the check run. To make this mechanism more robust and avoid conflicts, can we ensure uniqueness by incorporating additional context—such as the workflow file path—into the check run name itself or following some nomenclature.


For example:


- If it's a pre-merge workflow  
  - glymur: build-> build-pr-glymur
  - monza: build-> build-pr-monza


- If it's a nightly build workflow
  - glymur: build-> build-nightly-glymur
  - monza: build-> build-nightly-monza


                                                  
If there are any more variables that come in future, they can also be added further to distinguish

@basak-qcom @lool

basak-qcom:
> Coral workflows triggered by GitHub Check Run events rely solely on the check run name for matching with configured triggers.

To make this mechanism more robust, I'm asking if you can change Coral workflows so that they _do_ rely on more than solely the check run name for matching with configured triggers. Fixing Coral properly would be more robust than a fragile workaround on this end, no?

phanendra-qc:
configured Coral with unique checkrun (build / test-linux-deb / Submit glymur-crd boot) for glymur  , with that Axiom invocation is happening through coral.

lool:
> configured Coral with unique checkrun (build / test-linux-deb / Submit glymur-crd boot) for glymur , with that Axiom invocation is happening through coral.

@basak-qcom Phany closed this inadvertedly, but then confirmed to me on Teams that he addressed this in a different way (see his comment above); is that alright for your needs?

PR #400, opened 2026-05-07
fix!(rootfs): use dracut for newer Debian suites

gagath:
Debian has migrated from initramfs-tools to dracut for forky. Update the kernel depedencies step to reflect this change as well.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -395,7 +395,11 @@ actions:
     recommends: true
     packages:
       - apparmor
+{{- if eq $suite "trixie" }}
       - initramfs-tools
+{{- else }}
+      - dracut
+{{- end }}
       - kmod
       - linux-base
 


gagath:
The CI run is not the best to test this change. I did a local test with a custom kernel package and the forky suite, boot-tested the image in QEMU and it boots fine. The `/usr/sbin/update-initramfs` file is from `dracut` so that works, and the `initramfs-tools` package is not installed anymore in the image.

lool (approved):
@gagath Looks good, FYI I've been following https://www.conventionalcommits.org/en/v1.0.0/ as guidance for Conventional Commits (perhaps you have a different reference?), which suggests having the exclamation mark just before ":"

(I initially thought this would be a "feat", but since Debian is changing I agree "fix" is proper)

gagath:
> Looks good, FYI I've been following https://www.conventionalcommits.org/en/v1.0.0/ as guidance for Conventional Commits (perhaps you have a different reference?), which suggests having the exclamation mark just before ":"

Thanks for the reminder, I updated the commit message.

lool:
I had already approved, merging this now :)

gagath:
> I had already approved, merging this now :)

Yes, thanks. I was waiting for the CI run before merging it but was doing something else. :)

Issue #263, opened 2026-02-25
Use snapshots for regular daily builds


basak-qcom:
This ticket was originally about snapshot support, full desc below. The remaining part is basing the daily builds on the snapshot support that did land now (part 4 below).

---------------

What I'd like is:

1. When building the rootfs, allow as a build configuration parameter the specification of a specific apt snapshot for all apt sources used, where the source apt repositories support snapshots.
2. After the build, drop the snapshot use from sources.list, so that when the user runs `apt update/upgrade`, they get the latest again.
3. Log the snapshot used as part of the build logs or output artifacts.
4. By default, the build should pick the current date and specify that as the snapshot.

Rationale: this will permit builds to become reproducible in terms of specific package versions used.

lool:
We landed support for this some time back

basak-qcom:
@lool I don't think this is complete. Step 4 isn't done, which I think is important because it means that someone not paying attention to snapshot support will nonetheless be able to reproduce precisely a build they made previously. For example, our daily CI is currently not doing this. I want to be able to precisely repeat the behaviour of a particular daily CI job using the snapshot support.

PR #403, opened 2026-05-08
support to reproduce the build

vishwasudupa:
Record every installed package, its version, and APT source repo in
installed-packages.txt at the end of each debos rootfs build.  The
file is copied to ARTIFACTDIR and staged alongside other artifacts
in the GitHub Actions workflow.

ex: installed-packages.txt
openssl-provider-legacy 3.5.5-1~deb13u2 http://deb.debian.org/debian-security trixie-security/main arm64
pinentry-curses 1.3.1-2 http://deb.debian.org/debian trixie/main arm64
pipewire        1.4.2-1~qcom1   https://qartifactory-edge.qualcomm.com/artifactory/qsc-deb-releases trixie-overlay/main arm64

Add debos-recipes/qualcomm-linux-debian-reproduce.yaml and a
Makefile `reproduce` target that consume installed-packages.txt to
rebuild an identical rootfs:

- Parse the file outside the chroot (chroot: false) to reconstruct
  the exact APT sources and generate /root/install-packages.sh with
  every package pinned to its recorded version.
- Run the install script inside the chroot (chroot: true) with
  DEBIAN_FRONTEND=noninteractive to avoid interactive prompts.
- Accept locally-built .deb files (e.g. a custom kernel) via the
  LOCALDEBS variable, mirroring the rootfs recipe's localdebs path.

Usage:
  make rootfs   // would create installed-packages.txt and when run in ci it will also be uploaded

// to reproduce one can use qualcomm-linux-debian-reproduce.yaml , it recreates same outputs as qualcomm-linux-debian-rootfs.yaml (except installed-packages.txt since its reproduce)

  make reproduce
  make reproduce PACKAGES_FILE=installed-packages.txt
  make reproduce LOCALDEBS=debos-recipes/local-debs/

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-reproduce.yaml
+++ b/debos-recipes/qualcomm-linux-debian-reproduce.yaml
@@ -1,0 +1,126 @@
+{{- $packagesfile := or .packagesfile "installed-packages.txt" }}
+{{- $suite := or .suite "trixie" }}
+{{- $localdebs := or .localdebs "none" }}
+
+architecture: arm64
+
+actions:
+  - action: mmdebstrap
+    description: Bootstrap minimal base filesystem
+    suite: {{ $suite }}
+    components:
+      - main
+      - contrib
+      - non-free
+      - non-free-firmware
+    mirrors: ['http://deb.debian.org/debian']
+    variant: minbase
+    # ca-certificates needed for https APT sources; python3 needed for the
+    # pinned-install step below
+    include: [ca-certificates, python3]
+
+  - action: overlay
+    description: Install QSC APT keyring (needed for qsc-deb-releases repo)
+    source: overlays/qsc-deb-releases
+
+  - action: run
+    description: Parse packages list – write APT sources and install script into rootfs
+    chroot: false
+    command: |
+      set -eux
+      python3 -c "
+      import os
+      rootdir     = os.environ['ROOTDIR']
+      artifactdir = os.environ['ARTIFACTDIR']
+      pkgsfile    = os.path.join(artifactdir, '{{$packagesfile}}')
+      lines = [l.strip() for l in open(pkgsfile) if l.strip() and not l.startswith('#')]
+      pkgs, srcs, local_pkgs = {}, {}, []
+      for l in lines:
+          p = l.split('\t')
+          if len(p) < 3: continue
+          pkg, ver, src = p[0], p[1], p[2]
+          if src == 'local/unknown':
+              local_pkgs.append((pkg, ver)); continue
+          pkgs[pkg] = ver
+          parts = src.split()
+          if len(parts) >= 2:
+              url, f2 = parts[0], parts[1]
+              f3 = parts[2] if len(parts) > 2 else ''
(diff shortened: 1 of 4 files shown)


debos-recipes/qualcomm-linux-debian-reproduce.yaml
```diff
@@ -0,0 +1,126 @@
...
+              return '/etc/apt/keyrings/qsc-deb-releases.asc'
+          return ''
+      with open(os.path.join(srcdir, 'reproduced-build.sources'), 'w') as f:
+          for url, suite, comp in sorted(srcs):
+              sb = signed_by(url)
+              entry = 'Types: deb\nURIs: {}\nSuites: {}\nComponents: {}\n'.format(url, suite, comp)
+              if sb: entry += 'Signed-By: {}\n'.format(sb)
+              f.write(entry + '\n')
+      os.makedirs(os.path.join(rootdir, 'root'), exist_ok=True)
+      with open(os.path.join(rootdir, 'root/install-packages.sh'), 'w') as f:
```
basak-qcom, line 72: I'm not keen on generating a script and then running it. It makes development and debugging this obscure. Why is this necessary?

basak-qcom (changes requested):
Hi! This is an important feature that we definitely need. It's planned in #263 using a completely different method.

I'm not sure this method will work, as in the future we probably will not have all deb versiosn ever published available from a non-snapshot apt URL. That will bloat too much.

See #263 for the planned approach here.

lool:
Could you also explain why the current SBOMs generated by syft are not enough to capture installed-packages?

vishwasudupa:
> Could you also explain why the current SBOMs generated by syft are not enough to capture installed-packages?

 i looked at trixie-rootfs-sbom.syft.txt.gz, this has version info: so it is good
 does sbom also ensures from which apt repo ?
 
 will there be issues if same version deb is present in 2 different repo. with difference in code in background (ideally this should not happen but if copies are different) 
 
 so, i even traced apt repo from where it will be picked from in the file ( would apt url not be needed for reproducibility)
 
 ex: 
 evolution-data-server-common	3.56.2-0+deb13u1	http://deb.debian.org/debian trixie/main arm64
exfatprogs	1.2.9-1+deb13u1	http://deb.debian.org/debian trixie/main arm64
exo-utils	4.20.0-1	http://deb.debian.org/debian trixie/main arm64
fastrpc-support	1.0.2-1~bpo~qcom1	https://qartifactory-edge.qualcomm.com/artifactory/qsc-deb-releases trixie-overlay/main arm64
fastrpc-tests	1.0.2-1~bpo~qcom1	https://qartifactory-edge.qualcomm.com/artifactory/qsc-deb-releases trixie-overlay/main arm64

debos-recipes/qualcomm-linux-debian-reproduce.yaml
vishwasudupa: it is not necessary, will simplify in my next revision

lool:
(For the record, I think Robie, Agathe and I all prefer the new approach you've taken in #422, so closing this one :)

PR #404, opened 2026-05-08
fix!: remove support for "sid" suite

gagath:
The "sid" suite is an alias for "unstable" but it tends to not be be used less and less. For example, reproduce.debian.net has the following:

	https://reproduce.debian.net/all/stats/unstable/ -> HTTP 200

But sid is not a supported alias:

	https://reproduce.debian.net/all/stats/sid/ -> HTTP 404

Drop the support for the "sid" alias to be in line with what other people are doing in Debian.


lool:
Hmm ok, I don't care too strongly right now, but I kind of liked being consistent in using actual suite names (trixie, forky, sid) rather than their risk aliases...

gagath:
Superseded by #410.

PR #405, opened 2026-05-08
ci: migrate to reusable test workflow

mwasilew:
Replace current testing setup with reusable test-distro.yml workflow from meta-qcom.

See also https://github.com/qualcomm-linux/lava-test-plans/pull/30


mwasilew:
@lool I'll be picking this up soon

.github/workflows/build-debian.yml
```diff
@@ -67,14 +67,17 @@ jobs:
       LAVATOKEN: ${{ secrets.LAVATOKEN }}
```
obbardc, line 67: change commit title to `ci: migrate to reusable test workflow` to satisfy linter

.github/workflows/build-debian.yml
```diff
@@ -67,14 +67,17 @@ jobs:
...
     with:
```
obbardc, line 68: remove file `.github/workflows/.lava-test.yml.swp`

.github/workflows/build-debian.yml
```diff
@@ -67,14 +67,17 @@ jobs:
...
       suite: ${{ matrix.suite }}
+      # qcs6490-rb3gen2-vision-kit is the only board tested here:
       # qrb2210-rb1 testing is disabled until it can be resurrected; see #424
       # glymur-crd can't boot the distro kernel, only qcom-next
       # monaco-arduino-monza and monaco-evk only validated against the
       # linux-next/qcom-next/arduino
       # kernels so far, not the distro kernel; see
       # https://github.com/qualcomm-linux/qcom-deb-images/issues/440
-      boards_exclude: '["glymur-crd", "lemans-evk", "monaco-arduino-monza", "monaco-evk", "qcs615-ride", "qcs8300-ride", "qrb2210-rb1"]'
+      boards_include: 'qcs6490-rb3gen2-vision-kit'
```
obbardc, line 80: Just wondering why the other boards are not yet tested?

.github/workflows/build-daily.yml
```diff
@@ -75,6 +75,9 @@ jobs:
       LAVATOKEN: ${{ secrets.LAVATOKEN }}
     with:
       url: ${{ needs.build-daily.outputs.artifacts_url }}
+      # id of the run that built the image under test
+      build_id: ${{ github.run_id }}
       suite: ${{ matrix.suite }}
+      # every board but monaco-arduino-monza; see
       # https://github.com/qualcomm-linux/qcom-deb-images/issues/537
-      boards_exclude: '["monaco-arduino-monza"]'
+      boards_include: 'glymur-crd,lemans-evk,monaco-evk,qcs615-ride,qcs6490-rb3gen2-vision-kit,qcs8300-ride,qrb2210-rb1'
```
obbardc: why not keep `boards_exclude` ?

obbardc: oh, i see you removed `boards_exclude`, see later comment about that.

.github/workflows/lava-test.yml
```diff
@@ -6,16 +6,17 @@ on:
...
         required: true
         type: string
+      build_id:
+        required: true
+        type: string
       suite:
         description: Distribution suite (e.g. trixie, forky)
         type: string
         default: trixie
       boards_include:
```
obbardc, line 16: OK, we now don't have `boards_exclude` and have to manually set the list of boards to enable tests on. Not sure how I feel about that :/

.github/workflows/lava-test.yml
```diff
@@ -40,183 +50,48 @@ on:
...
-      jobmatrix: ${{ steps.listjobs.outputs.jobmatrix }}
-      has_jobs: ${{ steps.listjobs.outputs.has_jobs }}
     steps:
-      - name: Clone repository
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
-          fetch-depth: 0
-          persist-credentials: false
-      - name: Print trigger
+      - name: "Print ref"
```
obbardc:
```suggestion
      - name: Print lava-test-plans ref
```
or similar ?

.github/workflows/lava-test.yml
```diff
@@ -40,183 +50,48 @@ on:
...
-          save_job_details: true
-          # Stable per-suite/per-board name: the action deletes the previous
-          # same-named artifact before uploading, so a re-run overwrites in
-          # place and only the latest result per board ever exists.
-          result_file_name: test-results-${{ inputs.suite }}-${{ matrix.slug }}
-          # Job-details artifacts always keep the (changing) LAVA job id and are
-          # never overwritten, so scope them by suite + run attempt instead.
-          test_job_file_name_prefix: ${{ inputs.suite }}-${{ github.run_attempt }}-
+  test-debian:
+    needs: [prepare-env]
```
obbardc: remove `[]` from the needs, not needed

.github/workflows/lava-test.yml
```diff
@@ -40,183 +50,48 @@ on:
...
+    name: "Test nodistro"
```
obbardc: remove `""` from the name ?

obbardc (commented):
Looks good so far, some small thoughts!

.github/workflows/lava-test.yml
mwasilew: Ideally there would be just one list, but since different builds use different kernels I assume there needs to be some list of boards to test for each build. The include/exclude was added when test job templates were kept in the qcom-deb-images repository. Now the templates are moved to lava-test-plans. `exclude` makes sense only if you have a list of "all boards" somewhere. I don't think such list exists at the moment so from my pov it's best to just keep "include" list - list of boards to be tested. It's easier to maintain IMHO as there is nothing "hidden". Tests are only run on the boards listed in the workflow.

.github/workflows/build-debian.yml
mwasilew: I've no idea how this made it's way to the commit.

.github/workflows/build-debian.yml
mwasilew: done

.github/workflows/lava-test.yml
mwasilew: done

lool:
@mwasilew @obbardc where are we with this PR? I see some resolved, and some unresolved comments, and it's still draft, so I'm not clear on who is expected to act next

mwasilew:
@lool I think it should remain in draft for now. I'm trying to get the reusable workflow merged into [qualcomm/reusable-workflows](https://github.com/qualcomm/qcom-reusable-workflows) repository. There are a few hurdles on the way (like pinning the action versions). I'll try to proceed as quickly as possible. This is on top of my todo list.

.github/workflows/lava-test.yml
obbardc: makes sense.

mwasilew:
I pushed a new version that is currently available on `ci-testing` branch. The workflow is running here:
https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/35226840721
I'm moving this PR out of draft.

.github/known-failures/forky.yaml
```diff
@@ -0,0 +1,17 @@
+# Tests that are known to fail on forky images. See README.md for the format.
+#
+# Add the issue each failure is tracked in as its comment, and remove the entry
```
obbardc, line 3: just a small reminder for myself that we should add tracking issues for known-failures before merging this

.github/known-failures/forky.yaml
```diff
@@ -0,0 +1,17 @@
+# Tests that are known to fail on forky images. See README.md for the format.
```
obbardc, line 1: - is `known-failures` really descriptive enough? What about `expected-failures`? I originally thought of `flakes` but tthat;s not really true here.
- I was wondering if it was worth moving these files under `.github/test/known-failures` or something to not pollute the top level.

.github/known-failures/trixie.yaml
```diff
@@ -0,0 +1,17 @@
+# Tests that are known to fail on trixie images. See README.md for the format.
+#
+# Add the issue each failure is tracked in as its comment, and remove the entry
+# once the issue is fixed: a listed test that passes is reported as an
+# unexpected pass and fails the run.
+
+lemans-evk:
```
obbardc, line 7: I think we may have to add more matches into this file later, since we will be adding variant images into the mix. As long as it's possible to easily change the format of this later, all good.

.github/scripts/apply-known-failures.py
```diff
@@ -0,0 +1,369 @@
+#!/usr/bin/env python3
```
obbardc, line 1: I think this script should go into a `test` dir or similar. Nitpick!

.github/known-failures/forky.yaml
mwasilew: https://github.com/qualcomm-linux/qcom-deb-images/issues/645

.github/known-failures/forky.yaml
mwasilew: what would "expected failures" be? It's up to the test script author to determine the outcome. If you're testing for an exception and you get it, does this constitute "expected failure"? If so I would call it a pass. 

Known failures should only contain tests that fail (don't end with expected result whatever it is) when they should not. Something that fails when it's expected to is in my book a `pass`.

.github/known-failures/trixie.yaml
mwasilew: I think we should decide on the format now. This is an input for reporting action and the action will soon go to reusable-workflows. Better to cover all known use cases before that.

.github/known-failures/forky.yaml
obbardc: All good.

.github/known-failures/trixie.yaml
obbardc: No, all is fine. The "variant" changes I am thinking about are not well defined yet. I think they will just have different kernels, so I think for now this is fine. It will be 2-3 months away at least, so let's not block on it.

obbardc (approved):

PR #408, opened 2026-05-14
Enable multidtb fitimage

vishwasudupa:
Add a `dtbs` argument to scripts/build-linux-deb.py to enable packaging
DTB and DTBO artifacts as part of the kernel Debian build.

Add a `fitimage` argument to build fit-image.dtb.bin and include it as
dtb.bin under the flash-* directory, ensuring the FIT image is generated
and packaged during make flash

usage 

 make flash EXTRA_DEBOS_OPTS="-t fitimage:true"

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -73,8 +73,9 @@ jobs:
         run: cp -av "/efs/u-boot-rb1-latest/rb1-boot.img" .
 
       # mtools is needed for the flash recipe
+      # file, device-tree-compiler and u-boot-tools are needed by qcom-dtb-metadata
       - name: Install debos and dependencies of the recipes and local tests
-        run: apt -y install debian-archive-keyring debos make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml
+        run: apt -y install debian-archive-keyring debos file make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml device-tree-compiler u-boot-tools
 
       - name: Setup local APT repo
         run: |
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -18,6 +18,15 @@ actions:
     url: https://github.com/qualcomm-linux/qcom-ptool/archive/6c9922f220c884236303a94075eb7c62e0120af0.tar.gz
     name: qcom-ptool
     filename: qcom-ptool.tar.gz
+    sha256sum: f201dcaee55c6dda30afad7c98c76768dc02d29c567dbc91690b43f54cbeb41f
+    unpack: true
+
+  - action: download
+    description: Download qcom-dtb-metadata (DTB image build scripts and ITS/metadata files)
+    url: https://github.com/qualcomm-linux/qcom-dtb-metadata/archive/bf8f11f5274d850f71cc1af8b5a5c46683c14eee.tar.gz
+    name: qcom-dtb-metadata
+    filename: qcom-dtb-metadata.tar.gz
+    sha256sum: 4e2a4bf2c4939ebb7093f1dd3ce2eea69325e1b6f4514696c7c7475b010316ae
     unpack: true
 
 {{- $boards := list }}
@@ -63,6 +72,7 @@ actions:
     )
     "cdt_filename" "cdt_adp_air_sa6155p.bin"
     "dtb" "qcom/qcs615-ride.dtb"
+    "dtb_bin_type" "combineddtb"
 )}}
 {{- end }}
 {{- if eq $build_qcm6490 "true" }}
@@ -86,6 +96,7 @@ actions:
     )
     "cdt_filename" "cdt_vision_kit.bin"
     "dtb" "qcom/qcs6490-rb3gen2.dtb"
+    "dtb_bin_type" "combineddtb"
 )}}
 {{- end }}
 {{- if eq $build_qcs8300 "true" }}
@@ -109,6 +120,7 @@ actions:
     )
     "cdt_filename" "cdt_ride_sx.bin"
     "dtb" "qcom/qcs8300-ride.dtb"
+    "dtb_bin_type" "combineddtb"
 )}}
 {{- $boards = append $boards (dict
     "name" "monaco-evk"
@@ -174,6 +186,7 @@ actions:
     )
     "cdt_filename" "cdt_ride_sx.bin"
     "dtb" "qcom/qcs9100-ride-r3.dtb"
+    "dtb_bin_type" "combineddtb"
 )}}
 {{- $boards = append $boards (dict
     "name" "lemans-evk"
@@ -211,6 +224,7 @@ actions:
     )
     "u_boot_file" .u_boot_rb1
     "dtb" "qcom/qrb2210-rb1.dtb"
+    "dtb_bin_type" "combineddtb"
 )}}
 {{- end }}
 {{- $boards = append $boards (dict
@@ -225,6 +239,7 @@ actions:
         "sha256sum" "c606e95d0107f8c58d0dd9494e00624d1db7c4361cca20513bc78ef02ca28dd1"
     )
     "dtb" "qcom/qrb2210-arduino-imola.dtb"
+    "dtb_bin_type" "combineddtb"
 )}}
 
 {{- range $board := $boards }}
@@ -297,6 +312,19 @@ actions:
           ;;
       esac
 
+      # Generate dtb.bin (multi-DTB FIT image) from the qcom/ subtree of dtbs.tar.gz.
+      if [ -f "${ARTIFACTDIR}/dtbs.tar.gz" ]; then
+          QCOM_DTB_METADATA="$(ls -d "${ROOTDIR}/../qcom-dtb-metadata.tar.gz.d/qcom-dtb-metadata-"*)"
+          mkdir -p build/dtbs
+          # unpack only qcom/ DTBs
+          tar -C build/dtbs -xzf "${ARTIFACTDIR}/dtbs.tar.gz" qcom
+          "${QCOM_DTB_METADATA}/build-dtb-image.sh" \
+              --dtb-src build/dtbs/qcom \
+              --out "${ARTIFACTDIR}/dtb-multidtb.bin" \
+              --prune
+          echo "Multi-DTB FIT image generated: ${ARTIFACTDIR}/dtb-multidtb.bin"
+      fi
+
 {{- range $board := $boards }}
       ### board: {{ $board.name }}
       ### silicon family: {{ $board.silicon_family }}
@@ -305,6 +333,9 @@ actions:
       # in shell syntax
       board_name="{{$board.name}}"
       board_dtb="{{$board.dtb}}"
+      # default dtb_bin_type is multidtb
+      board_dtb_bin_type="{{or $board.dtb_bin_type "multidtb"}}"
+      echo "Board ${board_name}: dtb_bin_type=${board_dtb_bin_type}"
       boot_binaries_filename="{{$board.boot_binaries_download.filename}}"
       board_cdt_download_filename=""
       board_cdt_filename=""
@@ -365,6 +396,7 @@ actions:
                   "$board_cdt_filename" \
                   "$buildid" \
                   "$disk_type" \
+                  "$board_dtb_bin_type" \
           )
 
           # create board-specific flash directory
@@ -418,23 +450,25 @@ actions:
                   "${flash_dir}"
           fi
 
-          # generate a dtb.bin FAT partition with just a single dtb for the
-          # current board; long-term this should really be a set of dtbs and
-          # overlays as to share dtb.bin across boards
-          dtb_bin="${flash_dir}/dtb.bin"
-          rm -f "${dtb_bin}"
-          # dtb.bin is only used in UFS based boards at the moment and UFS
-          # uses a 4k sector size, so pass -S 4096 in
-          # qcom-ptool/platforms/*/partitions.conf, dtb_a and _b partitions
-          # are provisioned with 64MiB; create a 4MiB FAT that will
-          # comfortably fit in these and hold the target device tree, which is
-          # 4096 KiB sized blocks for mkfs.vfat's last argument
-          mkfs.vfat -S 4096 -C "${dtb_bin}" 4096
-          # extract board device tree from the root filesystem provided
-          # tarball
-          tar -C build -xvf "${ARTIFACTDIR}/dtbs.tar.gz" "$board_dtb"
-          # copy into the FAT as combined-dtb.dtb
-          mcopy -vmp -i "${dtb_bin}" "build/${board_dtb}" ::/combined-dtb.dtb
+          if [ "${board_dtb_bin_type}" = "combineddtb" ]; then
+              # generate a dtb-combineddtb.bin" FAT partition with just a single dtb for the
+              # current board; long-term this should really be a set of dtbs and
+              # overlays as to share dtb.bin across boards
+              dtb_bin="${flash_dir}/dtb-combineddtb.bin"
+              rm -f "${dtb_bin}"
+              # dtb-combineddtb.bin is only used in UFS based boards at the moment
+              # and UFS uses a 4k sector size, so pass -S 4096 in
+              # qcom-ptool/platforms/*/partitions.conf, dtb_a and _b partitions
+              # are provisioned with 64MiB; create a 4MiB FAT that will
+              # comfortably fit in these and hold the target device tree, which is
+              # 4096 KiB sized blocks for mkfs.vfat's last argument
+              mkfs.vfat -S 4096 -C "${dtb_bin}" 4096
+              # extract board device tree from the root filesystem provided
+              # tarball
+              tar -C build -xvf "${ARTIFACTDIR}/dtbs.tar.gz" "$board_dtb"
+              # copy into the FAT as combined-dtb.dtb
+              mcopy -vmp -i "${dtb_bin}" "build/${board_dtb}" ::/combined-dtb.dtb
+          fi
 {{- end }}
       fi
 {{- end }}
@@ -464,6 +498,7 @@ actions:
                           -u "//*[@storage_type='${storage}']/file_name[text()='efi.bin']" -v "${efi_img}" \
                           -u "//*[@storage_type='${storage}'][file_name/text()='rootfs.img']/file_path[text()='../']" -v "../../" \
                           -u "//*[@storage_type='${storage}']/file_name[text()='rootfs.img']" -v "${rootfs_img}" \
+                          -u "//download_file[file_name/text()='dtb-multidtb.bin']/file_path[text()='..']" -v "../../" \
                           "${copied_contents}"
                   fi
               done
--- a/scripts/bundle-flash-dirs.sh
+++ b/scripts/bundle-flash-dirs.sh
@@ -47,8 +47,8 @@ echo "ufs_dirs: $ufs_dirs"
 
 # word splitting is a feature in this case
 # shellcheck disable=SC2086
-tar -cvzf "$output_dir/$prefix-flash-emmc.tar.gz" disk-sdcard.img1 disk-sdcard.img2 $emmc_dirs
+tar -cvzf "$output_dir/$prefix-flash-emmc.tar.gz" disk-sdcard.img1 disk-sdcard.img2 dtb-multidtb.bin $emmc_dirs
 
 # word splitting is a feature in this case
 # shellcheck disable=SC2086
-tar -cvzf "$output_dir/$prefix-flash-ufs.tar.gz" disk-ufs.img1 disk-ufs.img2 $ufs_dirs
+tar -cvzf "$output_dir/$prefix-flash-ufs.tar.gz" disk-ufs.img1 disk-ufs.img2 dtb-multidtb.bin $ufs_dirs
--- a/scripts/gen-ptool.sh
+++ b/scripts/gen-ptool.sh
@@ -19,6 +19,9 @@ CDT_FILENAME="$3"
 BUILDID="$4"
 # disk storage, emmc, nvme, spinor or ufs
 DISK_TYPE="$5"
+# dtb type: multidtb (shared ../dtb-multidtb.bin at ARTIFACTDIR level) or
+# combineddtb (local dtb-combineddtb.bin in flash dir); default is combineddtb
+DTB_TYPE="${6:-combineddtb}"
 
 PARTITIONS_CONF="${QCOM_PTOOL}/platforms/${PLATFORM}/partitions.conf"
 
@@ -42,6 +45,19 @@ case "$DISK_TYPE" in
     ;;
 esac
 
+case "$DTB_TYPE" in
+  multidtb)
+    dtb="../dtb-multidtb.bin"
+    ;;
+  combineddtb)
+    dtb="dtb-combineddtb.bin"
+    ;;
+  *)
+    echo "unsupported dtb type $DTB_TYPE"
+    exit 1
+    ;;
+esac
+
 # build a map of partition names from partitions.conf to our names
 #
 # |--------|--------------|-------------------|-----------------|
@@ -53,8 +69,8 @@ esac
 # | CDTs   | cdt          | unset / per board | from download   |
 # |--------|--------------|-------------------|-----------------|
 partition_map="cdt=$(basename "${CDT_FILENAME}")"
-partition_map="${partition_map},dtb_a=dtb.bin"
-partition_map="${partition_map},dtb_b=dtb.bin"
+partition_map="${partition_map},dtb_a=${dtb}"
+partition_map="${partition_map},dtb_b=${dtb}"
 partition_map="${partition_map},efi=${esp}"
 partition_map="${partition_map},rootfs=${rootfs}"
 
@@ -74,4 +90,3 @@ fi
 
 # generate flashing files from qcom-partitions.xml
 "${QCOM_PTOOL}/ptool.py" -x ptool-partitions.xml
-
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 disk-sdcard.img*
 disk-ufs.img*
 dtbs.tar.gz
+dtb-multidtb.bin
+dtb-combineddtb.bin
 *.deb
 *.buildinfo
 *.changes
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,8 @@ clean:
 	rm -f $(DISK_SDCARD_IMAGES)
 	rm -f rootfs.tar
 	rm -f dtbs.tar.gz
+	rm -f dtb-multidtb.bin
+	rm -f dtb-combineddtb.bin
 
 .PHONY: clean-debos
 clean-debos:


debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -20,6 +21,15 @@ actions:
...
     unpack: true
 
+{{- if eq $fitimage "true" }}
+  - action: download
+    description: Download qcom-dtb-metadata (DTB image build scripts and ITS/metadata files)
+    url: https://github.com/qualcomm-linux/qcom-dtb-metadata/archive/e08e4e175779587c6ed91f3f4fd620ba722f5ae2.tar.gz
+    name: qcom-dtb-metadata
+    filename: qcom-dtb-metadata.tar.gz
+    unpack: true
+{{- end }}
```
gagath: Please add a sha256sum or similar to prevent unintended changes.

.github/workflows/debos.yml
```diff
@@ -74,7 +74,7 @@ jobs:
 
       # mtools is needed for the flash recipe
       - name: Install debos and dependencies of the recipes and local tests
-        run: apt -y install debian-archive-keyring debos make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml
+        run: apt -y install debian-archive-keyring debos make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml device-tree-compiler
```
gagath: Optimization: only install device-tree-compiler if $fitimage is true (probably in a separate step).

scripts/build-linux-deb.py
```diff
@@ -311,7 +311,7 @@ def main():
         )
 
     log_i("Building Linux deb")
-    build_command = make_base_command + [DEB_PKG_SET]
+    build_command = make_base_command + ["dtbs"] + [DEB_PKG_SET]
```
gagath: Does this change imply that the dtbs.tar.gz was not generated without this option? I do not remember seeing this issue, but maybe the recipe used a previously generated dtbs.tar.gz.

gagath (commented):
Do we expect individual users to choose between a FAT and FIT dtb manually, or is it board-specific? Does our boards support both?

lool:
> Do we expect individual users to choose between a FAT and FIT dtb manually, or is it board-specific? Does our boards support both?

When the boot firmware allows it, we want to switch our dtb.bin to be a FIT image instead of the current legacy payload. I believe that should be possible on all current targets except Agatti (RB1, UNO Q; supported in upcoming boot firmware), Glymur CRD (no dtb.bin supported; supported in upcoming boot firmware)

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -418,6 +441,12 @@ actions:
                   "${flash_dir}"
           fi
 
+{{- if eq $fitimage "true" }}
+          # copy the shared FIT/DTB image as dtb.bin into each board's flash
+          # directory; the bootloader expects the partition image named dtb.bin
+          cp --preserve=mode,timestamps \
+              "${ARTIFACTDIR}/fit_dtb.bin" "${flash_dir}/dtb.bin"
```
lool: in the spirit of common disk-ufs and disk-sdcard, could you move this one level up?

With this approach, all boards have to use FIT image or all boards have to use a legacy combined-dtb. Perhaps wiser to have a per board flag in the boards db? If it's unconditionally supported, then let's not even guard this with if fitimage, and drop support for the old format.

lool (changes requested):
Nice PR! Please make the shared artifact common like the other ones, and please comment on per board or global change; thanks!

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: Hmm I didn't think of that because we embed a checksum in the URL, but you're right; I think we need this for ptool as well

bjordiscollaku:
One suggestion: rather than passing the kernel tree DTB source path via `--dtb-src`, I'd recommend switching to `--kernel-deb` and providing the `.deb` package produced by `build-linux-deb.py` directly.  For reference, take a look at a recent PR I raised for this in distro-images: https://github.com/qualcomm-linux/qcom-distro-images/pull/75/changes.

`build-dtb-image.sh` already handles `.deb` extraction internally (see https://github.com/qualcomm-linux/qcom-dtb-metadata/pull/83), it unpacks the package, locates the DTB directory, and proceeds from there, so there's no need to expose the intermediate source path. This keeps the step self-contained and avoids any dependency on the kernel tree layout.

More importantly, this aligns with the forward-looking direction: the expectation for deb-images is to eventually consume the kernel `.deb` artifact produced by `pkg-linux-qcom` rather than building inline, so it'll serve as a placeholder for that as well.

.github/workflows/debos.yml
vishwasudupa: I have commented below on how to be able to support both FAT and FIT image, if suggested approach if finalized, then probably $fitimage might be redundant

will update based on approach finalized

scripts/build-linux-deb.py
vishwasudupa: dtbo's were not present without this option enabled

vishwasudupa:
> > Do we expect individual users to choose between a FAT and FIT dtb manually, or is it board-specific? Does our boards support both?
> 
> When the boot firmware allows it, we want to switch our dtb.bin to be a FIT image instead of the current legacy payload. I believe that should be possible on all current targets except Agatti (RB1, UNO Q; supported in upcoming boot firmware), Glymur CRD (no dtb.bin supported; supported in upcoming boot firmware)

@gagath yes multi support based on board is needed
thanks @lool for pointing out , yes there are 4 combinations based on firmware support:

1) board supporting both FAT and FIT (kodiak, lemans, monaco)
2) board that support only combined.dtb ( qcs8300-ride, qcs9100-ride,qcs615-ride.dtb)
3) legacy boards (RB1, UNO Q )
4) glymur ( multidtb-support is yet to come)

to support all combinations i am thinking of this solution 

to begin with we can create 2 dtb bins (combined-dtb.bin and multidtb-dtb.bin) at top level

we can use existing ["dtb"](https://github.com/qualcomm-linux/qcom-deb-images/blob/2db4a5695e50ea03e4d31a5b6b0dcb9b04876e62/debos-recipes/qualcomm-linux-debian-flash.yaml#L88) variable in flash yaml to even handle an entry called multidtb

ie: 

if "dtb" "multidtb" -> pack  multidtb-dtb.bin  renamed as dtb.bin to flash-<target>-emmc/ufs
else if "dtb" "qcom/qcs8300-ride.dtb" -> pack combined-dtb.bin renamed as dtb.bin to flash-<target>-emmc/ufs


similar approach is happening in meta-qcom where [QCOM_DTB_DEFAULT](https://github.com/qualcomm-linux/meta-qcom/blob/64f5032b790e365a58bef45096efe889a6e840bb/conf/machine/include/qcom-common.inc#L47) variable defaulted to multidtb and for machines not using multidtb it is changed in machine.conf (eg [-ride](https://github.com/qualcomm-linux/meta-qcom/blob/64f5032b790e365a58bef45096efe889a6e840bb/conf/machine/qcs9100-ride-sx.conf#L9) boards)

@lool , @gagath what do you think about this approach?

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: yes will move to one level up

I have proposed a solution below to support both FAT and FIT based on board. if that is agreed i will drop the gaurd with if fitimage

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: I have updated with sha256sum
also locally verified when sha was altered , the build threw error

I have also updated for ptool in latest patch

vishwasudupa:
pr[ 413](https://github.com/qualcomm-linux/qcom-deb-images/pull/413) would also be needed for multidtb feature 

it gets latest uefi that supports multidtb

lool:
> One suggestion: rather than passing the kernel tree DTB source path via `--dtb-src`, I'd recommend switching to `--kernel-deb` and providing the `.deb` package produced by `build-linux-deb.py` directly. For reference, take a look at a recent PR I raised for this in distro-images: https://github.com/qualcomm-linux/qcom-distro-images/pull/75/changes.
> 
> `build-dtb-image.sh` already handles `.deb` extraction internally (see [qualcomm-linux/qcom-dtb-metadata#83](https://github.com/qualcomm-linux/qcom-dtb-metadata/pull/83)), it unpacks the package, locates the DTB directory, and proceeds from there, so there's no need to expose the intermediate source path. This keeps the step self-contained and avoids any dependency on the kernel tree layout.
> 
> More importantly, this aligns with the forward-looking direction: the expectation for deb-images is to eventually consume the kernel `.deb` artifact produced by `pkg-linux-qcom` rather than building inline, so it'll serve as a placeholder for that as well.

I prefer the original approach in the PR, the kernel might have been installed by apt in the rootfs, and we might not have kept the kernel deb.

The approach is:
- kernels get built in different ways (Debian, pkg-linux-qcom overlay deb, bindeb-pkg, local kernel deb etc.)
- kernels gets pulled in rootfs as part of image build
- we save the dtb collection from the highest version kernel as dtbs.tgz
- this is used as the flashable image generation step

lool:
> > > Do we expect individual users to choose between a FAT and FIT dtb manually, or is it board-specific? Does our boards support both?
> > 
> > 
> > When the boot firmware allows it, we want to switch our dtb.bin to be a FIT image instead of the current legacy payload. I believe that should be possible on all current targets except Agatti (RB1, UNO Q; supported in upcoming boot firmware), Glymur CRD (no dtb.bin supported; supported in upcoming boot firmware)
> 
> @gagath yes multi support based on board is needed thanks @lool for pointing out , yes there are 4 combinations based on firmware support:
> 
> 1. board supporting both FAT and FIT (kodiak, lemans, monaco)
> 2. board that support only combined.dtb ( qcs8300-ride, qcs9100-ride,qcs615-ride.dtb)
> 3. legacy boards (RB1, UNO Q )
> 4. glymur ( multidtb-support is yet to come)
> 
> to support all combinations i am thinking of this solution
> 
> to begin with we can create 2 dtb bins (combined-dtb.bin and multidtb-dtb.bin) at top level
> 
> we can use existing ["dtb"](https://github.com/qualcomm-linux/qcom-deb-images/blob/2db4a5695e50ea03e4d31a5b6b0dcb9b04876e62/debos-recipes/qualcomm-linux-debian-flash.yaml#L88) variable in flash yaml to even handle an entry called multidtb
> 
> ie:
> 
> if "dtb" "multidtb" -> pack multidtb-dtb.bin renamed as dtb.bin to flash--emmc/ufs else if "dtb" "qcom/qcs8300-ride.dtb" -> pack combined-dtb.bin renamed as dtb.bin to flash--emmc/ufs
> 
> similar approach is happening in meta-qcom where [QCOM_DTB_DEFAULT](https://github.com/qualcomm-linux/meta-qcom/blob/64f5032b790e365a58bef45096efe889a6e840bb/conf/machine/include/qcom-common.inc#L47) variable defaulted to multidtb and for machines not using multidtb it is changed in machine.conf (eg [-ride](https://github.com/qualcomm-linux/meta-qcom/blob/64f5032b790e365a58bef45096efe889a6e840bb/conf/machine/qcs9100-ride-sx.conf#L9) boards)
> 
> @lool , @gagath what do you think about this approach?

we discussed this in an unrelated call today, and I've suggested introducing a dtb_bin_type variable in the boards db, and defaulting to FIT image where we expect this to be supported.

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -297,6 +309,19 @@ actions:
...
       esac
 
+{{- if eq $fitimage "true" }}
+      # Generate fit_dtb.bin from the qcom/ subtree of dtbs.tar.gz.
+      # Only qcom/ is extracted to avoid unpacking unrelated vendor DTBs
+      # and to reduce the find(1) search time inside build-dtb-image.sh.
+      QCOM_DTB_METADATA="$(ls -d "${ROOTDIR}/../qcom-dtb-metadata.tar.gz.d/qcom-dtb-metadata-"*)"
+      mkdir -p build/dtbs-src
+      tar -C build/dtbs-src -xzf "${ARTIFACTDIR}/dtbs.tar.gz" qcom
+      "${QCOM_DTB_METADATA}/build-dtb-image.sh" \
```
lool: This default to 4MiB in size; meta-qcom does the same thing:
```
classes/linux-qcom-dtbbin.bbclass:DTBBIN_SIZE ?= "4096"
```

meta-qcom defaults to 4KiB sector size:
```
QCOM_VFAT_SECTOR_SIZE ??= "4096"
```

and so does the script:
```
# -S 5: sector size code → 2^(5+7) = 4096 bytes (matches original mkfs.vfat -S 4096)
# No -F: let mformat auto-select FAT type based on image size, matching mkfs.vfat behaviour
mformat -i "${DTB_BIN}" -S 5 ::
```

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: We discussed this with Vishwas and Srinivas today and I want to share notes for other reviewers.

While partitions.conf documents different dtb.bin sizes across platforms, in practice meta-qcom always generates a 4MiB partition on all platforms right now (no platform overrides this, it doesn't read the partitions.conf size).

We could make it possible to generate a dtb.bin per platform in qcom-deb-images, but I would prefer sticking to a common file for two reasons:
1. this projects a common implementation across platforms
2. this saves build time and disk space by not calling this per board and having one copy per board

vishwasudupa:
@lool 
we had discussed on creating 2 binaries combineddtb.bin and multidtb.bin and use it across all the boards

but currently combined.dtb is only packing single dtb per board 
https://github.com/qualcomm-linux/qcom-deb-images/blob/2db4a5695e50ea03e4d31a5b6b0dcb9b04876e62/debos-recipes/qualcomm-linux-debian-flash.yaml#L437

hence dtb.bin is different for every board

so to implement common combineddtb.bin solution , should i bring in support to even concatenate multiple dtb and created combined.dtb ?

lool:
> @lool we had discussed on creating 2 binaries combineddtb.bin and multidtb.bin and use it across all the boards
> 
> but currently combined.dtb is only packing single dtb per board
> 
> https://github.com/qualcomm-linux/qcom-deb-images/blob/2db4a5695e50ea03e4d31a5b6b0dcb9b04876e62/debos-recipes/qualcomm-linux-debian-flash.yaml#L437
> 
> hence dtb.bin is different for every board
> 
> so to implement common combineddtb.bin solution , should i bring in support to even concatenate multiple dtb and created combined.dtb ?

I don't think we want to invest further in combined-dtb, this was the only thing available and I mimicked what meta-qcom was doing (one DTB in a per board combined-dtb dtb.bin); for FIT image, meta-qcom includes all DTBs, so I think we ought to do the same.

.github/workflows/debos.yml
lool: These are the deps for the CI runs, I think we'll always need device-tree-compiler since we generate for all possible boards, and some boards will use the new format. This should be a per board flag though.

vishwasudupa:
> > @lool we had discussed on creating 2 binaries combineddtb.bin and multidtb.bin and use it across all the boards
> > but currently combined.dtb is only packing single dtb per board
> > https://github.com/qualcomm-linux/qcom-deb-images/blob/2db4a5695e50ea03e4d31a5b6b0dcb9b04876e62/debos-recipes/qualcomm-linux-debian-flash.yaml#L437
> > 
> > hence dtb.bin is different for every board
> > so to implement common combineddtb.bin solution , should i bring in support to even concatenate multiple dtb and created combined.dtb ?
> 
> I don't think we want to invest further in combined-dtb, this was the only thing available and I mimicked what meta-qcom was doing (one DTB in a per board combined-dtb dtb.bin); for FIT image, meta-qcom includes all DTBs, so I think we ought to do the same.

in that case, i will need to create separate dtb,bin for each target and kept under respective flash-<target>-dir
based on  dtb_bin_type flag respective multidtb/combined dtb will be copied to flash folder
this means, won't be able keep dtb,bin common across target. would that be fine?

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: this is a good solution since it removes duplication

but if we need to implement this solution. need to improvise also on combined dtb logic to concatenate dtbs.
since more work on combined dtb, also need to verify form uefi/kernel perspective 
else common combined-dtb.bin wont server across targets

lool:
> in that case, i will need to create separate dtb,bin for each target and kept under respective flash--dir based on dtb_bin_type flag respective multidtb/combined dtb will be copied to flash folder this means, won't be able keep dtb,bin common across target. would that be fine?

I don't understand why? The code can:
1) generate a common dtb_fit.bin for all boards that support that
2) loop over boards and either generate a dtb_combined.bin as dtb.bin or refer to the common dtb_fit.bin, based on the value of the dtb_bin_type

vishwasudupa:
> > in that case, i will need to create separate dtb,bin for each target and kept under respective flash--dir based on dtb_bin_type flag respective multidtb/combined dtb will be copied to flash folder this means, won't be able keep dtb,bin common across target. would that be fine?
> 
> I don't understand why? The code can:
> 
> 1. generate a common dtb_fit.bin for all boards that support that
> 2. loop over boards and either generate a dtb_combined.bin as dtb.bin or refer to the common dtb_fit.bin, based on the value of the dtb_bin_type

ok i will implement this but there would be non uniformitiy amongst dtb.bin across boards, some will have within folder some will be reffring to common dtb_fit.bin
currently 7 amongst 10 boards onboarded is combined dtb

@lool 
also build on PR is failing, it is because it is not able to get lemans-evk dtb

when i locally build with qcom-next kernel, builds pass
how should i handle this issue

lool:
> @lool also build on PR is failing, it is because it is not able to get lemans-evk dtb
> 
> when i locally build with qcom-next kernel, builds pass how should i handle this issue

What's happening is that in qcom-dtb-metadata we document all possible target boards, but based on the actual kernel built by Debian or Qualcomm (e.g. mainline, linux-next, qcom-next), different set of platforms might show up.

We need to fix the FIT image building script to allow missing platforms. Let me know if you need help with that. My preference would be to patch the script in qcom-dtb-metadata with a new optional flag for that, but we could also have a separate script that would process the .its and skip sections that can't be built due to missing DT.

I suspect .its is hard to parse properly, perhaps this should be redesigned in qcom-dtb-metadata so that we have a simpler data structure such as YAML or JSON that generates the .its.

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -418,23 +450,29 @@ actions:
...
-          mkfs.vfat -S 4096 -C "${dtb_bin}" 4096
-          # extract board device tree from the root filesystem provided
-          # tarball
-          tar -C build -xvf "${ARTIFACTDIR}/dtbs.tar.gz" "$board_dtb"
-          # copy into the FAT as combined-dtb.dtb
-          mcopy -vmp -i "${dtb_bin}" "build/${board_dtb}" ::/combined-dtb.dtb
+          if [ "${board_dtb_bin_type}" = "multidtb" ]; then
+              # copy the pre-built multi-DTB image as dtb.bin into the flash directory
+              cp --preserve=mode,timestamps \
+                  "${ARTIFACTDIR}/multidtb-dtb.bin" "${flash_dir}/dtb.bin"
```
lool: please no copy of this data

lool (changes requested):
the first line of the commit message is weird; also, please no cp per board for the common data

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: @lool , wanted a suggestion 

there is a possibility of few targets would expect dtb.bin in raw format instead of vfat-wrapped dtb image due to u-boot requirements
it is still under discussion 
https://github.com/qualcomm-linux/meta-qcom/pull/2094

in this case glymur might have raw fit image dtb.bin and others might have vfat based binary
so shall i keep dtb.bin common or continue to copy seperately

vishwasudupa: discussion tells either, all boards will be raw or all would be FAT compatible
I have implemented request change to not copy dtb.bin in case of multidtb in latest patch

scripts/gen-ptool.sh
```diff
@@ -42,6 +45,15 @@ case "$DISK_TYPE" in
     ;;
 esac
 
+case "$DTB_TYPE" in
```
lool, line 48: please add error case to the case statement

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -305,6 +332,9 @@ actions:
       # in shell syntax
       board_name="{{$board.name}}"
       board_dtb="{{$board.dtb}}"
+      # defailt dtb_bin_type is multidtb
```
lool: default

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -297,6 +312,18 @@ actions:
           ;;
       esac
 
+      # Generate multidtb-dtb.bin (FIT image) from the qcom/ subtree of dtbs.tar.gz.
+      if [ -f "${ARTIFACTDIR}/dtbs.tar.gz" ]; then
+          QCOM_DTB_METADATA="$(ls -d "${ROOTDIR}/../qcom-dtb-metadata.tar.gz.d/qcom-dtb-metadata-"*)"
+          mkdir -p build/dtbs-src
```
lool: just dtbs would be fine IMO

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -297,6 +312,18 @@ actions:
...
+          tar -C build/dtbs-src -xzf "${ARTIFACTDIR}/dtbs.tar.gz" qcom
```
lool: could you add a comment "unpack only qcom/ DTBs"

scripts/build-linux-deb.py
lool: I remain skeptical of this change; I see a bunch of dtbos in the linux-image deb

lool (commented):
Hi @vishwasudupa 
I've left a few easy to address review comments on the first patch; I remain skeptical of the second patch, can you give an example of a dtbo that ends in the deb only after your change? I suspect this should be a kernel patch instead.

scripts/gen-ptool.sh
vishwasudupa: added in latest patch

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: added in latest patch

scripts/build-linux-deb.py
vishwasudupa: 2026/06/05 11:13:45 set -eux... | [WARN] --prune: dropped: lemans-el2.dtbo

if this is not included, all below dtbos are getting pruned with above kind of warning
lemans-el2.dtbo monaco-el2.dtbo talos-el2.dtbo lemans-camx-el2.dtbo hamoa-iot-evk-camera-imx577.dtbo monaco-camx-el2.dtbo x1-el2.dtbo lemans-evk-ifp-mezzanine.dtbo monaco-evk-ifp-mezzanine.dtbo lemans-evk-camera-csi1-imx577.dtbo monaco-evk-camera-imx577.dtbo hamoa-evk-camx.dtbo purwa-evk-camx.dtbo

.github/workflows/debos.yml
```diff
@@ -74,7 +74,7 @@ jobs:
 
       # mtools is needed for the flash recipe
```
lool, line 75: Could you add a comment on a new line that "file, device-tree-compiler and u-boot-tools are needed by qcom-dtb-metadata"?

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -18,6 +18,14 @@ actions:
     url: https://github.com/qualcomm-linux/qcom-ptool/archive/6c9922f220c884236303a94075eb7c62e0120af0.tar.gz
     name: qcom-ptool
     filename: qcom-ptool.tar.gz
+    sha256sum: f201dcaee55c6dda30afad7c98c76768dc02d29c567dbc91690b43f54cbeb41f
```
lool, line 21: It would have been nice to add this in a separate commit, could you mention it in the commit message?

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -18,6 +18,14 @@ actions:
...
+    unpack: true
+  - action: download
```
lool, line 24: We're currently skipping a line between actions, it would be nice to have one here

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -297,6 +312,19 @@ actions:
...
+      if [ -f "${ARTIFACTDIR}/dtbs.tar.gz" ]; then
+          QCOM_DTB_METADATA="$(ls -d "${ROOTDIR}/../qcom-dtb-metadata.tar.gz.d/qcom-dtb-metadata-"*)"
+          mkdir -p build/dtbs
+          # unpack only qcom/ DTBs
+          tar -C build/dtbs -xzf "${ARTIFACTDIR}/dtbs.tar.gz" qcom
+          "${QCOM_DTB_METADATA}/build-dtb-image.sh" \
+              --dtb-src build/dtbs/qcom \
+              --out "${ARTIFACTDIR}/dtb.bin" \
+              --prune
+          echo "Multi-DTB FIT image generated: ${ARTIFACTDIR}/dtb.bin"
```
lool: Maybe we should have named this `dtb-fit.bin` or similar

scripts/bundle-flash-dirs.sh
```diff
@@ -47,8 +47,8 @@ echo "ufs_dirs: $ufs_dirs"
 
 # word splitting is a feature in this case
 # shellcheck disable=SC2086
-tar -cvzf "$output_dir/$prefix-flash-emmc.tar.gz" disk-sdcard.img1 disk-sdcard.img2 $emmc_dirs
+tar -cvzf "$output_dir/$prefix-flash-emmc.tar.gz" disk-sdcard.img1 disk-sdcard.img2 dtb.bin $emmc_dirs
 
 # word splitting is a feature in this case
 # shellcheck disable=SC2086
-tar -cvzf "$output_dir/$prefix-flash-ufs.tar.gz" disk-ufs.img1 disk-ufs.img2 $ufs_dirs
+tar -cvzf "$output_dir/$prefix-flash-ufs.tar.gz" disk-ufs.img1 disk-ufs.img2 dtb.bin $ufs_dirs
```
lool: Note to self: this is the first file conflict between the two flash tarballs. I think it's ok, not sure how to handle that elegantly...

scripts/gen-ptool.sh
```diff
@@ -42,6 +45,19 @@ case "$DISK_TYPE" in
...
```
lool, line 48: This script actually doesn't care about the type of dtb.bin being used, only about the path to use, so we should probably have passed that instead...

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: Thanks; closing this thread!

scripts/build-linux-deb.py
lool: I checked our latest build-linux-deb.py based build of mainline (to see the "proper way") and wanted to see if there are any qcom dtbos in there; there's exactly one:
```
dpkg -c linux-image-7.1.0-rc6_7.1.0~rc6-1_arm64.deb |grep dtbo|grep qcom
-rw-r--r-- root/root      2000 2026-06-01 08:22 ./usr/lib/linux-image-7.1.0-rc6/qcom/talos-evk-camera-imx577.dtbo
```

That seems to be exactly the kind of dtbos we want, and it's shipped in the kernel package; how is this achieved? We can see `arch/arm64/boot/dts/qcom/Makefile` has:
```
dtb-$(CONFIG_ARCH_QCOM) += talos-evk-camera-imx577.dtbo
talos-evk-camera-imx577-dtbs    := talos-evk.dtb talos-evk-camera-imx577.dtbo
dtb-$(CONFIG_ARCH_QCOM) += talos-evk-camera-imx577.dtb
```

So the reason you're not seeing these dtbos being installed is because the upstream kernel Makefile is not installing them... yet we need them installed, especially since we might leverage these dtbos from the HLOS in the future! So you're working around this situation by calling extra targets at kernel installation time.

You were working on a patch to do this upstream, was it not accepted?

My proposal for this:
1) submit a patch to install the required dtbos upstream (or get help to do so)
2) (optional) ask qcom-next maintainers to include this patch
3) submit a separate PR to call the "dtbs" target ahead of "bindeb-pkg" (your current patch), but **with a comment explaining it's a workaround and linking to the upstream patch submission**

lool (changes requested):
Hi @vishwasudupa 

I gave this another top to bottom read and left further comments (I only did a quick review this morning until I found enough feedback to provide); I'm sorry for the many rounds we're doing

For me this is mergeable except for three things:
- I think we should change the call to gen-ptool to take a dtb.bin pathname instead of a dtb.bin type
- please submit the patch to build-linux-deb.py in a separate PR, with a link to the upstream kernel submission
- please split the addition of the ptool sha256sum in its own commit or mention it in your commit message

I've left other minor suggestions here and there, but it otherwise looks good to go.

Could you also confirm on which platform and how you've tested this? :)

Thanks for your nice work!

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: dtb.bin is what is mentioned in rawprogram.xml and content.xml, hence i kept same, 
if needed for someone to differentiate, i will create a extra symlink called dtb-fit.bin or mention as comment
or else, reprogram, content.xml different cases for ufs/emmc and nested folders like (nvme/spinor) all needs to be added.

scripts/gen-ptool.sh
vishwasudupa: doing path based is not straight forward
1) this script runs before flash folder creation 
2) in case of combined dtb the dtb.bin is inside the flash folder and for multidtb (others) it used from outside, hence  cant just put a check like how it is done for other img's 
3) hence only with variable, how to populate xml can be determined, post that flash dir is populated

.github/workflows/debos.yml
vishwasudupa: added in latest patch

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: added info in commit message

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: My thinking was: we used the generic word dtb.bin to refer to that partition, but now it has two possible contents, either a combined DTB, or a FIT image. So we could 1) generate dtb-$implementation.bin for each, and 2) call ptool to s/dtb.bin/<actual path + name>/.

vishwasudupa:
I have an update
on  build_command = make_base_command + ["dtbs"] + [DEB_PKG_SET] in build-linux-deb.py

I have already raised a upstream kernel patch https://lore.kernel.org/all/20260428123725.3457865-1-vudupa@qti.qualcomm.com/

in qcom-next kernel the patch was submitted and merged 
1) https://github.com/qualcomm-linux/kernel/commit/7998dbf1695da00ef7a54c3fe90328e251400c90
2) https://github.com/qualcomm-linux/kernel/commit/abfeee71ec2257a591930d301f965151539d1380

with  "dtbs "change in   build-linux-deb.py **removed,** i tried full build
i see qcom-next deb has all dtbo except purwa-evk-camx.dtbo ( need to raise change to include dtbo) so we are good almost good on qcom-next branch 
I will not raise any new pr on qcom-deb-images to include dtbs

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: I have updated as per this suggestion in latest patch

scripts/gen-ptool.sh
vishwasudupa: updated in latest patch

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -41,6 +50,7 @@ actions:
     )
     "cdt_filename" "CRD/cdt_glymur_crd_0.1.0.bin"
     "dtb" "qcom/glymur-crd.dtb"
+    "dtb_bin_type" "combineddtb"
```
nivasnaik: Please remove this for Glymur as well. Enable fitimage for Glymur.

nivasnaik:
@lool can we prioritize the review/merge of this PR? We need this to enable multiDTB and DTB overlay

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: updated in latest patch

scripts/gen-ptool.sh
lool: I reviewed the logic, and I think what I suggested was possible and perhaps the code would be smaller, but your implementation is consistent with the DISK_TYPE approach, so I'm happy to leave it as is.

scripts/bundle-flash-dirs.sh
lool: we should probably add a comment to make this obvious

scripts/build-linux-deb.py
lool: we've agreed to tackle this separately; I'm leaving that open for reference; see also thread below

lool (approved):
Ok; gave this another look and I think this is ready for merging.

I've spotted a minor error: dtb-combineddtb.bin is never generated at the root, so it makes no sense with this pathname in gitignore or in the make clean target; this should be fixed.

A few notes:
- we've agreed to tackle the kernel dtb bundling separately
- I think passing just the dtb.bin path instead of dtb.bin type to gen-ptool.sh would be possible, but your implementation is consistent with the disk_type, so happy to keep that
- I've closed a number of resolved thread; thanks for the updates!
- for dtb-multidtb.bin being present in both tarballs, we should add a comment, but that's minor
- the new code makes the dtbs.tar.gz a required input now, even if the user selects only boards with combined-dtb, but that mostly requires it too

scripts/build-linux-deb.py
lool: @vishwasudupa I tested this change, and it didn't work for me

PR #410, opened 2026-05-18
fix(rootfs): rename sid to unstable

gagath:
This allows to reduce the if/else complexity and potential for error if one checks only for one or the other but not both.

This change also allows to drop the QArtifactory overlay handling that did have support for "unstable" but not for "sid", which could lead to unexpected results.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -11,6 +11,11 @@
 {{- $variantid = "xfce" }}
 {{- end }}
 
+# Rename sid to unstable to reduce if/else complexity
+{{- if eq $suite "sid" }}
+{{- $suite = "unstable" }}
+{{- end }}
+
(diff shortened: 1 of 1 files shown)


lool (approved):
I like that! great idea

I could see us doing the same thing with testing and forky, but I guess we would want the forky name to be the one that sticks; should we do it the other way around and have "unstable" be mapped to "sid"? That would mean we use specific suite names rather than aliases everywhere.

gagath:
> I could see us doing the same thing with testing and forky, but I guess we would want the forky name to be the one that sticks

I see it staying as-is. When forky releases, we shift from trixie to forky and from forky to what’s next. We could alias testing to forky, but it raise more question for when a release happens that it solves IMO.

> should we do it the other way around and have "unstable" be mapped to "sid"? That would mean we use specific suite names rather than aliases everywhere.

I do not think so. I have seen multiple times in Debian that unstable works where sid does not.

PR #422, opened 2026-05-21
rootfs: snapshot based enablement

vishwasudupa:
enable snapshot based build
build with -t snapshot can take true/date/snapshot-id

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,9 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
   `Debian’s linux-image-arm64`. Can (and should) be set to `none` if you are
   providing local kernel package instead.
 - `suite`: Debian suite to use, defaults to `trixie`.
+- `snapshot`: use a Debian snapshot archive for a reproducible build (`YYYYMMDD`
+  or `YYYYMMDDTHHMMSSZ`); logged to `/etc/buildinfo` as `SNAPSHOT=<date>`.
+  Live mirrors are restored in the final image to allow upgrades.
 
 For the image recipe:
 
--- a/debos-recipes/qualcomm-linux-debian-image.yaml
+++ b/debos-recipes/qualcomm-linux-debian-image.yaml
@@ -2,6 +2,7 @@
 {{- $imagesize := or .imagesize "6GiB" }}
 {{- $imagetype := or .imagetype "ufs" }}
 {{- $image := printf "disk-%s.img" $imagetype }}
+{{- $snapshot := or .snapshot "" }}
 
 architecture: arm64
 sectorsize: {{if eq $imagetype "ufs"}}4096{{else}}512{{end}}
@@ -77,6 +78,16 @@ actions:
     setup-fstab: true
     append-kernel-cmdline: clk_ignore_unused pd_ignore_unused audit=0 deferred_probe_timeout=30
 
+{{- if $snapshot }}
+  - action: run
+    description: Enable snapshot sources for image build
+    chroot: true
+    command: |
+      set -eux
+      apt-snapshot-toggle enable
+      apt-get update
+{{- end }}
+
   - action: apt
     description: Make system bootable with systemd-boot
     recommends: true
@@ -140,6 +151,17 @@ actions:
       EOF
       # enable unit
       systemctl enable debos-grow-rootfs
+
+{{- if $snapshot }}
+  - action: run
+    description: Restore APT sources to live mirrors and clean up snapshot helpers
+    chroot: true
+    command: |
+      set -eux
+      apt-snapshot-toggle disable
+      rm -f /etc/apt/sources.list.d/snapshot_*.sources
+      rm -f /usr/local/bin/apt-snapshot-toggle
+{{- end }}
 
   - action: run
     description: Extract partition images
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -5,6 +5,7 @@
 {{- $kernelpackage := or .kernelpackage "linux-image-arm64" }}
 {{- $overlays := or .overlays "qsc-deb-releases" }}
 {{- $buildid := or .buildid "" }}
+{{- $snapshot := or .snapshot "" }}
 {{- $suite := or .suite "trixie" }}
 
 {{- $variantid := "console" }}
@@ -57,6 +58,30 @@ actions:
       EOF
 {{- end }}
 
+{{- if $snapshot }}
+  - action: run
+    description: Install apt-snapshot-toggle into rootfs
+    chroot: false
+    command: |
+      set -eux
+      cat > "${ROOTDIR}/usr/local/bin/apt-snapshot-toggle" <<'TOGGLE'
+      #!/bin/sh
+      mode="$1"
+      for f in /etc/apt/sources.list.d/*.sources; do
+          case "$f" in
+            */snapshot_*) want=$([ "$mode" = "enable" ] && echo yes || echo no) ;;
+            *)             want=$([ "$mode" = "enable" ] && echo no  || echo yes) ;;
+          esac
+          if ! grep -q "^Enabled:" "$f"; then
+              echo "Warning: no Enabled: field in ${f}" >&2
+          else
+              sed -i "s/^Enabled:.*/Enabled: ${want}/" "$f"
+          fi
+      done
+      TOGGLE
+      chmod +x "${ROOTDIR}/usr/local/bin/apt-snapshot-toggle"
+{{- end }}
+
   # after debootstrap, only a basic etc/apt/sources.list is created; add more
   # modern etc/apt/sources.list.d/*.sources; sources.list is removed after
   # applying all overlays, and followed by an APT update and full-upgrade
@@ -65,6 +90,16 @@ actions:
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
+      if echo "{{ $snapshot }}" | grep -qE '^([0-9]{8}|[0-9]{8}T[0-9]{6}Z)$'; then
+        SNAPSHOT="{{ $snapshot }}"
+      else
+        echo "ERROR: invalid snapshot '{{ $snapshot }}'. Use: YYYYMMDD | YYYYMMDDTHHMMSSZ" >&2
+        exit 1
+      fi
+      touch /etc/buildinfo && chmod 644 /etc/buildinfo
+      echo "SNAPSHOT=${SNAPSHOT}" >> /etc/buildinfo
+{{- end }}
       tee /etc/apt/sources.list.d/debian.sources <<EOF
       # Debian archive
       Types: deb
@@ -72,6 +107,7 @@ actions:
       Suites: {{ $suite }}
       Components: main contrib non-free non-free-firmware
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
+      Enabled: yes
 
 {{- if ne $suite "unstable" }}
 
@@ -81,6 +117,7 @@ actions:
       Suites: {{ $suite }}-updates
       Components: main contrib non-free non-free-firmware
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
+      Enabled: yes
 
       # Debian security updates
       Types: deb
@@ -88,8 +125,21 @@ actions:
       Suites: {{ $suite }}-security
       Components: main contrib non-free non-free-firmware
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
+      Enabled: yes
 {{- end }}
       EOF
+
+{{- if $snapshot }}
+      # Derive snapshot_debian.sources from debian.sources by rewriting mirror URLs
+      sed \
+        -e "s#http://deb\.debian\.org/debian/#https://snapshot.debian.org/archive/debian/${SNAPSHOT}/#g" \
+        -e "s#http://deb\.debian\.org/debian-security/#https://snapshot.debian.org/archive/debian-security/${SNAPSHOT}/#g" \
+        -e "/^Enabled:/a\\# TODO: remove Check-Valid-Until once Debusine-based snapshots are available" \
+        -e "/^Enabled:/a\\Check-Valid-Until: no" \
+        /etc/apt/sources.list.d/debian.sources \
+        > /etc/apt/sources.list.d/snapshot_debian.sources
+      apt-snapshot-toggle enable
+{{- end }}
 
 {{- if and (ne $suite "unstable") (ne $suite "testing") }}
   - action: run
@@ -97,6 +147,9 @@ actions:
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
+      SNAPSHOT=$(grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2)
+{{- end }}
       tee /etc/apt/sources.list.d/debian-backports.sources <<EOF
       # Debian backports
       Types: deb
@@ -106,6 +159,18 @@ actions:
       Enabled: yes
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
       EOF
+
+{{- if $snapshot }}
+      # Derive snapshot_debian-backports.sources from debian-backports.sources by rewriting mirror URLs
+      sed \
+        -e "s#http://deb\.debian\.org/debian/#https://snapshot.debian.org/archive/debian/${SNAPSHOT}/#g" \
+        -e "s#http://deb\.debian\.org/debian\$#https://snapshot.debian.org/archive/debian/${SNAPSHOT}/#" \
+        -e "/^Enabled:/a\\# TODO: remove Check-Valid-Until once Debusine-based snapshots are available" \
+        -e "/^Enabled:/a\\Check-Valid-Until: no" \
+        /etc/apt/sources.list.d/debian-backports.sources \
+        > /etc/apt/sources.list.d/snapshot_debian-backports.sources
+      apt-snapshot-toggle enable
+{{- end }}
       tee /etc/apt/preferences.d/debian-backports.pref <<EOF
       # for binary packages built from these source packages, score the version from
       # Debian backports higher as to get hardware enabled or better hardware support
@@ -164,6 +229,15 @@ actions:
       set -eux
       # remove old style, incomplete sources, superseded in apt-sources overlay
       rm -vf /etc/apt/sources.list
+{{- if $snapshot }}
+      SNAPSHOT=$(grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2)
+      for f in /etc/apt/sources.list.d/*.sources; do
+          case "$f" in */snapshot_*) continue ;; esac
+          if grep -q "^Enabled: yes" "$f"; then
+              echo "Warning: $(basename "$f" .sources) has no snapshot support; packages from this source will be the latest available" >&2
+          fi
+      done
+{{- end }}
       # APT update and upgrade to pickup changes from overlays
       apt-get -y update
       apt-get -y full-upgrade
@@ -562,6 +636,15 @@ actions:
       set -eux
       rm -v /etc/apt/sources.list.d/apt-local-repo.sources
       apt-get update
+{{- end }}
+
+{{- if $snapshot }}
+  - action: run
+    description: Restore APT sources to live mirrors
+    chroot: true
+    command: |
+      set -eux
+      apt-snapshot-toggle disable
 {{- end }}
 
   - action: run


debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -74,30 +90,43 @@ actions:
...
+{{- end }}
+
 {{- if and (ne $suite "unstable") (ne $suite "testing") }}
   - action: run
     description: Add Debian {{ $suite }}-backports APT configuration
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
+      SNAPSHOT_DATE=$(cat /etc/snapshot-date)
```
basak-qcom: For integration with `/etc/buildinfo`, maybe adjust the later step to use `>>` so you could write it here directly without it getting overwritten?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -461,21 +490,49 @@ actions:
...
+      echo "SNAPSHOT=${SNAPSHOT_DATE}" >> /etc/buildinfo
+
+  - action: run
+    description: Write snapshot to build artifact
+    chroot: false
+    command: |
+      set -eux
+      SNAPSHOT_DATE=$(grep '^SNAPSHOT=' "${ROOTDIR}/etc/buildinfo" | cut -d= -f2 || echo "unknown")
+      echo "debos: snapshot build used: ${SNAPSHOT_DATE}"
+      echo "${SNAPSHOT_DATE}" > "${ARTIFACTDIR}/snapshot.txt"
```
basak-qcom: I'm curious to hear @lool 's opinion. My suggestion would be to not ship this in artifacts unless we can articulate why it is needed, if it makes sense to do it, and if the format is correct. I do think that the entry in /etc/buildinfo is appropriate, since we do need to have the snapshot date logged _somewhere_, and this seems like the right place.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -4,6 +4,7 @@
 {{- $kernelpackage := or .kernelpackage "linux-image-arm64" }}
 {{- $overlays := or .overlays "qsc-deb-releases" }}
 {{- $buildid := or .buildid "" }}
+{{- $snapshot := or .snapshot "" }}
```
basak-qcom, line 8: I know you're still working on this, but just a note to get some specification of how this parameter is intended to be used and the accepted forms documented somewhere please.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: updated in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: provided a note in readme

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: I like the etc/buildinfo usage for this; it's also the place that felt the most natural to me

For output artifacts, we currently don't have a clean build metadata file, which I think we could benefit from for CI, but it would be nice to make this broader in scope and capture the requirements first. For instance, perhaps it's relevant for our SBOM.

There were other cases where we wanted to capture CI metadata with our builds (GH workflow build id, build_url, boards.json, snapshot date...), but I would suggest addressing this separately if we can.

vishwasudupa: i have removed snapshot.txt in latest patch
on target buildinfo already holds SNAPSHOT_DATE

this can be addressed at time of CI metadata creation

lool (changes requested):
There are a couple of apt install calls in the image recipe as well; I suspect we want to flip sources back and forth there too :-/

The ideal fix would be to prevent the need for these extra apt installs; the reason they are deferred there is to be local to their integration which depends on the final FS UUID, but perhaps they could be installed from the rootfs recipe and reconfigured later.

vishwasudupa:
> There are a couple of apt install calls in the image recipe as well; I suspect we want to flip sources back and forth there too :-/
> 
> The ideal fix would be to prevent the need for these extra apt installs; the reason they are deferred there is to be local to their integration which depends on the final FS UUID, but perhaps they could be installed from the rootfs recipe and reconfigured later.

updated in latest patch

lool (changes requested):
Nice work! Making great progress here

I've taken a brief look, but I can tell I will want to give this a more formal review for which I don't have the time right now; I have a long train journey tomorrow so I'll tentatively do it then; the immediate things I spotted:
- I like the doc update
- the commit message is out of date (no mention of image), not in the style of the project, and not formal enough; you're making an important change, explain why, how, and polish the sentences and punctuation to deliver this cool feature
- I don't like the sed back and forth on APT configs with sed; try to generate them exactly once for the target archive/mirror, back that up, then generate them once for the snapshot URL, then move the backups back and forth; don't repeat yourself, use a wrapper / script

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -30,8 +28,12 @@ actions:
       - contrib
       - non-free
       - non-free-firmware
-    mirrors: ['http://deb.debian.org/debian']
+    mirrors: [{{if and $snapshot (ne $snapshot "true")}}'http://snapshot.debian.org/archive/debian/{{ $snapshot }}'{{else}}'http://deb.debian.org/debian'{{end}}]
     variant: minbase
+{{- if and $snapshot (ne $snapshot "true") }}
+    aptopts:
+      - 'Acquire::Check-Valid-Until "false"'
```
lool: This is necessary to use snapshot.debian.org after a few days, but we have to be conscious that this removes the protection against reply attacks. I think, but I'm not sure, that Debusine based snapshots would avoid the need for this workaround What I'd like to suggest:
- let's use https for snapshot.debian.org; this breaks caching with e.g. apt-cacher-ng
- let's open an issue and add a comment above this line that we want to revert this once we have Debusine based snapshots

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,6 +78,25 @@ actions:
...
+      set -eux
+      SNAPSHOT_DATE=$(grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2)
+      sed -i "s|http://deb.debian.org/debian/|http://snapshot.debian.org/archive/debian/${SNAPSHOT_DATE}/|g" \
+          /etc/apt/sources.list.d/debian.sources
+      sed -i "s|http://deb.debian.org/debian-security/|http://snapshot.debian.org/archive/debian-security/${SNAPSHOT_DATE}/|g" \
+          /etc/apt/sources.list.d/debian.sources
+      if [ -f /etc/apt/sources.list.d/debian-backports.sources ]; then
+        sed -i "s|http://deb.debian.org/debian|http://snapshot.debian.org/archive/debian/${SNAPSHOT_DATE}|g" \
+            /etc/apt/sources.list.d/debian-backports.sources
+      fi
```
gagath: Maybe the amount of `sed` can be reduced by having both `debian-snapshot-YYYYMMDD.sources` and `debian.sources` coexist, and the `sed` call is only done to change the value of `Enabled`. Ditto for backports and security.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -30,8 +28,12 @@ actions:
       - contrib
       - non-free
       - non-free-firmware
-    mirrors: ['http://deb.debian.org/debian']
+    mirrors: [{{if and $snapshot (ne $snapshot "true")}}'http://snapshot.debian.org/archive/debian/{{ $snapshot }}'{{else}}'http://deb.debian.org/debian'{{end}}]
```
gagath: This can be split in multiple lines to ease readability, ie create a if/else block and put the "mirrors" lines in the block.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -65,10 +67,22 @@ actions:
...
+      # snapshot:true → today's UTC timestamp; snapshot:<id> → use as-is
+      if [ "{{ $snapshot }}" = "true" ]; then
+        SNAPSHOT_DATE=$(date -u +%Y%m%dT%H%M%SZ)
+      else
+        SNAPSHOT_DATE="{{ $snapshot }}"
+      fi
+      touch /etc/buildinfo && chmod 644 /etc/buildinfo
+      echo "SNAPSHOT=${SNAPSHOT_DATE}" >> /etc/buildinfo
+      # disable Valid-Until check for snapshot repos
+      echo 'Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf.d/99snapshot
```
gagath: If this fits into snapshot-debian.sources even better, so it does still check valid until for debian.sources.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,6 +78,25 @@ actions:
...
     append-kernel-cmdline: clk_ignore_unused pd_ignore_unused audit=0 deferred_probe_timeout=30
 
+{{- if $snapshot }}
+  - action: run
+    description: Pin APT sources to snapshot for image build
+    chroot: true
+    command: |
+      set -eux
+      SNAPSHOT_DATE=$(grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2)
+      sed -i "s|http://deb.debian.org/debian/|http://snapshot.debian.org/archive/debian/${SNAPSHOT_DATE}/|g" \
```
lool: There are a couple of issues with these seds:
- the same calls are duplicated a few times
- there is on guarantee that we've transformed all sources to a snapshot capable source

This is the proposal I would have instead:
- create a new script to transform APT sources from regular to snapshot; e.g. make-apt-sources-snapshots; it would have a map of normal archive => snapshot archive
- backup all APT sources before calling the script (cp /etc/apt/sources.list.d/)
- call this script on all *.sources files, after a "modernize-sources" call to get rid of *.list entries
- make it fail if any of the archives does not support snapshots (e.g. it should fail with qartifactory); unless some forcing/override flag is present (which we'll need to use for now)
- call this script from both the rootfs and image recipe
- at the end of the rootfs and image recipes, restore the backups

But Agathe just shared another idea with me that she will pass here, which is perhaps simpler:
- pick a convention for the snapshot version of each *.sources, e.g. snapshot_debian.sources
- maintain these next to each APT sources we create
- set Enabled: yes, or no on the snapshots_* files/on the other files at the start/end of rootfs/image recipes

lool (changes requested):
Please update and polish the commit message - other comments inline

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -65,10 +67,22 @@ actions:
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
+      # snapshot:true → today's UTC timestamp; snapshot:<id> → use as-is
```
lool: Hi @vishwasudupa, we discussed the concept of building against "today"'s snapshot with @gagath, and, first it should have used the string "today" rather than "true", but more importantly we think we should keep it out of this PR and clarify the goals first. If it's only for our own CI, then perhaps it's simpler to just set that date in our github workflow; we could do that in a separate commit or even PR.

lool (changes requested):
Another pass on "today"

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: I wanted to treat snapshot even as a binary variable like (true/false). hence kept it as true
probably yes today also makes better sence for user, I will add today
will also try to add a check id anything other than today/date/valid id, fail out

debos-recipes/qualcomm-linux-debian-image.yaml
vishwasudupa: snapshot.sources seam simple and straight forward to me
will experiment on this approach and get back

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: I do not like one variable being able to be multiple types. What is the use of the "today" value and the mechanics you have added around it? I think we are not going to need it and it complexifies the code by having a different behavior than when using dates by not setting the `Acquire::Check-Valid-Until "false"` when it is set.

I think you should just remove the support for the "today" value and treat the variable as a date only.

>  will also try to add a check id anything other than today/date/valid id, fail out

I do not think it is necessary if we only allow dates, and will also complexify the code. The recipe will already crash when calling `apt` if something not supported by snapshots is used.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: I am not happy at all about this workaround. For the moment I would prefer our images to expire after something like 15 days rather than have this enabled. It also enables it for the whole Debian archive, which will allow replay attacks on deb.debian.org using http.

Let’s wait for Debusine support and have images expire in the meantime to not compromise on security?

debos-recipes/qualcomm-linux-debian-image.yaml
vishwasudupa: this was a intresting solution, i have implemented in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: this line is removed in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: yes, was able to fit in .sources

debos-recipes/qualcomm-linux-debian-image.yaml
vishwasudupa: I have updated in snapshot-debian.sources

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: today, was mostly a developer freedom for him to build on that day , for sure determined with snapshots. 
though not giving snapshot means it is building today, but version mentioned by debian might change from morning to evening

so when running in snapshot day, wanted to make sure the debian versions are pinned when someone wants to try reproducibility

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,6 +78,22 @@ actions:
...
+{{- if $snapshot }}
+  - action: run
+    description: Enable snapshot sources for image build
+    chroot: true
+    command: |
+      set -eux
+      # snapshot-debian.sources already has Enabled: no — flip to yes
+      sed -i 's/Enabled: no/Enabled: yes/' /etc/apt/sources.list.d/snapshot-debian.sources
+      sed -i 's/Enabled: yes/Enabled: no/' /etc/apt/sources.list.d/debian.sources
+      [ -f /etc/apt/sources.list.d/snapshot-debian-backports.sources ] && \
```
gagath: Since you are breaking this chaining over multiple lines given how long it is, I think you should use an `if/fi` block here. Cosmetic, but makes the code more readable IMO.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,6 +78,22 @@ actions:
...
+          sed -i 's/Enabled: no/Enabled: yes/' /etc/apt/sources.list.d/snapshot-debian-backports.sources || true
```
gagath: What is the reason you ignore failure with `|| true` here?

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,6 +78,22 @@ actions:
...
+      [ -f /etc/apt/sources.list.d/debian-backports.sources ] && \
+          sed -i 's/Enabled: yes/Enabled: no/' /etc/apt/sources.list.d/debian-backports.sources || true
```
gagath: Ditto, why ignore failure to perform the sed?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -564,6 +528,22 @@ actions:
...
 
+{{- if $snapshot }}
+  - action: run
+    description: Restore APT sources to live mirrors
+    chroot: true
+    command: |
+      set -eux
+      # Re-enable regular sources, disable snapshot sources via Enabled: toggle
+      sed -i 's/Enabled: no/Enabled: yes/' /etc/apt/sources.list.d/debian.sources
+      [ -f /etc/apt/sources.list.d/debian-backports.sources ] && \
```
gagath: This could use a `for` loop:

```bash
for f in debian-backports snapshot-debian snapshot-debian-backports
do
    if [ -f ... ]
    then
        sed -i ...
    fi
done
```

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -65,13 +63,25 @@ actions:
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
```
gagath, line 93: Still not happy about the support for "today". And the name is wrong because you specify the time as well so the correct name would be "now" instead of "today".

Is it really necessary? As a developer I would much prefer to pass the date of yesterday manually with "-t snapshot:20260528" and remove all of this logic from the recipe. Creating snapshot images is not something you should do often, so the need to automate the support for "today" or "now" is not there IMO.

Please remove this.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -81,22 +91,61 @@ actions:
...
       Components: main contrib non-free non-free-firmware
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
+      Enabled: yes
 {{- end }}
       EOF
 
+{{- if $snapshot }}
+      # Disable regular sources so snapshot sources take over
+      sed -i 's/Enabled: yes/Enabled: no/' /etc/apt/sources.list.d/debian.sources
+      # Snapshot sources — Enabled: yes; Check-Valid-Until scoped to this file
```
gagath: Please do not put comments that repeat what is already present in the code. This comment brings nothing.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -106,6 +155,21 @@ actions:
       Enabled: yes
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
       EOF
+
+{{- if $snapshot }}
+      # Disable regular backports so snapshot backports take over
+      sed -i 's/Enabled: yes/Enabled: no/' /etc/apt/sources.list.d/debian-backports.sources
+      # Snapshot backports — Enabled: yes; Check-Valid-Until scoped to this file
```
gagath: Please do not put comments that repeat what is already present in the code. This comment brings nothing.

README.md
```diff
@@ -145,13 +145,16 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
...
 - `xfcedesktop`: install an Xfce desktop environment; default: console only
   environment
-- `gnomedesktop`: install a GNOME desktop environment; default: console only environment
 - `overlays`: a `,`-separated list of rootfs overlays to add from
   `debos-recipes/overlays/`. See the *Supported overlays* section below.
 - `kernelpackage`: name of the kernel package to install from apt; defaults to
   `Debian’s linux-image-arm64`. Can (and should) be set to `none` if you are
   providing local kernel package instead.
 - `suite`: Debian suite to use, defaults to `trixie`.
+- `snapshot`: pin all apt sources to a Debian snapshot for a reproducible
```
gagath: This description is confusing, pinning is a specific term used in Debian packaging and what is done here is not package pinning.

README.md
```diff
@@ -145,13 +145,16 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
...
+  build. Accepts `true` (today's date), `YYYYMMDD`, or `YYYYMMDDTHHMMSSz`.
+  Sources are restored to live mirrors after the build; the date is logged to
```
gagath: Put the name of the variable used in /etc/buildinfo to not have users to guess it.

gagath (changes requested):
Thanks for using a separate .source file, this is starting to look good!

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: I am not convinced by this explanation. Please remove support for the "today" value (see comment in next review).

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: The whole point of this is to be able to recreate older images; 15d is not a large enough window

I'll be happy to use Debusine as well, I'm personally happy with the compromise in the meantime

gagath: The new implementation seems less worse by not enabling it for the whole Debian archive. So I think this is now missing https as you mentioned for snapshots, as well as an issue and a comment to detail what we would like to achieve in the long run.

debos-recipes/qualcomm-linux-debian-image.yaml
vishwasudupa: I have added for loop and added with if,fi in latest patch

debos-recipes/qualcomm-linux-debian-image.yaml
vishwasudupa: true is there so the whole command never fails, even if the file is missing or sed fails., it was to avoid errors, when file was not present

but now in latest patch, with if,fi, this is removed

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: added in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: user, can directly gives todays date, hence i have removed today option in latest patch

README.md
vishwasudupa: I have improvised in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -65,13 +66,23 @@ actions:
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
+      if echo "{{ $snapshot }}" | grep -qE '^([0-9]{8}|[0-9]{8}T[0-9]{6}Z)$'; then
+        SNAPSHOT_DATE="{{ $snapshot }}"
+      else
+        echo "ERROR: invalid snapshot '{{ $snapshot }}'. Use: YYYYMMDD | YYYYMMDDTHHMMSSz"; exit 1
```
gagath: 1. This error message says to use lowercase "z" but it needs to be uppercase "Z".
2. Error messages should go to stderr by convention instead of stdout, use >&2.
3. Do we really need this? The apt-update will fail if the format is wrong. Just let it crash and remove this check.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -141,6 +162,25 @@ actions:
       # enable unit
       systemctl enable debos-grow-rootfs
```
gagath, line 153: This is copy/pasted in multiple recipes, maybe it should live in a script in `scripts/disable-snapshot-sources.sh` or something similar to avoid duplication.

gagath: @vishwasudupa You did not answer to this?

README.md
```diff
@@ -152,6 +152,9 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
   `Debian’s linux-image-arm64`. Can (and should) be set to `none` if you are
   providing local kernel package instead.
 - `suite`: Debian suite to use, defaults to `trixie`.
+- `snapshot`: use a Debian snapshot archive for a reproducible build (`YYYYMMDD`
+  or `YYYYMMDDTHHMMSSz`); logged to `/etc/buildinfo` as `SNAPSHOT=<date>`.
```
gagath: Uppercase Z.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -81,22 +92,60 @@ actions:
...
+
 {{- if and (ne $suite "unstable") (ne $suite "testing") }}
   - action: run
     description: Add Debian {{ $suite }}-backports APT configuration
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
+      SNAPSHOT_DATE=$(grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2)
+{{- end }}
```
gagath, line 152: Do we really need this, instead of reusing the `{{ $snapshot }}` value?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -106,6 +155,20 @@ actions:
       Enabled: yes
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
       EOF
+
+{{- if $snapshot }}
+      # Disable regular backports so snapshot backports take over
+      sed -i 's/Enabled: yes/Enabled: no/' /etc/apt/sources.list.d/debian-backports.sources
+      tee /etc/apt/sources.list.d/snapshot-debian-backports.sources <<EOF
```
gagath: Add a blank line to separate the sed call from the tee call for clarity.

gagath (changes requested):

gagath:
Please also make sure your commit message respects the [conventional commit convention](https://www.conventionalcommits.org/en/v1.0.0/). This is not the first time we ask for this, as it is the convention used in the repo as you might have observed.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: in case users does wrong format, it is warned earlier, that is why i added this check
i have implemented 1 and 2 in latest patch

debos-recipes/qualcomm-linux-debian-image.yaml
vishwasudupa: sed disable is only done 2 times
can add it as function, but i believe it is good to do it at 3rd instance in future, because there is not enough use cases to make it generic with only 2

README.md
vishwasudupa: updated in latest patch

vishwasudupa:
> Please also make sure your commit message respects the [conventional commit convention](https://www.conventionalcommits.org/en/v1.0.0/). This is not the first time we ask for this, as it is the convention used in the repo as you might have observed.

updated as feat, suggest me if i have to make it more changes

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: yes
 the variable SNAPSHOT_DATE does not persist across blocks of yaml
 hence to populate in this section which is only present for unstable branch enablement this is needed

debos-recipes/qualcomm-linux-debian-image.yaml
gagath: I agree in the end. Thanks.

gagath (approved):
Thanks for your patience. I have no other comments for this PR and am fine with it. @lool please proceed to final review.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,6 +78,26 @@ actions:
...
+          if [ -f /etc/apt/sources.list.d/${f}.sources ]; then
+              sed -i 's/Enabled: no/Enabled: yes/' /etc/apt/sources.list.d/${f}.sources
+          fi
+      done
+      for f in debian debian-backports; do
+          if [ -f /etc/apt/sources.list.d/${f}.sources ]; then
+              sed -i 's/Enabled: yes/Enabled: no/' /etc/apt/sources.list.d/${f}.sources
+          fi
+      done
+      apt-get update
```
lool, line 88: I like this approach, except for the hardcoding of specific .sources files. I would prefer, as suggested in your commit message, to use snapshot_* files for the snapshots.

The logic to enable snapshots would be like (untested):
```
for f in /etc/apt/sources.list.d/*.sources; do
    enable="no"
    case $f in
      snapshot_*)
        enable="yes"
      ;;
    esac
    if ! grep -q ^Enabled:; then
        echo "Warning: no Enabled line in ${f}"
    fi
    sed -i "s/^Enabled:.*/Enabled: ${enable}/" "$f"
done
```

Note that I added a warning there.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -141,6 +162,25 @@ actions:
...
+{{- if $snapshot }}
+  - action: run
+    description: Restore APT sources to live mirrors
+    chroot: true
+    command: |
+      set -eux
+      # Re-enable regular sources, disable snapshot sources via Enabled: toggle
+      for f in debian debian-backports; do
+          if [ -f /etc/apt/sources.list.d/${f}.sources ]; then
+              sed -i 's/Enabled: no/Enabled: yes/' /etc/apt/sources.list.d/${f}.sources
```
lool: regex should be anchored (`^Enabled:`)

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -141,6 +162,25 @@ actions:
...
       systemctl enable debos-grow-rootfs
 
+{{- if $snapshot }}
+  - action: run
+    description: Restore APT sources to live mirrors
+    chroot: true
+    command: |
+      set -eux
+      # Re-enable regular sources, disable snapshot sources via Enabled: toggle
+      for f in debian debian-backports; do
```
lool: I don't like that this fails silently if there are no Enabled: lines, or if it's somehow not formatted as the regex suggests.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -65,13 +66,23 @@ actions:
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
+      if echo "{{ $snapshot }}" | grep -qE '^([0-9]{8}|[0-9]{8}T[0-9]{6}Z)$'; then
+        SNAPSHOT_DATE="{{ $snapshot }}"
+      else
+        echo "ERROR: invalid snapshot '{{ $snapshot }}'. Use: YYYYMMDD | YYYYMMDDTHHMMSSZ" >&2; exit 1
```
lool: exit 1 on separate line please

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -65,13 +66,23 @@ actions:
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
+      if echo "{{ $snapshot }}" | grep -qE '^([0-9]{8}|[0-9]{8}T[0-9]{6}Z)$'; then
+        SNAPSHOT_DATE="{{ $snapshot }}"
```
lool: Let's perhaps just call this SNAPSHOT, since that's what we use in buildinfo?

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -77,6 +78,26 @@ actions:
     setup-fstab: true
     append-kernel-cmdline: clk_ignore_unused pd_ignore_unused audit=0 deferred_probe_timeout=30
 
+{{- if $snapshot }}
+  - action: run
+    description: Enable snapshot sources for image build
+    chroot: true
+    command: |
+      set -eux
+      # Enable snapshot sources, disable regular sources via Enabled: toggle
```
lool: This is more of a note to myself:

While using Enabled: is handy for an end-user, it would also be possible to rename files to .disabled or similar. It's probably the same complexity to implement one or the other, but using .disabled would keep our sources shorter.

However, another advantage of Enabled is that we can warn if a file doesn't have this stanza.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -81,22 +92,61 @@ actions:
...
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
+      Enabled: yes
 {{- end }}
       EOF
 
+{{- if $snapshot }}
+      # Disable regular sources so snapshot sources take over
+      sed -i 's/Enabled: yes/Enabled: no/' /etc/apt/sources.list.d/debian.sources
+
+      tee /etc/apt/sources.list.d/snapshot-debian.sources <<EOF
```
lool: Why not generate this file from /etc/apt/sources.list.d/debian.sources? e.g. with a single sed as follows (untested):
```
sed "s#//deb\.debian\.org/debian/#//snapshot.debian.org/archive/debian/${SNAPSHOT_DATE}/#" /etc/apt/sources.list.d/debian.sources >/etc/apt/sources.list.d/snapshot_debian.sources
```

This would avoid the two defnitions getting out of sync, and keep the recipe file shorter.

lool (changes requested):
I've gave this a full read; generally it looks nice! I would however like a few hopefully simple changes:
1) I thing the snapshot APT sources files should follow a naming pattern such as `snapshot_<name of file they replace>.sources`; this allows generalizing the implementation and permits other APT sources files to benefit from the snapshot enabling/disabling logic

2) while there are less sed calls overall, I still don't like the numerous sed calls, I would think we could have a single enable snapshots / disable other sources, or disable snapshots / enable other sources script that we call 4 times (start and end of each recipe)

3) I'd like to avoid duplication between regular sources and snapshot sources, by generating snapshot sources; in the future, we could make the snapshot server parametrizable too (not needed right now though)

4) I'd like some warning if it looks like some sources are there but not snapshot ready; we could warn if an APT sources file lacks Enabled: lines for instance. Ideally, we'd check if each sources file as a corresponding snapshot and fail if running with the snapshot parameters, but that's a bit more complex

debos-recipes/qualcomm-linux-debian-image.yaml
vishwasudupa: updated in latest patch

debos-recipes/qualcomm-linux-debian-image.yaml
vishwasudupa: added in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: added in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: updated in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: used sed to create snapshot sources

vishwasudupa:
> I've gave this a full read; generally it looks nice! I would however like a few hopefully simple changes:
> 
> 1. I thing the snapshot APT sources files should follow a naming pattern such as `snapshot_<name of file they replace>.sources`; this allows generalizing the implementation and permits other APT sources files to benefit from the snapshot enabling/disabling logic
> 2. while there are less sed calls overall, I still don't like the numerous sed calls, I would think we could have a single enable snapshots / disable other sources, or disable snapshots / enable other sources script that we call 4 times (start and end of each recipe)
> 3. I'd like to avoid duplication between regular sources and snapshot sources, by generating snapshot sources; in the future, we could make the snapshot server parametrizable too (not needed right now though)
> 4. I'd like some warning if it looks like some sources are there but not snapshot ready; we could warn if an APT sources file lacks Enabled: lines for instance. Ideally, we'd check if each sources file as a corresponding snapshot and fail if running with the snapshot parameters, but that's a bit more complex

I have implemented 1,2,4 in latest patch

for 2 : i had to create script in yaml, tried to keep in qcom-deb-images scripts, but pushing from this script to chroot from host was not straight forward, since chroot was not able to get path and pull the script in

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -81,22 +117,38 @@ actions:
...
       Components: main contrib non-free non-free-firmware
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
+      Enabled: yes
 {{- end }}
       EOF
 
+{{- if $snapshot }}
+      # Derive snapshot_debian.sources from debian.sources by rewriting mirror URLs
+      sed \
+        -e "s#//deb\.debian\.org/debian/#//snapshot.debian.org/archive/debian/${SNAPSHOT}/#g" \
```
lool: let's use https

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -106,6 +158,17 @@ actions:
       Enabled: yes
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
       EOF
+
+{{- if $snapshot }}
+      # Derive snapshot_debian-backports.sources from debian-backports.sources by rewriting mirror URLs
+      sed \
+        -e "s#//deb\.debian\.org/debian/#//snapshot.debian.org/archive/debian/${SNAPSHOT}/#g" \
```
lool: ditto, https

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -164,6 +227,14 @@ actions:
       set -eux
       # remove old style, incomplete sources, superseded in apt-sources overlay
       rm -vf /etc/apt/sources.list
+{{- if $snapshot }}
+      SNAPSHOT=$(grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2)
+      for f in /etc/apt/sources.list.d/*.sources; do
+          case "$f" in */snapshot_*) continue ;; esac
+          grep -q "^Enabled: yes" "$f" && \
+              echo "Warning: $(basename "$f" .sources) has no snapshot support; packages not pinned to ${SNAPSHOT}" >&2 || true
```
lool: please use if / then

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -564,6 +635,15 @@ actions:
...
 
+{{- if $snapshot }}
+  - action: run
+    description: Restore APT sources to live mirrors
+    chroot: true
+    command: |
+      set -eux
+      apt-snapshot-toggle disable
+{{- end }}
+
```
lool, line 649: Please remove apt-snapshot binary and snapshot sources

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: Please drop the part about "pinned to"; no APT pinning here

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -81,22 +117,38 @@ actions:
...
+      Enabled: yes
 {{- end }}
       EOF
 
+{{- if $snapshot }}
+      # Derive snapshot_debian.sources from debian.sources by rewriting mirror URLs
+      sed \
+        -e "s#//deb\.debian\.org/debian/#//snapshot.debian.org/archive/debian/${SNAPSHOT}/#g" \
+        -e "s#//deb\.debian\.org/debian-security/#//snapshot.debian.org/archive/debian-security/${SNAPSHOT}/#g" \
+        -e "/^Enabled:/a\\Check-Valid-Until: no" \
```
lool, line 138: Please add a comment that Check-Valid-Until is a workaround until Debusine snapshots are available

lool (changes requested):
Hey! I did another pass on this; I closed many threads that you've addressed; thanks!

Here is what remains IMO:
1) commit message: drop the capital after feat:, mention the debos recipes changes, explain the why (reproducibility) and the security trade-off
2) use https for snapshot.debian.org (tweak sed regexp)
3) add a comment above the "Check-Valid-Until: no" workaround to revert it once Debusine-based snapshots are available
4) please drop apt-snapshot-toggle and the snapshot_*.sources files after build, at the end of the image recipe.
5) `|| true` for warning logic: please use if/then block
6) reword the warning that says "not pinned to ${SNAPSHOT}"; "pinning" is a specific APT term and not what's happening here

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: updated in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: updated in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: updated in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: these is again used in image yaml, 
it has been removed there at end

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: updated in latest patch

debos-recipes/qualcomm-linux-debian-rootfs.yaml
vishwasudupa: updated in latest patch

debos-recipes/qualcomm-linux-debian-image.yaml
vishwasudupa: all enabled have now been anchored

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -81,22 +117,39 @@ actions:
...
       Components: main contrib non-free non-free-firmware
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
+      Enabled: yes
 {{- end }}
       EOF
 
+{{- if $snapshot }}
+      # Derive snapshot_debian.sources from debian.sources by rewriting mirror URLs
+      sed \
+        -e "s#http://deb\.debian\.org/debian/#https://snapshot.debian.org/archive/debian/${SNAPSHOT}/#g" \
```
lool, line 135: could have allowed https with https?://

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -81,22 +117,39 @@ actions:
...
+        -e "s#http://deb\.debian\.org/debian-security/#https://snapshot.debian.org/archive/debian-security/${SNAPSHOT}/#g" \
+        -e "/^Enabled:/a\\# TODO: remove Check-Valid-Until once Debusine-based snapshots are available" \
```
lool, line 137: Ah I thought the comment would be in the recipe, not the APT sources file, but it's ok I guess we can spot it there too

lool (approved):
Thanks for the updates, looks good! I've spotted some new minor things, I'll will still merge as is

Issue #424, opened 2026-05-23
RB1 testing of forky


lool:
Hi

Currently RB1 testing fails on forky as the kernel doesn't have the right configs turned on; this should be fixed when a linux kernel higher than 7.1~rc3 makes it into forky:
https://tracker.debian.org/news/1751344/accepted-linux-71rc3-1exp1-source-into-experimental/

For now, we should disable RB1 testing with forky, and keep this issue open to re-enable it later.

gagath:
After [testing 7.1.3+deb14-arm64 using LAVA](https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/29749773042) the [crypto self-tests fail and the kernel panics as a result](https://lava.infra.foundries.io/scheduler/job/326025#L2707). There is an [ongoing patch to fix the crypto self-tests on SM8650](https://lore.kernel.org/all/20260717-qce-fix-self-tests-v6-0-455775fe5f6c@oss.qualcomm.com/) that is not in -next yet so it cannot be imported in the Debian kernel at the moment. This will have to wait a bit more.

lool:
Potential workaround: module_blacklist=qcrypto (preferred) or cryptomgr.notests

obbardc:
still relevant; also @gagath do you want to offload this to me?

gagath:
Sure, thanks, reassigned

PR #425, opened 2026-05-23
ci: skip RB1 boot tests on forky

lool:
forky's kernel doesn't yet have the configs needed for RB1 to boot.
Make boards_exclude in build-daily.yml's test matrix conditional on
the suite so qrb2210-rb1 is skipped only on forky; trixie testing is
unaffected.

Refs: #424

### DIFF
--- a/.github/workflows/build-daily.yml
+++ b/.github/workflows/build-daily.yml
@@ -39,4 +39,5 @@ jobs:
     with:
       url: ${{ needs.build-daily.outputs.artifacts_url }}
       suite: ${{ matrix.suite }}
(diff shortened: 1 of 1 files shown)


gagath (approved):
The [Github documentation on expression evaluation](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#operators) is a bit bad, by not detailing operator precedence. Nevertheless, it shows that we can use parens for logical grouping, which will simplify the proposed PR and better convey the idea that we exclude rb1 only for forky.

Will push the simplification by amending your commit and then merge it.

PR #426, opened 2026-05-27
qualcomm-linux-debian-flash: update boot to 00065

apaigava:
Tech pack based release path and versionzing is for SPF agnostic paths across Glymur target. This enabled SP build label versioning of boot bins.

Known fixes:

Glymur
fix(tz): improve AC policy logic and ensure operations stay within timing and access constraints to prevent faults


lool (approved):
Mostly approving this based on the Glymur changes described in the commit message and the sha256sum matching as we don't have Glymur in the loop for testing on PR yet.

PR #427, opened 2026-05-28
ci: glymur/qcom-next: switch to controlled kernel tag consumption model

phanendra-qc:
Glymur/daily-qcom-next nightly currently consumes the latest tag from qcom-next.
Instead of tracking tags continuously, switch to a controlled update model: Kernel team will raise a PR to bump the tag only after validation of qcom-next tag on Debian.

### DIFF
--- a/.github/workflows/linux-daily-glymur.yml
+++ b/.github/workflows/linux-daily-glymur.yml
@@ -20,7 +20,7 @@ jobs:
     uses: ./.github/workflows/linux.yml
     with:
       kernel_name: glymur
-      build_linux_deb_extra_args: '--qcom-next prune.config qcom.config'
+      build_linux_deb_extra_args: '--qcom-next --ref qcom-next-7.1-rc4-20260524 prune.config qcom.config'
       debos_extra_args: '-t dtb:qcom/glymur-crd.dtb'
       # we hardcode the DT, and that's not compatible with the QEMU machine model
       skip_qemu_tests: true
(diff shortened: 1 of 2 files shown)


phanendra-qc:
@knaveen-qc  @sgaud-quic @nivasnaik @vishwasudupa

lool (changes requested):
Hi

I'm ok with approach and the diff; keep in mind that the Glymur daily build is transient; we should really be using the qcom-next daily build instead, so you probably want to patch that too...

Please fix your commit message to the in the style of the project and pass it through spell checking, it's hard to parse right now :-/

lool (approved):

gagath:
- The first commit uses "qcom->next" instead of "qcom-next" branch name.
- The two commits have an almost identical description, they could be merged into a single commit.
- The first commit does not respect the [conventional commit convention](https://www.conventionalcommits.org/en/v1.0.0/) by starting with the file name directly.
- The commit message lines are not wrapped to 79 characters.
- The commit message mentions "nightly" but we use the term "daily" so it should be used as well to reduce confusion.

This is overly pedantic, but you need to learn to write good commits. Do not take it personally, but as an opportunity to learn. Thanks!

gagath (changes requested):
See previous comment.

gagath:
CI passed, taking over the commit with the applied fixes since this was already approved.

PR #414, opened 2026-05-19
rootfs: Remove adbd package

phanendra-qc:
For qcom debian distributed targets, SSH is only recommended to use for testing and also For Glymur compute, ADB is not PORed. 

removing adbd package from rootfs , as it is leading to axiom usb device identifier failures.

PR #431, opened 2026-06-03
fix(rootfs) : Append soc serial number to USB product String

bmaturu18:
To uniquely identify a device across EDL mode qcom test tool relies on the USB product descriptor. The ADB serial cannot be used in all modes, but the SoC serial number is available consistently.

Include the SoC serial in the USB gadget product string (strings/0x409/product) so it can serve as a stable identifier across all boot modes.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -482,6 +482,25 @@ actions:
           update-locale LANG=en_US.UTF-8
       fi
 
+  - action: run
+    description: Configure USB gadget with serial number identification
+    chroot: true
+    command: |
+      set -eux
+
+      FILE="/usr/lib/android-sdk/platform-tools/adbd-usb-gadget"
+
+      if [ -e "$FILE" ]; then
+        sed -i '/^setup()/,/^}/ {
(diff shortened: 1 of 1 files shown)


lool (changes requested):
We don't patch files managed by dpkg; we should patch the adbd package to pick up the serial, see this effort:
https://salsa.debian.org/android-tools-team/android-platform-tools/-/merge_requests/10
mentioned from https://github.com/qualcomm-linux/qcom-deb-images/pull/414

bmaturu18:
Hello Loic, i am aware of the changes, until https://salsa.debian.org/android-tools-team/android-platform-tools/-/merge_requests/10 this change is merged thought of fixing via rootfs yaml, we are working on https://github.com/qualcomm-linux/pkg-android-platform-tools, once this is in good shape then we can close this issue..

lool:
Hi @bmaturu18 this should still be delivered as a deb package; we can do that with a branch of the Debian package; Phany has started that too, see the PR (Cc: @basak-qcom)

gagath:
At this point this PR, its content and its title are disconnected from what we discussed here. I have created issue #435 to summarize where we are and what needs to be done. Now closing this, please continue discussion in the issue.

PR #432, opened 2026-06-03
feat!(rootfs): do not use a default password

gagath:
Default passwords are a bit of a hassle to remember between "debian", "qcom" or anything else. Remove the default password altogether and continue to enforce a password on first login.

This reduces user friction when using the image on the first time while keeping the security in check by asking for a new password on first login.

This change is inspired by what is done in arduino-deb-images and aims to reduce this delta:
https://github.com/arduino/arduino-deb-images/blob/arduino/debos-recipes/qualcomm-linux-debian-rootfs-arduino-extra.yaml#L50

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,9 +264,8 @@ NB: It's also possible to run qdl from the host while the board is not connected
 
 #### Login
 
-Once the image has booted, you can log in as the `debian` user, with the
-default `debian` password. The image should then ask you to change this default
-password to a safe one.
+Once the image has booted, you can log in as the `debian` user. The image
+should then ask you to define a password for this user.
 
 ## Development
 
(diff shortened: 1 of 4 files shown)


lool (commented):
@gagath Seems fine to me, but I would have expected changes to the ci/lava/ files too

gagath:
> @gagath Seems fine to me, but I would have expected changes to the ci/lava/ files too

Good catch, I will work around the LAVA prompts and ask for a new review when LAVA works.

phanendra-qc:
@gagath  we need to get axiom post boot scripts also aligned to this before mainling , thanks.
@smuppand  @bhargav0610 FYA

bhargav0610:
> Agathe Porte (@gagath) we need to get axiom post boot scripts also aligned to this before mainling , thanks. Srikanth Muppandam (@smuppand) bhargav0610 FYA

Thanks for letting us know before only.

PR #433, opened 2026-06-03
docs: add AI-assisted contribution disclosure requirement

gagath:
Add a section in CONTRIBUTING.md to require disclosure of AI-assisted contributions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,3 +61,82 @@ Here are a few things you can do that will increase the likelihood of your pull 
   If you want to make multiple independent changes, please consider submitting them as separate pull requests.
 - Write a [good commit message](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
 - It's a good idea to arrange a discussion with other developers to ensure there is consensus on large features, architecture changes, and other core code changes. PR reviews will go much faster when there are no surprises.
+
+# AI-Assisted Contributions
+
+- Contributions that were developed with the help of generative AI tools (e.g.,
+  coding assistants, chat-based models, AI agents) must follow the guidelines
+  in this document.
+- These guidelines apply to all forms of contribution including code,
+  documentation, tests, and configuration.
+- AI tools assisting with development on this project must follow the standard
+  contribution process, including any project-specific guidelines, coding
+  style, and procedures documented in this repository.
+- **Fully autonomous submissions are not permitted.** A human contributor must
+  be actively involved in reviewing, validating, and taking ownership of every
+  change before it is submitted.
+
+## Licensing and Legal Requirements
+
+All contributions—whether human-written or AI-assisted—must comply with the
+project's licensing requirements:
+
+- Contributors are responsible for ensuring that AI-generated content does not
+  introduce code with incompatible licenses or infringe on third-party
+  intellectual property.
+- Do not knowingly submit AI output that reproduces copyrighted material
+  verbatim. If you suspect a suggestion originates from a specific copyrighted
+  source, do not include it.
+
+## DCO and Sign-off
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -61,3 +61,62 @@ Here are a few things you can do that will increase the likelihood of your pull
   If you want to make multiple independent changes, please consider submitting them as separate pull requests.
 - Write a [good commit message](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
 - It's a good idea to arrange a discussion with other developers to ensure there is consensus on large features, architecture changes, and other core code changes. PR reviews will go much faster when there are no surprises.
+
+# AI-Assisted Contributions
+
+* Contributions that were developed with the help of generative AI tools (e.g., coding assistants, chat-based models, AI agents) must follow the guidelines in this document.
+* These guidelines apply to all forms of contribution including code, documentation, tests, and configuration.
+* AI tools assisting with development on this project must follow the standard contribution process, including any project-specific guidelines, coding style, and procedures documented in this repository.
+* **Fully autonomous submissions are not permitted.** A human contributor must be actively involved in reviewing, validating, and taking ownership of every change before it is submitted.
```
lool: This is too verbose and sounds too much like some internal company guidelines; I would have expect this to be worded as a human to another human, or from one opensource developer to another. e.g. I would have thought of something like:

"Humans will be reviewing your changes, so AI-assisted changes should ultimately be delivered under the responsibility of a human and easy to review by humans (that perhaps will also be using AI tools!)"

CONTRIBUTING.md
```diff
@@ -61,3 +61,62 @@ Here are a few things you can do that will increase the likelihood of your pull
...
+
+## Licensing and Legal Requirements
+
+All contributions—whether human-written or AI-assisted—must comply with the project's licensing requirements:
+
+* All code must be compatible with the project's [license](LICENSE.txt) (e.g. BSD-3-Clause).
+* Use appropriate [SPDX license identifiers](https://spdx.dev/learn/handling-license-info/) in source files.
+* Contributors are responsible for ensuring that AI-generated content does not introduce code with incompatible licenses or infringe on third-party intellectual property.
+* Do not knowingly submit AI output that reproduces copyrighted material verbatim. If you suspect a suggestion originates from a specific copyrighted source, do not include it.
```
lool: I again think that is too verbose, we already convey the license of the code, and I don't think we need a subsection for SPDX in the AI specific section of the contribution guidelines.

CONTRIBUTING.md
```diff
@@ -61,3 +61,62 @@ Here are a few things you can do that will increase the likelihood of your pull
...
+
+## DCO and Sign-off
```
lool, line 91: Perhaps expand "Developer Certificate of Origin"(?)

lool (changes requested):
I liked reusing the Linux policy, I found the first part too verbose and not very human friendly; I made some suggestions on how we could perhaps use a different tone

I'm generally ok with the proposed policies

CONTRIBUTING.md
gagath: I re-read this and prefer the initial version.

Contributing to a project is not only submitting changes (your version) but also reviewing and validating (initial version). I am keen to keep the original wording and do not see a problem with it.

CONTRIBUTING.md
gagath: Dropped the two first bullet points.

CONTRIBUTING.md
gagath: There is a link to the DCO in the first paragraph, as well as the expansion of the acronym. I think that’s enough.

gagath (commented):
Wrapped everything and removed the SPDX section.

lool (approved):

PR #442, opened 2026-06-09
fix(debos/flash): copy over sail_nor when present

### LINKED ISSUE #437: Daily monza: Missing SAIL/SPINOR in trixie-flash-emmc
The generated trixie-flash-emmc package, does not contain the SPINOR elements, flashing the SPINOR is required to boot the SAIL domain, and so the whole system.

lool:
Flash dir assembly dropped the sail_nor payload directory from the
boot binaries, leaving out some SPI-NOR provisioning files for some
boards, notably monza.

The sail_nor directory holds standalone rawprogram/gpt/firehose/elf
files. Copy it into the per-board flash directory for any board
whose boot binaries ship it.

Of the public boot-binaries packages, only QCS8300 (Monaco) and
QCS9100 (LeMans) currently ship a sail_nor directory; Glymur,
QCM6490 and QCS615 ship none. The directory is always named exactly
"sail_nor" and sits at the top level of the unpacked boot binaries,
so a simple conditional copy works.

Fixes: #437

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -404,6 +404,14 @@ actions:
               \) \
               -exec cp --preserve=mode,timestamps -v '{}' "${flash_dir}" \;
 
+          # copy the sail_nor directory (standalone rawprogram/gpt/firehose/elf
+          # payloads) for SAIL domain flashing on SPI-NOR; only some silicon
+          # families ship it
+          sail_nor_dir="build/${board_name}_boot-binaries/sail_nor"
+          if [ -d "${sail_nor_dir}" ]; then
+              cp --preserve=mode,timestamps -rv "${sail_nor_dir}" "${flash_dir}/"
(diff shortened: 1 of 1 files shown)


gagath (approved):

gagath:
Maybe using hard links would save us some copy time with `--link`, my two cents, no idea how big these files are.

lool:
> Maybe using hard links would save us some copy time with `--link`, my two cents, no idea how big these files are.

Fairly small, I'm generally worried about hardlinks so tend to avoid them
```
lminier@LMINIER-mac bb % du -hs QCS8300_bootbinaries
 16M	QCS8300_bootbinaries
lminier@LMINIER-mac bb % du -hs QCS8300_bootbinaries/sail_nor
2.7M	QCS8300_bootbinaries/sail_nor
```

gagath:
> Fairly small, I'm generally worried about hardlinks so tend to avoid them

Ack. Hardlinks are cool though! At least when running Linux. :)

PR #443, opened 2026-06-13
fix(ci): don't require monaco-evk to be in cam

lool:
monaco-evk LAVA template searches for device with the cambrige-lab
tag, but all of them are actually in Hyderabad rather than
Cambridge. Drop the tag constraint and let LAVA select any device.


lool:
self-approving / merging as trivial and fixing CI

PR #445, opened 2026-06-14
HACK: insert missing dtbos into linux-image deb

lool:
Workaround dtbos missing from dtb install target by post-processing
the linux-image deb to add them. This is to explore solutions, but I
don't like this implementation.

### DIFF
--- a/scripts/build-linux-deb.py
+++ b/scripts/build-linux-deb.py
@@ -3,8 +3,11 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import argparse
+import hashlib
+import shutil
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 
 # git repo/ref to use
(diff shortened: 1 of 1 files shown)


scripts/build-linux-deb.py
```diff
@@ -154,6 +157,78 @@ def check_dependencies():
...
+
+    See the upstream fix that makes this unnecessary once merged
+    (lore.kernel.org message-id, Vishwas Udupa):
+        20260428123725.3457865-1-vudupa@qti.qualcomm.com
+    """
+    dts_qcom = linux_dir / "arch" / "arm64" / "boot" / "dts" / "qcom"
+    dtbos = sorted(dts_qcom.glob("*.dtbo"))
+    if not dtbos:
+        log_i("No qcom/*.dtbo found to add to the linux-image deb; skipping")
+        return
```
gagath, line 184: I think we should err out instead of "silently" continuing? Is there a case where the dtbos are not generated and we should continue?

scripts/build-linux-deb.py
```diff
@@ -154,6 +157,78 @@ def check_dependencies():
...
+    Until the upstream Makefile fix (which adds the overlays to dtb-y) lands,
+    repack the produced linux-image deb to add every already-built qcom/*.dtbo
+    under the same usr/lib/linux-image-<release>/qcom/ path dtbs_install uses.
+
+    See the upstream fix that makes this unnecessary once merged
+    (lore.kernel.org message-id, Vishwas Udupa):
+        20260428123725.3457865-1-vudupa@qti.qualcomm.com
+    """
+    dts_qcom = linux_dir / "arch" / "arm64" / "boot" / "dts" / "qcom"
+    dtbos = sorted(dts_qcom.glob("*.dtbo"))
```
gagath, line 181: Why sort?

scripts/build-linux-deb.py
```diff
@@ -154,6 +157,78 @@ def check_dependencies():
...
+    # the kernel release is what names both the deb and the install subdir
+    release = subprocess.check_output(
+        make_base_command + ["-s", "kernelrelease"],
+        cwd=linux_dir,
+        text=True,
+    ).strip()
+
+    # bindeb-pkg writes the debs to the parent of the source tree
+    deb_dir = linux_dir.resolve().parent
+    image_debs = sorted(deb_dir.glob(f"linux-image-{release}_*.deb"))
```
gagath, line 195: Why sort?

scripts/build-linux-deb.py
```diff
@@ -154,6 +157,78 @@ def check_dependencies():
...
+            (dst / dtbo.name).chmod(0o644)
+
+        # rebuild in place; dpkg-deb recomputes the control fields it owns
+        subprocess.run(
+            ["dpkg-deb", "-b", str(extract), str(image_deb)],
+            check=True,
+        )
+
+    sha = hashlib.sha256(image_deb.read_bytes()).hexdigest()
+    log_i(f"Repacked {image_deb.name} (sha256 {sha})")
```
gagath, line 229: What is the rationale for computing and displaying a sha here?

scripts/build-linux-deb.py
```diff
@@ -154,6 +157,78 @@ def check_dependencies():
...
+    image_debs = sorted(deb_dir.glob(f"linux-image-{release}_*.deb"))
+    image_debs = [d for d in image_debs if "-dbg_" not in d.name]
+    if len(image_debs) != 1:
+        fatal(
+            f"Expected exactly one linux-image-{release} deb in {deb_dir}, "
+            f"found: {[d.name for d in image_debs]}"
+        )
+    image_deb = image_debs[0]
+
+    install_subdir = f"usr/lib/linux-image-{release}/qcom"
```
gagath, line 204: We have Pathlib at our disposal here, may as well make use of it?

scripts/build-linux-deb.py
```diff
@@ -154,6 +157,78 @@ def check_dependencies():
         fatal(f"Missing build-dependencies: {' '.join(missing)}")
 
 
+def add_qcom_dtbos_to_image_deb(linux_dir, make_base_command):
+    """
+    Workaround: add qcom/ DTB overlays (*.dtbo) to the linux-image deb.
+
+    The kernel only installs (via dtbs_install) the DTBs listed in dtbs-list,
```
gagath, line 164: The "why" is verbose and fits in a Github issue more than a docstring IMO. In general docstrings should describe what the function is doing, not why. Comments can be used for the "why", but I would just put a comment that links to the tracking GH issue where this function is called.

gagath (changes requested):
Regarding HACK things, I think we need to have a bit more tracking. Some ideas:

- Create a Github issue that references the hack, why it is necessary, what is the upstream status, do we have a good feel that the proper solution will be merged upstream
- Regularly update this issue tracking the upstream progress
- Probably have a "HACK" tag to have the issue more visible
- The commit and code should refer to the Github issue and probably also have a HACK string in a comment so we can find it easily
- Have a link to the LKML patch instead of a reference for ease of consultation

lool:
> Regarding HACK things, I think we need to have a bit more tracking.

Apologies, I didn't mean to request a review – I use draft PRs to show an idea or, here, how something would look like.

There are some good ideas in your bullet list, but I don't want that commit to go in, I wanted to show that I find it too invasive.

PR #446, opened 2026-06-14
return to combined dtbs

### LINKED ISSUE #444: New FIT based dtb.bin lacks key DTBs
Hi

monaco-evk fails to boot with e.g. linux-next https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/27493606800

This is since the new feature of the flash recipe to allow using a FIT image, but it doesn't work with linux-next as it doesn't install any of the required dtbos for monaco-evk's configs.

There are a lot of issues compounding:
- upstream linux Makefile should install these; Vishwas sent a fix upstream [here](https://lore.kernel.org/all/20260428123725.3457865-1-vudupa@qti.qualcomm.com/) but it wasn't reviewed/acked; however, similar patches were included in qcom-next e.g. [this one](https://github.com/qualcomm-linux/kernel/commit/7998dbf1695da00ef7a54c3fe90328e251400c90)
- I thought we'd have an easy workaround for the qcom-deb-images kernel builds with a patch likes [this one from Vishwas](https://github.com/qualcomm-linux/qcom-deb-images/commit/ebada166abfcdb2213286eeb1c54bf1cbbdd842d), however it doesn't work
- qcom-dtb-metadata's data does NOT include a config without overlays for monaco-evk, which might or might not be a bug, I'm not sure yet
- its FIT image generation script is also not making any distinction between skipping a config due to missing overlays and skipping a silicon family entirely, which perhaps we need

I have a few ideas about interim workarounds until we install dtbos properly – both here and in qcom-dtb-metadata.

lool:
Return to combined DTBs to fix regressions due to introduction of FIT images.
- **fix(flash)!: combineddtb for monaco and lemans**
- **feat(flash)!: default dtb_bin_type to combineddtb**
- **chore(debos): Drop combineddtb from board entries**

Fixes: #444

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -50,6 +50,7 @@ actions:
     )
     "cdt_filename" "CRD/cdt_glymur_crd_0.1.0.bin"
     "dtb" "qcom/glymur-crd.dtb"
+    "dtb_bin_type" "multidtb"
 )}}
 {{- if eq $build_qcs615 "true" }}
 {{- $boards = append $boards (dict
@@ -72,7 +73,6 @@ actions:
     )
     "cdt_filename" "cdt_adp_air_sa6155p.bin"
     "dtb" "qcom/qcs615-ride.dtb"
-    "dtb_bin_type" "combineddtb"
 )}}
 {{- end }}
 {{- if eq $build_qcm6490 "true" }}
@@ -96,7 +96,6 @@ actions:
     )
     "cdt_filename" "cdt_vision_kit.bin"
     "dtb" "qcom/qcs6490-rb3gen2.dtb"
-    "dtb_bin_type" "combineddtb"
 )}}
 {{- end }}
 {{- if eq $build_qcs8300 "true" }}
@@ -120,7 +119,6 @@ actions:
     )
     "cdt_filename" "cdt_ride_sx.bin"
     "dtb" "qcom/qcs8300-ride.dtb"
-    "dtb_bin_type" "combineddtb"
 )}}
 {{- $boards = append $boards (dict
     "name" "monaco-evk"
@@ -186,7 +184,6 @@ actions:
     )
     "cdt_filename" "cdt_ride_sx.bin"
     "dtb" "qcom/qcs9100-ride-r3.dtb"
-    "dtb_bin_type" "combineddtb"
 )}}
 {{- $boards = append $boards (dict
     "name" "lemans-evk"
@@ -224,7 +221,6 @@ actions:
     )
     "u_boot_file" .u_boot_rb1
     "dtb" "qcom/qrb2210-rb1.dtb"
-    "dtb_bin_type" "combineddtb"
 )}}
 {{- end }}
 {{- $boards = append $boards (dict
@@ -239,7 +235,6 @@ actions:
         "sha256sum" "c606e95d0107f8c58d0dd9494e00624d1db7c4361cca20513bc78ef02ca28dd1"
     )
     "dtb" "qcom/qrb2210-arduino-imola.dtb"
-    "dtb_bin_type" "combineddtb"
 )}}
 
 {{- range $board := $boards }}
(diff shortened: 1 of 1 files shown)


debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -208,6 +211,7 @@ actions:
     )
     "cdt_filename" "cdt_rb8_core_kit.bin"
     "dtb" "qcom/lemans-evk.dtb"
+    "dtb_bin_type" "combineddtb"
```
gagath: So the default is now "combineddtb", but you set it here as well? I think something is wrong.

gagath (changes requested):

vishwasudupa:
i see this pr is making default as combineddtb

in that case we can remove all entries of  "dtb_bin_type" "combineddtb"

and for glymur, we can mention "dtb_bin_type" "multidtb"

 is qcs6490-rb3gen2-vision-kit, lemans-evk not facing issue with multidtb and linux-next kernel?

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -142,6 +143,7 @@ actions:
     )
     "cdt_filename" "cdt_qcs8275_iq_8275_evk_pro_sku.bin"
     "dtb" "qcom/monaco-evk.dtb"
+    "dtb_bin_type" "combineddtb"
```
vishwasudupa: since below default is changed to combineddtb
this can be removed

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -163,6 +165,7 @@ actions:
     )
     "cdt_filename" "cdt_qcs8275-Monza.bin"
     "dtb" "qcom/monaco-arduino-monza.dtb"
+    "dtb_bin_type" "combineddtb"
```
vishwasudupa: since below default is changed to combineddtb
this is redundant it can be removed

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: since below default is changed to combineddtb
this is redundant it can be removed

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: I've pushed this change as a new commit

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: I've pushed this change as a new commit, since both Vishwas and you prefer that I drop entries that set the value to the default

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: I've pushed this change as a new commit

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: I've pushed this change as a new commit

gagath: > I've pushed this change as a new commit

Please rework your commits so that the commit that changes the default also fixes this instead of introducing a new commit.

debos-recipes/qualcomm-linux-debian-flash.yaml
vishwasudupa: looks good now

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: I personally preferred it that way, but happy to squash

lool: done

gagath:
Now the second commit mostly overrides what is done in the first commit. Can you do all of these changes in a single commit please? Thanks.

lool:
> Now the second commit mostly overrides what is done in the first commit. Can you do all of these changes in a single commit please? Thanks.

done

PR #447, opened 2026-06-15
fix: dtb-multidtb.bin in spinor needs to be referenced 2 level above

vishwasudupa:
When dtb-multidtb is referenced from rawprograms in SPINOR, it needs to point two levels above. This is required to keep dtb-multidtb common across boards and to satisfy the SPINOR structure.

this fix is needed only for multidtb spinor directory structure, hence it is done post xml generation

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -509,6 +509,11 @@ actions:
                           -u "//download_file[file_name/text()='dtb-multidtb.bin']/file_path[text()='..']" -v "../../" \
                           "${copied_contents}"
                   fi
+                  # fix rawprogram*.xml in spinor dir: ../dtb-multidtb.bin -> ../../dtb-multidtb.bin
+                  find "${target_dir}/spinor" -name 'rawprogram*.xml' \
+                      -exec xmlstarlet ed -L \
+                          -u "//program[@filename='../dtb-multidtb.bin']/@filename" -v "../../dtb-multidtb.bin" \
+                      '{}' \;
               done
       done
 


lool (approved):
My initial intuition was that we should fix the filename passed to ptool, but then I realized we move the spinor directory around due to ptool data files not having the proper structure (see work from Igor on how this might be addressed). So that seems like a reasonable post-move update to do.

lool:
NB: I'd like to merge https://github.com/qualcomm-linux/qcom-deb-images/pull/446 before this one

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -509,6 +509,11 @@ actions:
                           -u "//download_file[file_name/text()='dtb-multidtb.bin']/file_path[text()='..']" -v "../../" \
                           "${copied_contents}"
                   fi
+                  # fix rawprogram*.xml in spinor dir: ../dtb-multidtb.bin -> ../../dtb-multidtb.bin
+                  find "${target_dir}/spinor" -name 'rawprogram*.xml' \
+                      -exec xmlstarlet ed -L \
+                          -u "//program[@filename='../dtb-multidtb.bin']/@filename" -v "../../dtb-multidtb.bin" \
```
gagath, line 515: This line is over-indented compared to the other twos.

gagath (approved):
Please fix the over-indented line, otherwise LGTM.

debos-recipes/qualcomm-linux-debian-flash.yaml
lool: Sorry, missed this comment before merging

lool: I think the idea was to convey that this is the continuation of the -exec, not the -find

PR #371, opened 2026-04-17
feature(Makefile): install missing dependencies in debos container

abhi5mand:
While running `make flash` on ARM machines, the base image `ghcr.io/go-debos/debos:latest` is missing required tools such as`mtools`, causing failures (e.g. `mcopy` not found).

Problem:
Developers using `CONTAINER_IMAGE ?= ghcr.io/go-debos/debos:latest` encounter errors because required dependencies are not present inside the container.

Solution:
Introduce `check_install_dependency.sh`, which:
- checks for missing dependencies
- generates `Dockerfile.auto-deps` on top of the base debos image
- installs only missing packages
- builds and returns a new container image when required

The script is invoked from the Makefile to automatically select the correct container image.

Usage:
  chmod +x check_install_dependency.sh

PR #454, opened 2026-06-18
feat(debos): validate host build dependencies per use-case

bjordiscollaku:
Nothing on main validates the host tools the builds need before running
debos, and the two hand-maintained apt lists have drifted: the README
build-deps line is missing device-tree-compiler, u-boot-tools and file
that the CI step installs. A local `make` with a tool missing fails
somewhere inside debos with a confusing error rather than a clear one up
front.

check-deps.sh now owns the package lists, keyed by use-case. The Makefile
validates the right use-case before each debos call (or installs it with
AUTO_INSTALL_DEPS=yes), CI and the README run one line,
`scripts/check-deps.sh --install host rootfs image flash test`, and no
package name appears anywhere else.

| use-case | packages |
|---|---|
| host   | debos, make, ipxe-qemu, linux-image-arm64 (arch-dependent) |
| rootfs | mmdebstrap, debian-archive-keyring |
| image  | dosfstools, e2fsprogs, parted, fdisk, udev |
| flash  | dosfstools, mtools, xmlstarlet, python3-defusedxml, unzip, device-tree-compiler, u-boot-tools, file, ca-certificates |
| test   | qemu-efi-aarch64, qemu-utils, ipxe-qemu, python3-pexpect, python3-pytest |

Validate reads the dpkg-query Status field and requires "ok installed"
rather than trusting `dpkg -s`, which also exits 0 for a package in
config-files or a half-configured state; the install path is a plain
`apt-get install -y --no-install-recommends`. qemu-system-arm stays off
the lists: it is a Depends of debos, so installing debos pulls it in.

Four packages are named that the old apt line never was; they reached the
build via apt Recommends and are dropped once installs use
--no-install-recommends. linux-image-arm64 is a Recommends of debos, the
kernel fakemachine boots from /boot (the CI job container ships none); the
script installs linux-image-$(dpkg --print-architecture), and only where
that meta-package resolves, since a normal host already has a kernel.
ipxe-qemu is a Recommends of qemu-system-arm and ships
/usr/share/qemu/efi-virtio.rom, without which qemu will not start the
default virtio-net-pci NIC, so fakemachine and the boot test fail at
startup. qemu-utils is a Recommends of qemu-system-arm and ships qemu-img,
which ci/qemu_test.py runs first to build the boot overlay.
ca-certificates' postinst runs update-ca-certificates to generate the CA
bundle /etc/ssl/certs/ca-certificates.crt, which debos (a Go binary)
verifies https downloads against; nothing in the install set Depends on
it. udev is in the image list for a different reason: debos's
image-partition execs udevadm trigger --settle directly. It is a
Recommends of debos, and its only other route is the kernel's
initramfs-tools alternative dependency, so image names it rather than
relying on either.

Checked by replaying the workflow in the debian:trixie CI container
(public.ecr.aws/debian/debian:trixie, arm64): union install; per-use-case
validate against present, missing and config-files states; and rootfs,
UFS image and flash builds under the qemu fakemachine backend, plus the
qemu boot test, all green in that container. That does not cover the
self-hosted arm64 runner environment (/efs, artifact upload); the CI run
on this PR exercises that.

Provisioning the debos container image stays out of scope, as does
folding in build-linux-deb.py and build-u-boot-rb1.sh.

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -73,10 +73,10 @@ jobs:
       - name: Copy U-Boot for RB1 from EFS
         run: cp -av "/efs/u-boot-rb1-latest/rb1-boot.img" .
 
-      # mtools is needed for the flash recipe
+      # dosfstools is needed for the image recipe, mtools is needed for the flash recipe
       # file, device-tree-compiler and u-boot-tools are needed by qcom-dtb-metadata
       - name: Install debos and dependencies of the recipes and local tests
-        run: apt -y install debian-archive-keyring debos file make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml device-tree-compiler u-boot-tools
+        run: apt -y install debian-archive-keyring debos dosfstools file make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml device-tree-compiler u-boot-tools
 
       - name: Setup local APT repo
         run: |
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -73,10 +73,10 @@ jobs:
       - name: Copy U-Boot for RB1 from EFS
         run: cp -av "/efs/u-boot-rb1-latest/rb1-boot.img" .
 
-      # dosfstools is needed for the image recipe, mtools is needed for the flash recipe
-      # file, device-tree-compiler and u-boot-tools are needed by qcom-dtb-metadata
-      - name: Install debos and dependencies of the recipes and local tests
-        run: apt -y install debian-archive-keyring debos dosfstools file make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml device-tree-compiler u-boot-tools
+      # check-deps.sh holds the package lists per use-case, so the recipes, CI
+      # and README share a single source of truth
+      - name: Install debos and recipe/test dependencies
+        run: scripts/check-deps.sh --install host rootfs image flash test
 
       - name: Setup local APT repo
         run: |
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ ifeq ($(USE_CONTAINER),yes)
 		--mount "type=bind,source=$(CURDIR),destination=$(DEBOS_WORKDIR)" \
 		--security-opt label=disable \
 		$(CONTAINER_IMAGE) \
-		$(DEBOS_OPTS)
+		$(DEBOS_OPTS) -t auto_install_deps:true
 else
 	# Working directory for native debos
 	DEBOS_WORKDIR := $(CURDIR)
--- a/Makefile
+++ b/Makefile
@@ -31,13 +31,18 @@ ifeq ($(USE_CONTAINER),yes)
 	KVM_DEVICE := $(if $(wildcard /dev/kvm),--device /dev/kvm)
 	# Working directory as seen from inside the container
 	DEBOS_WORKDIR := /recipes
+	# The debos container is a disposable instance, so it is safe to let the
+	# recipes auto-install their missing host dependencies into it. Pass the
+	# flag before $(DEBOS_OPTS) so that EXTRA_DEBOS_OPTS can still override it:
+	# debos applies the last -t value given for a key.
 	DEBOS_CMD := docker run --rm --interactive --tty \
 		$(KVM_DEVICE) \
 		--user $(shell id -u) --workdir $(DEBOS_WORKDIR) \
 		--mount "type=bind,source=$(CURDIR),destination=$(DEBOS_WORKDIR)" \
 		--security-opt label=disable \
 		$(CONTAINER_IMAGE) \
-		$(DEBOS_OPTS) -t auto_install_deps:true
+		-t auto_install_deps:true \
+		$(DEBOS_OPTS)
 else
 	# Working directory for native debos
 	DEBOS_WORKDIR := $(CURDIR)
--- a/Makefile
+++ b/Makefile
@@ -31,22 +31,38 @@ ifeq ($(USE_CONTAINER),yes)
 	KVM_DEVICE := $(if $(wildcard /dev/kvm),--device /dev/kvm)
 	# Working directory as seen from inside the container
 	DEBOS_WORKDIR := /recipes
-	# The debos container is a disposable instance, so it is safe to let the
-	# recipes auto-install their missing host dependencies into it. Pass the
-	# flag before $(DEBOS_OPTS) so that EXTRA_DEBOS_OPTS can still override it:
-	# debos applies the last -t value given for a key.
 	DEBOS_CMD := docker run --rm --interactive --tty \
 		$(KVM_DEVICE) \
 		--user $(shell id -u) --workdir $(DEBOS_WORKDIR) \
 		--mount "type=bind,source=$(CURDIR),destination=$(DEBOS_WORKDIR)" \
 		--security-opt label=disable \
 		$(CONTAINER_IMAGE) \
-		-t auto_install_deps:true \
 		$(DEBOS_OPTS)
 else
 	# Working directory for native debos
 	DEBOS_WORKDIR := $(CURDIR)
 	DEBOS_CMD := debos $(DEBOS_OPTS)
+endif
+
+# AUTO_INSTALL_DEPS=yes lets the dependency check install missing packages
+# instead of failing; off by default so a local build never installs without
+# being asked. CI installs everything up front instead (see debos.yml); the
+# throwaway container can't install (non-root), so the check is skipped there.
+AUTO_INSTALL_DEPS ?= no
+CHECK_INSTALL := $(if $(filter yes,$(AUTO_INSTALL_DEPS)),--install,)
+
+# $(call run_debos,<use-case>,<recipe>[,<extra debos opts>])
+# Native builds validate (or install) the use-case's host dependencies before
+# running debos. The container path runs debos directly; provisioning the
+# throwaway container's tools is a separate, image-level concern.
+ifeq ($(USE_CONTAINER),yes)
+define run_debos
+$(DEBOS_CMD) $(3) $(2)
+endef
+else
+define run_debos
+scripts/check-deps.sh $(CHECK_INSTALL) $(1) && $(DEBOS_CMD) $(3) $(2)
+endef
 endif
 
 # Use http_proxy from the environment, or apt's http_proxy if set, to speed up
@@ -58,30 +74,30 @@ export http_proxy
 all: disk-ufs.img disk-sdcard.img
 
 rootfs.tar dtbs.tar.gz: debos-recipes/qualcomm-linux-debian-rootfs.yaml
-	$(DEBOS_CMD) $<
+	$(call run_debos,rootfs,$<)
 
 DISK_UFS_IMAGES := disk-ufs.img \
 	disk-ufs.img1 \
 	disk-ufs.img2
 
 $(DISK_UFS_IMAGES): debos-recipes/qualcomm-linux-debian-image.yaml rootfs.tar
-	$(DEBOS_CMD) $<
+	$(call run_debos,image,$<)
 
 DISK_SDCARD_IMAGES := disk-sdcard.img \
 	disk-sdcard.img1 \
 	disk-sdcard.img2
 
 $(DISK_SDCARD_IMAGES): debos-recipes/qualcomm-linux-debian-image.yaml rootfs.tar
-	$(DEBOS_CMD) -t imagetype:sdcard $<
+	$(call run_debos,image,$<,-t imagetype:sdcard)
 
 .PHONY: flash
 flash: debos-recipes/qualcomm-linux-debian-flash.yaml dtbs.tar.gz
-	$(DEBOS_CMD) $<
+	$(call run_debos,flash,$<)
 
 .PHONY: test
 test: disk-ufs.img
 	# rootfs/ is a build artifact, so should not be scanned for tests
-	py.test-3 --ignore=rootfs
+	scripts/check-deps.sh $(CHECK_INSTALL) test && py.test-3 --ignore=rootfs
 
 .PHONY: clean
 clean:
--- a/debos-recipes/qualcomm-linux-debian-image.yaml
+++ b/debos-recipes/qualcomm-linux-debian-image.yaml
@@ -15,7 +15,7 @@ actions:
     command: |
       "${RECIPEDIR}/../scripts/check-deps.sh" \
           {{if eq $auto_install_deps "true"}}--install{{end}} \
-          dosfstools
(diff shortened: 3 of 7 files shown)


Makefile
```diff
@@ -37,7 +37,7 @@ ifeq ($(USE_CONTAINER),yes)
 		--mount "type=bind,source=$(CURDIR),destination=$(DEBOS_WORKDIR)" \
 		--security-opt label=disable \
 		$(CONTAINER_IMAGE) \
-		$(DEBOS_OPTS)
+		$(DEBOS_OPTS) -t auto_install_deps:true
```
lool: Can you add a comment explaining that it's fine in the container instance to do that?

Also please move this flag before DEBOS_OPTS so that it can be user disabled

scripts/check-deps.sh
```diff
@@ -0,0 +1,36 @@
...
+# If dpkg is not available (e.g. inside a fakemachine qemu VM), skip silently.
+# Dependencies are expected to be present on the real host in that case.
+[ -f /var/lib/dpkg/status ] || exit 0
+
+missing=""
+for pkg in "$@"; do
+    dpkg -s "$pkg" >/dev/null 2>&1 || missing="$missing $pkg"
+done
+
+[ -z "$missing" ] && exit 0
```
lool: please expand as if/then for readability

scripts/check-deps.sh
```diff
@@ -0,0 +1,36 @@
...
+
+install=false
+if [ "${1:-}" = "--install" ]; then
+    install=true
+    shift
+fi
+
+# If dpkg is not available (e.g. inside a fakemachine qemu VM), skip silently.
+# Dependencies are expected to be present on the real host in that case.
+[ -f /var/lib/dpkg/status ] || exit 0
```
lool: I wonder if we could just check for "dpkg" being available as a command instead

lool: Could you:
1) expand if/then
2) log a warning that we're skipping the check

scripts/check-deps.sh
```diff
@@ -0,0 +1,36 @@
...
+for pkg in "$@"; do
+    dpkg -s "$pkg" >/dev/null 2>&1 || missing="$missing $pkg"
+done
+
+[ -z "$missing" ] && exit 0
+
+if [ "$install" = true ]; then
+    sudo=""
+    [ "$(id -u)" -eq 0 ] || sudo="sudo"
+    $sudo apt-get update -qq
```
lool: Please
1) log a message that we're updating APT sources; this helps users and us debugging what's going on, which sources are failing etc.
2) drop -qq

scripts/check-deps.sh
```diff
@@ -0,0 +1,36 @@
...
+    $sudo apt-get install -y --no-install-recommends $missing
```
lool: please log the list of packages being installed

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -4,10 +4,19 @@
...
 sectorsize: {{if eq $imagetype "ufs"}}4096{{else}}512{{end}}
 
 actions:
+  - action: run
+    description: Check host build dependencies
+    chroot: false
+    command: |
+      "${RECIPEDIR}/../scripts/check-deps.sh" \
+          {{if eq $auto_install_deps "true"}}--install{{end}} \
+          dosfstools
```
lool: Hmm I assume that's for debos to craete the EFI partition? why is e2fsprogs not needed then?

.github/workflows/debos.yml
```diff
@@ -73,10 +73,10 @@ jobs:
       - name: Copy U-Boot for RB1 from EFS
         run: cp -av "/efs/u-boot-rb1-latest/rb1-boot.img" .
 
-      # mtools is needed for the flash recipe
+      # dosfstools is needed for the image recipe, mtools is needed for the flash recipe
       # file, device-tree-compiler and u-boot-tools are needed by qcom-dtb-metadata
       - name: Install debos and dependencies of the recipes and local tests
-        run: apt -y install debian-archive-keyring debos file make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml device-tree-compiler u-boot-tools
+        run: apt -y install debian-archive-keyring debos dosfstools file make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml device-tree-compiler u-boot-tools
```
lool: could we drop these and set `-t auto_install_deps:true` during our CI?

lool (changes requested):
Thanks @bjordiscollaku this looks very nice! I made a number of requests for small changes

Makefile
bjordiscollaku: Done in b9ac53d. Moved the flag ahead of `$(DEBOS_OPTS)`: `EXTRA_DEBOS_OPTS` expands at the tail of `DEBOS_OPTS`, and debos keeps the last `-t` value for a given key, so `EXTRA_DEBOS_OPTS='-t auto_install_deps:false'` now wins and a user can opt out. Added a comment noting the container is a disposable instance, so auto-installing into it is safe.

scripts/check-deps.sh
bjordiscollaku: Done in b9ac53d. Expanded to `if [ -z "$missing" ]; then exit 0; fi`.

scripts/check-deps.sh
bjordiscollaku: Done in b9ac53d. The guard is an explicit `if` now and logs a `W:` line to stderr before skipping. I also dropped the word "silently" from the comment, since it now warns.

scripts/check-deps.sh
bjordiscollaku: Done in b9ac53d. Dropped `-qq` and added an `I: refreshing APT package lists` line before the update, so a failing source shows up in the output.

scripts/check-deps.sh
bjordiscollaku: Done in b9ac53d. Logs `I: installing missing host packages:<list>` right before the install.

scripts/check-deps.sh
bjordiscollaku: I dug into this one and I'd keep the file test. It's the more reliable signal here, and it's a fakemachine subtlety worth spelling out.

For the image and flash recipes this runs as a `chroot: false` action under `-b qemu`, i.e. inside the fakemachine guest rather than on the real host. fakemachine bind-mounts the host `/usr` into the guest (plus a couple of `/etc` paths) but not `/var`, so in that environment the `dpkg` binary is present (it lives under the mounted `/usr`) while `/var/lib/dpkg/status` is not. That's exactly the case the guard exists to catch.

If we switched to `command -v dpkg`, the check would no longer skip in the guest; it'd run `dpkg -s` against an empty admin DB, conclude every package is missing, and then either fail image/flash hard (validate path) or try to `apt-get` inside the VM (install path). Testing the status database is what separates a real, queryable host from a fakemachine guest that just happens to have the dpkg binary on `$PATH`.

I've added a comment in the script explaining this. If you'd like a stronger guard I'm happy to gate on `[ -r /var/lib/dpkg/status ]` or on debos's `IN_FAKE_MACHINE` env instead, just not on the binary's presence.

debos-recipes/qualcomm-linux-debian-image.yaml
bjordiscollaku: Yep, that's for the vfat ESP. e2fsprogs is needed by the same `image-partition` step for the ext4 root, but I left it out on purpose: in trixie it's Priority `important` (downgraded from `required` back in #897277), and the default debootstrap set is required+important, so it's reliably present in any normal base image and in the debos container. `dosfstools` is only `optional`, which is why it's the one that actually goes missing in a from-source container (the #371 case) and has to be listed. The other lists follow the same rule: they enumerate the `optional`-priority tools a stripped container can lack.

If you'd prefer the image recipe to be explicit/symmetric I'll add `e2fsprogs` to its set too; it's a no-op on a normal host but covers an aggressively minimised container.

.github/workflows/debos.yml
bjordiscollaku: I'd keep the explicit list for CI, for two reasons.

That apt line installs more than the recipes' host deps: it also brings in `debos` itself, the local test harness (`python3-pexpect`, `python3-pytest`), `qemu-*`, `device-tree-compiler` and `u-boot-tools`. None of those are in any recipe's `check-deps` set, so `auto_install_deps:true` wouldn't pull them and the job would lose debos and the pytest run.

And in CI the image/flash recipes run under `-b qemu`, so their check runs inside the fakemachine guest where `/var/lib/dpkg/status` doesn't exist and it skips, so `auto_install_deps` is effectively inert for those two. Only rootfs runs natively. The host tools image/flash need (mkfs.vfat, mtools, and so on) have to be on the runner before debos starts, which is what this step provides.

Side note while we're here: most of that filesystem tooling is already pulled in transitively. The Debian `debos` package Recommends `dosfstools`/`e2fsprogs`/`parted`/`fdisk` and we install with recommends on, which is why main builds vfat+ext4 green without an explicit `dosfstools`. The line I added is really just belt-and-suspenders against a future `--no-install-recommends`.

If the duplication between the recipe lists and this step is what's bugging you, I'd rather lift the dep lists into one shared spot than drop the CI install. Happy to do that as a follow-up.

debos-recipes/qualcomm-linux-debian-image.yaml
lool: It's logical and best practice to list it as it's an explicit dependency; priority: important packages are not required to be present in a stripped down container. I also like the symmetry with vfat.

.github/workflows/debos.yml
lool: Yes, duplication was what was on my mind; how could we avoid it?

Maybe we shouldn't be running this script from within the recipe, but just before calling debos on the recipe.

debos-recipes/qualcomm-linux-debian-image.yaml
bjordiscollaku: Added in d1df400: image.yaml now lists `dosfstools e2fsprogs`. You're right that priority is the wrong reason to omit it. e2fsprogs is `important`, not `required`, so a stripped-down container is not guaranteed to carry it, and it is a real dependency of the same image-partition step that formats the ext4 root. The symmetry with the vfat/dosfstools case is nicer too.

.github/workflows/debos.yml
bjordiscollaku: Agreed. We could move it out of the recipe and run it just before the debos call, say from the Makefile, and keep each recipe's deps as a variable there. That way they're not repeated between the recipes and the CI apt step. What do you think?

lool: Yup; I think this would look clean. I wonder whether the script should have the list of use cases embedded in it, so that makefile / CI / README basically leverage `check-deps.sh host` , `check-deps.sh image`, `check-deps.sh flash` etc.

bjordiscollaku: Yep agreed, and the use cases would be:

- host: debos, make, qemu-system-arm
- rootfs: mmdebstrap, debian-archive-keyring
- image: dosfstools, e2fsprogs
- flash: dosfstools, mtools, xmlstarlet, python3-defusedxml, unzip, device-tree-compiler, u-boot-tools, file
- test: qemu-efi-aarch64, qemu-system-arm, python3-pexpect, python3-pytest

Makefile, CI, and README each call `check-deps.sh <use-case>`. Does that grouping look right?

lool: I haven't cross-checked the lists of packages, but +1 on the target names

lool:
@bjordiscollaku BTW I just remembered, we also have a couple more implementations of similar logic in the same repo:
- scripts/build-linux-deb.py
- scripts/build-u-boot-rb1.sh

but we can address these later or even let them be. Slight preference for a Python implementation eventually, but also happy with the current shell one to start with, it's simple enough.

obbardc:
My comment would be to not include qemu in the list, since it's a dependency on the Debian package of debos

bjordiscollaku:
> @bjordiscollaku BTW I just remembered, we also have a couple more implementations of similar logic in the same repo:
> 
> * scripts/build-linux-deb.py
> * scripts/build-u-boot-rb1.sh
> 
> but we can address these later or even let them be. Slight preference for a Python implementation eventually, but also happy with the current shell one to start with, it's simple enough.

Sounds good, I'll leave those two for a follow-up.

bjordiscollaku:
> My comment would be to not include qemu in the list, since it's a dependency on the Debian package of debos

Agreed, dropping `qemu-system-arm`. I'd keep `qemu-efi-aarch64` though, it's the aarch64 UEFI firmware the boot test needs and debos doesn't pull it (only Recommends: `ovmf`).

obbardc:
> > My comment would be to not include qemu in the list, since it's a dependency on the Debian package of debos
> 
> Agreed, dropping `qemu-system-arm`. I'd keep `qemu-efi-aarch64` though, it's the aarch64 UEFI firmware the boot test needs and debos doesn't pull it (only Recommends: `ovmf`).

Yeah, that sounds good to me. Thanks!

scripts/check-deps.sh
```diff
@@ -0,0 +1,47 @@
...
+# skip: the dependencies are expected to be present on the real host that
+# invoked debos.
+if [ ! -f /var/lib/dpkg/status ]; then
+    echo "W: no dpkg database found; skipping host dependency check" >&2
+    exit 0
+fi
+
+missing=""
+for pkg in "$@"; do
+    dpkg -s "$pkg" >/dev/null 2>&1 || missing="$missing $pkg"
```
basak-qcom: This is not a robust way of determining if a package is installed. Packages can be in a removed (but not purged) or broken state and would still appear under `dpkg -s`. Instead you could use `apt-get install -s` to see what apt-get would do, to see if any package installations are required.

I would call `sudo apt-get install -y --no-install-recommends "$@"` in the `--install` case, and `apt-get install -qqs --no-install-recommends "$@"|grep -E '^(Inst|Conf)'` in the non- `--install` case. Wouldn't that be much cleaner, and avoid having to second guess apt with dpkg?

It would have the effect of failing if an `apt upgrade` is pending, but generally that seems like a good idea anyway, and I prefer that over the fragility of calling dpkg directly. If you want to avoid that, then you could call `dpkg-query -f '${Status}' -W "$@"` and check for `ok` and `installed` in the second and third columns or similar.

basak-qcom (changes requested):

scripts/check-deps.sh
bjordiscollaku: Thanks for the feedback Robie.

`--install`: I'll drop the pre-check and just call `apt-get install -y --no-install-recommends` on the set, since it's idempotent.

Validate-only: you preferred the simulate, but I'd take the `dpkg-query` option you offered, because of one hole I checked in the simulate. On a name apt can't resolve (a typo, or a dropped/renamed package), `apt-get install -qqs <name>` prints "Unable to locate package" to stderr and nothing to stdout, so `grep -E '^(Inst|Conf)'` matches nothing and the check reads it as satisfied. `dpkg-query -f '${Status}' -W` reports it not-installed there, so the gate fails closed instead.

PR #457, opened 2026-06-23
feat(debos/rootfs)!: add QLI APT repository

lool:
Add the main (Debusine backed) QLI APT repository.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -222,6 +222,32 @@ actions:
       EOF
 {{- end }}
 
+# Qualcomm Linux maintains overlays for trixie/forky
+{{- if ne $suite "unstable" }}
+  - action: run
+    description: Create Qualcomm Linux APT sources for {{ $suite }}
+    chroot: true
+    command: |
+      set -eux
+      tee /etc/apt/sources.list.d/qli.sources <<EOF
+      # Qualcomm Linux repository
+      Types: deb
+      URIs: https://deb.debusine.qualcomm.com/qualcomm/qli
+      Suites: {{ $suite }}
+      Components: main contrib non-free-firmware non-free
+      Signed-By:
+       -----BEGIN PGP PUBLIC KEY BLOCK-----
+       .
+       mDMEag8p/xYJKwYBBAHaRw8BAQdAdB6JSNF1OXxnsTgp4VTUekW52BM7e6ZQVRsq
+       QT5QDaS0JEFyY2hpdmUgc2lnbmluZyBrZXkgZm9yIHF1YWxjb21tL3FsaYiQBBMW
+       CgA4FiEEOwuFfyf8aPE5SQakb8qSvoHfw8IFAmoPKf8CGwMFCwkIBwIGFQoJCAsC
+       BBYCAwECHgECF4AACgkQb8qSvoHfw8Lz1gEA9XocADbvqUgZQc0LceThn7vMI98d
+       kTJoiInuulQ6rEUBANo+GOKILH71VRnZ5jWtsu7IlVk7oUMlTtC0eE5tcBwB
+       =bX6V
+       -----END PGP PUBLIC KEY BLOCK-----
+      EOF
+{{- end }}
+
   - action: run
     description: Remove old APT sources and upgrade packages
     chroot: true


lool:
Couple of implementation notes:
- didn't bother with snapshots yet
- enabled unconditionally

lool:
NB: GPG key is subject to change in a few months

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -222,6 +222,31 @@ actions:
       EOF
 {{- end }}
 
+{{- if ne $suite "unstable" }}
```
obbardc, line 226: why not building for unstable ? just wondering :-)

lool: This repo is our (new) overlay on top of Debian suites; we have no unstable overlay, only a forky one; packages are short-lived in unstable, so it would be hard to maintain an overlay on top of that, albeit technically possible :)

obbardc: Makes sense now you explain, maybe it's worth a comment in the recipe explaining?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -222,6 +222,31 @@ actions:
...
+    description: Create Qualcomm Linux APT sources for {{ $suite }}
+    chroot: true
+    command: |
+      set -eux
+      tee /etc/apt/sources.list.d/qli.sources <<EOF
+      # Qualcomm Linux repository
+      Types: deb
+      URIs: https://deb.debusine.qualcomm.com/qualcomm/qli
+      Suites: {{ $suite }}
+      Components: main
```
basak-qcom: I think we might need non-free-firmware (hexagon-dsp-binaries) and contrib (fastrpc). apt prints warnings when these components aren't published by the repository, which is the case currently, but I think they'll appear when published to and otherwise these packages will be invisible to us.

I'm not entirely sure how Debusine maps components but I assume it defaults to taking it from the control file. We don't have a live example right now as we're not publishing hexagon-dsp-binaries in our Debusine overlay yet, and fastrpc hasn't caught up on the move to contrib.

We could also only use "main" in our overlay, but then that's a delta to maintain against Debian, so it'd be easier not to I think.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
basak-qcom: Should we perhaps map unstable to forky for the purposes of this block, rather than ignore it? I think that might be less surprising to someone building qcom-deb-images "for unstable".

gagath: People building for unstable (me) are already outside of the supported path and are on their own. I do not think that mapping forky to unstable makes more sense for now. However I agree that there should be a clear comment explaining why we do not support unstable for the QLI overlay, which this PR seems to be currently missing.

basak-qcom: > People building for unstable (me) are already outside of the supported path and are on their own.

I think this is fine, but then I think:

1) We should not leave mines checking for people doing this to step on, ignoring the unstable use case completely; and
2) We should explicitly fail (but just in one place) if we detect unstable, ready for someone working on an unstable build to bypass by themselves.

Then someone trying to build "unsupported" for unstable could expect either forky behaviour or failure, rather than a surprising half attempt to make it work that then doesn't.

Does that make sense?

Concretely that means removing this `if` entirely and maybe adding a wider "fail if unstable" assertion somewhere.

lool: Agathe cared to build for unstable; I will add a comment and leave the overall question of unstable to a group discussion/separate issue or PR. I agree we could have clearer docs + enforce clear checks/behavior on supported combos.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: Good point; we will have contrib (fastrpc), non-free-firmware (hexagon-dsp-binaries, fastcv-binaries), and non-free (sensorhub, adreno) so I'll list them all. Would be nice if Debusine coudl ALWAYS publish the components even if the list is empty.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: Added

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: Added

gagath (approved):
LGTM

PR #459, opened 2026-06-28
fix(debos/image): disable debos-grow-rootfs service after first successful run

abhi5mand:
debos-grow-rootfs.service runs on every boot even after the partition
and filesystem were already grown. Add ExecStartPost to disable the 
service after a successful first run.


lool (approved):
Looks sensible; thanks!

PR #461, opened 2026-07-01
qualcomm-linux-debian-flash: update boot to 00073

jayantsa1:
Known fixes:
Glymur

S2RAM is auto resuming
Improved UEFI compliance, USB/Type‑C interoperability Improved PCIe stability
Enabled FastCV capabilities and resolved stability and compilation


lool (approved):
Approved, note that we don't have glymur-crd enabled for our PR CI yet, so we will only see test results tomorrow.

PR #462, opened 2026-07-02
fix: use /usr/bin instead of /bin for Debian

gagath:
Now that usrmerge is done in Debian, we know that /bin is a symlink to /usr/bin. Directly use /usr/bin where it makes sense to avoid an extra symlink dereference.

Note that this is not done for the scripts/ directory because we cannot assume that the user of these scripts is running an usrmerged system. Same for the recipe scripts not run in the Debian chroot. We only do this for everything that we are sure is running on a Debian target.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-image.yaml
+++ b/debos-recipes/qualcomm-linux-debian-image.yaml
@@ -106,7 +106,7 @@ actions:
       set -eux
       # Fixes error when running outside qemu
       dpkg-divert --local --rename --add /etc/kernel/postinst.d/u-boot-efi-dtb
-      ln -s /bin/true /etc/kernel/postinst.d/u-boot-efi-dtb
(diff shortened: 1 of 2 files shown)


lool:
@gagath will this work on trixie as well? or should we defer this post-forky release?

gagath:
Will work on trixie:

```console
root@trixie-nano:~# cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.5
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
root@trixie-nano:~# ls -l /bin
lrwxrwxrwx 1 root root 7 May  8 16:10 /bin -> usr/bin
```

lool (approved):

PR #463, opened 2026-07-07
fix(debos/flash): update qcom-dtb-metadata to pick up Glymur SOC ID fix

abhi5mand:
Update qcom-dtb-metadata to revision b07b6e7792b8c1a24d9758353fdefbe9591eb5a3

This pulls in qualcomm-linux/qcom-dtb-metadata#109, which changes the Glymur
SOC ID mapping from SIP to COB, aligning it with the supported hardware
variant.


lool (approved):
I reviewed the git log and diff between old and new rev, and saw only safe additions/tweaks.

PR #464, opened 2026-07-07
chore(lava/rb3gen2): use new device_type

lool:
Device Types were renamed after a LAVA upgrade; use the new
qcs6490-rb3gen2-vision-kit device_type name, matching the CI
template filename and the upstream kernel DTB name.


lool:
self-merging since @mwasilew did the same change in #465

PR #465, opened 2026-07-07
ci: fix LAVA templates to match 2026.05-qualcomm release

mwasilew:
The CompressOverlay fix causes the overlay to always be packaged from "/". Prior to this release the top directory in the overlay tarball was named "lava-XXX" where XXX is the job ID. LAVA stripped the rest of the path during compression. This means that the jobs had to uncompress the tarball to a proper location if lava_test_results_dir variable was used. After the fix tarball can be decompressed at "/" and overlay will be placed in the right directory.


mwasilew:
@lool this PR assumes that the rootfs is mounted rw. If this is not the case, please let me know where should the overlay be written to?

mattface (approved):

lool (changes requested):
The two changes call for 2 commits, but I already did one of these two in #464 which I've now also merged; can you rebase and then I'll merge yours?

lool (approved):
Looks good; thanks

PR #466, opened 2026-07-08
Ci/qdl lava templates

mwasilew:
ci: move LAVA templates to native qdl deployment

Use deploy/boot to qdl from LAVA actions. Support for QDL is available since 2026.05-qualcomm and 2026.06 (upstream).

### DIFF
--- a/ci/lava/glymur-crd/boot.yaml
+++ b/ci/lava/glymur-crd/boot.yaml
@@ -1,37 +1,35 @@
 actions:
 - deploy:
-    images:
-      image:
-        headers:
-          Authorization: Q_S3_TOKEN
-        url: "{{BUILD_DOWNLOAD_URL}}/{{SUITE}}-flash-emmc.tar.gz"
-    postprocess:
-      docker:
-        image: ghcr.io/foundriesio/lava-lmp-sign:main
-        steps:
-        - export IMAGE_PATH=$PWD
-        - cp overlay*.tar.gz overlay.tar.gz
-        - echo "OVERLAY=overlay.tar.gz" >> $IMAGE_PATH/flash.settings
-        - echo "OVERLAY_PATH=/" >> $IMAGE_PATH/flash.settings
-        - echo "ROOTFS_IMAGE=disk-sdcard.img2" >> $IMAGE_PATH/flash.settings
-        - echo "DEVICE_TYPE=debian-glymur-crd_nvme" >> $IMAGE_PATH/flash.settings
-        - echo "STORAGE=nvme" >> $IMAGE_PATH/flash.settings
-        - echo "FIREHOSE_PROGRAM=xbl_s_devprg_ns.melf" >> $IMAGE_PATH/flash.settings
-        - cat $IMAGE_PATH/flash.settings
-    timeout:
-      minutes: 200
-    to: downloads
-- deploy:
-    images:
-      image:
-        url: downloads://{{SUITE}}-flash-emmc.tar.gz
-      settings:
-        url: downloads://flash.settings
-      overlay:
-        url: downloads://overlay.tar.gz
+    rootfs_image: disk-sdcard.img2
+    overlay_path: /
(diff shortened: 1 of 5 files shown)


mwasilew:
Marking as draft because RB1 flashing times out. I'm working with @igoropaniuk to fix it in our LAVA instance.

mwasilew:
glymur: https://lava.infra.foundries.io/scheduler/job/309705
rb3gen2: https://lava.infra.foundries.io/scheduler/job/309706
rb1: https://lava.infra.foundries.io/scheduler/job/309703 (fails due to timeout in hash calculations)
rb4: https://lava.infra.foundries.io/scheduler/job/309721(failed due to UFS issue?)

mwasilew:
With new qdl flashing RB1 works: https://lava.infra.foundries.io/scheduler/job/310259

ci/lava/glymur-crd/boot.yaml
```diff
@@ -1,37 +1,35 @@
...
-    images:
-      image:
-        url: downloads://{{SUITE}}-flash-emmc.tar.gz
-      settings:
-        url: downloads://flash.settings
-      overlay:
-        url: downloads://overlay.tar.gz
+    rootfs_image: disk-sdcard.img2
+    overlay_path: /
+    qcomflash:
```
lool, line 5: I don't think it's something for this PR, but I found the name "qcomflash" for this a bit surprising

ci/lava/monaco-evk/boot.yaml
```diff
@@ -1,35 +1,36 @@
...
+    patch: patch0.xml
+    path: flash_monaco-evk_ufs/sail_nor
+    storage: spinor
+    debug: true
+    timeout:
+      minutes: 15
+    method: qdl
+- boot:
+    firehose_program: prog_firehose_ddr.elf
+    rawprogram: rawprogram0.xml rawprogram1.xml rawprogram2.xml rawprogram3.xml
```
lool, line 25: This is not really feedback on this PR, but rather a note on the upstream implementation

I like the new method-specific yaml syntax, but I miss the globing, and I feel this is putting us further away from common templates that we can share across qualcomm devices; there are other ways to address this though

ci/lava/monaco-evk/boot.yaml
```diff
@@ -1,35 +1,36 @@
...
+- boot:
+    firehose_program: prog_firehose_ddr.elf
+    rawprogram: rawprogram0.xml
+    patch: patch0.xml
+    path: flash_monaco-evk_ufs/sail_nor
+    storage: spinor
+    debug: true
+    timeout:
+      minutes: 15
+    method: qdl
```
lool, line 22: I have the impression method defines which fields will be expected/used in that yaml subtree, so it should probably come first

lool (approved):
I've reviewed the changes, they look good to me; I've left a few long-term improvement notes if you're interested; thanks for running and linking to test jobs

ci/lava/glymur-crd/boot.yaml
mwasilew: I was never good at naming things. If you would like to change it, it needs to go here: https://gitlab.com/lava/lava/-/blob/master/lava_dispatcher/actions/deploy/qdl.py?ref_type=heads#L37

ci/lava/monaco-evk/boot.yaml
mwasilew: https://github.com/qualcomm-linux/lava-test-plans/commit/8b7701a414ad23330732385b6f2c37d09f4539b5
Maybe this fixes your issue? This would be a part of #405

ci/lava/monaco-evk/boot.yaml
mwasilew: Key order isn't significant, but I agree it would make more sense to know what `method` and `to` are up front (reading top to bottom)

PR #468, opened 2026-07-08
install kernel headers

### LINKED ISSUE #467: linux-headers missing on default rootfs is blocker for DKMS standalone package installation
default  qcom-deb-image rootfs installign kernelpackage derived from https://github.com/qualcomm-linux/qcom-deb-images/blob/9526d3aedf3dff443cee20fc5d1cf78b44d5f03f/.github/workflows/debos.yml#L112

But standalone/Traditional  DKMS installation needs linux-headers package also needs to be installed, DKMS installation is blocker for REL.1
As we don't have kernel deb packages are hosting onto DEBUSINE APT yet to install linux-headers.

can we have linux-headers debian package also being installed on default rootfs ? till we have  kernel debs hosted onto DEBUSINE  or approach of DKMS debian package being packed into kernel implementation mainlined,

lool:
Refactor `kernel_package` / `kernelpackage` semantics to take a list of
packages, and use this to install headers:
- **refactor(kernel)!: accept a list of kernel packages**
- **feat(kernel)!: install linux-headers along -image**

Headers are not installed when using standard Debian kernels, as
they can be installed from APT.

Fixes: #467

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -107,9 +107,17 @@ jobs:
           kernel_packages="$KERNEL_PACKAGES"
           suite="$SUITE"
           case "$kernel_packages" in
-            # if kernel package is from EFS, determine actual package name(s)
+            # NB: this is slightly weak as we collapse
+            # kernelpackages which is potentially a list and only
+            # look at the special value efs/* instead of splitting
+            #
+            # if kernel package is from EFS, determine actual package name(s);
+            # install the generated linux-headers-* packages alongside the
+            # linux-image-* ones since these are ephemeral and not published in
+            # an APT repo (unlike official Debian kernels, installed from APT)
+            # 
             efs/*)
-              kernel_packages="$(find local-apt-repo/kernel -type f -name 'linux-image-*' -not -name '*dbg*'|xargs -n1 basename|cut -f1 -d_|paste -sd,)"
+              kernel_packages="$(find local-apt-repo/kernel -type f \( -name 'linux-image-*' -o -name 'linux-headers-*' \) -not -name '*dbg*'|xargs -n1 basename|cut -f1 -d_|sort|paste -sd,)"
               ;;
             default-ci)
               case "$suite" in
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -11,11 +11,11 @@ on:
         description: List of overlays to use
         type: string
         default: qsc-deb-releases
-      kernelpackage:
+      kernelpackages:
         description: >
-          Kernel package to install from APT repositories (e.g.
-          linux-image-arm64) or name of a EFS kernel package directory
-          (e.g. efs/mainline)
+          Comma-separated list of kernel packages to install from APT
+          repositories (e.g. linux-image-arm64) or name of a EFS kernel
+          package directory (e.g. efs/mainline)
         type: string
         default: default-ci
       debos_extra_args:
@@ -40,7 +40,7 @@ permissions:
 env:
   # image build id; used for SBOM generation
   BUILD_ID: ${{ github.run_id }}-${{ github.run_attempt }}
-  KERNEL_PACKAGE: ${{ inputs.kernelpackage }}
+  KERNEL_PACKAGES: ${{ inputs.kernelpackages }}
   SUITE: ${{ inputs.suite }}
   DEBOS_EXTRA_ARGS: -t suite:${{ inputs.suite }} ${{ inputs.debos_extra_args }}
   PREFIX: ${{ inputs.suite }}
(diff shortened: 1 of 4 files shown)


debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -2,7 +2,8 @@
 {{- $gnomedesktop := or .gnomedesktop "false" }}
 {{- $localdebs := or .localdebs "none" }}
 {{- $aptlocalrepo := or .aptlocalrepo "none" }}
-{{- $kernelpackage := or .kernelpackage "linux-image-arm64" }}
+# $kernelpackage (singular) is deprecated, superseded by $kernelpackages
```
gagath, line 5: Can we add a print here when the old `${kernelpackage}` argument is used so that users see they are using a deprecated option and have a chance to migrate to the new option?

.github/workflows/debos.yml
```diff
@@ -104,23 +104,26 @@ jobs:
...
-          case "$kernel_package" in
-            # if kernel package is from EFS, determine actual package name
+          case "$kernel_packages" in
+            # if kernel package is from EFS, determine actual package name(s);
+            # install the generated linux-headers-* packages alongside the
+            # linux-image-* ones since these are ephemeral and not published in
+            # an APT repo (unlike official Debian kernels, installed from APT)
             efs/*)
-              kernel_package="$(find local-apt-repo/kernel -type f -name 'linux-image-*' -not -name '*dbg*'|xargs -n1 basename|cut -f1 -d_)"
+              kernel_packages="$(find local-apt-repo/kernel -type f \( -name 'linux-image-*' -o -name 'linux-headers-*' \) -not -name '*dbg*'|xargs -n1 basename|cut -f1 -d_|sort -u|paste -sd,)"
```
gagath: I am not sure I understand the use of `sort -u`. File names should already be unique? And sorting has no clear value either?

.github/workflows/debos.yml
```diff
@@ -104,23 +104,26 @@ jobs:
...
-          kernel_package="$KERNEL_PACKAGE"
+          kernel_packages="$KERNEL_PACKAGES"
           suite="$SUITE"
-          case "$kernel_package" in
-            # if kernel package is from EFS, determine actual package name
+          case "$kernel_packages" in
+            # if kernel package is from EFS, determine actual package name(s);
+            # install the generated linux-headers-* packages alongside the
+            # linux-image-* ones since these are ephemeral and not published in
+            # an APT repo (unlike official Debian kernels, installed from APT)
```
gagath, line 117: This has the caveat of checking only the first package name. It works but should be documented as a fragile implementation. If I pass `kernelpackages=efs/mykernel,strace` then everything will probably explode.

gagath (changes requested):

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: Good idea; I've pushed a new commit adding a warning and pausing briefly to give people a chance to read it

.github/workflows/debos.yml
lool: Just stability of that output (ls not sorting in a deterministic order), happy to drop it if you prefer

.github/workflows/debos.yml
lool: Yeah, do you think we should have docs for the CI variables, or what kind of fix would we deliver?

Right now, we have user set kernelpackages (list of packages to install), or we have two special values:
- default-ci
- efs/<kernel flavor>
which are just for our own CI.

gagath: Actionable item is to just document that this check is a bit weak for the `efs/*` case as it will only check for the first item of the list. I do not expect a code fix but a comment saying this is intentional.

.github/workflows/debos.yml
gagath: Then keep the sort and remove the `-u`. If we have multiple entries with the same name we have a bigger problem. Thanks.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
gagath: LGTM, the sleep is a good idea. Maybe say the build will continue after 10s so users do not ^C out of despair.

.github/workflows/debos.yml
lool: updated comment with this caveat

.github/workflows/debos.yml
lool: dropped -u

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: added another message stating we're sleeping for 10s

gagath (approved):

PR #120, opened 2025-08-06
Added qcom-preflight-workflow

njjetha:
## Summary
qcom-preflight-checks calls a [reusable-workflow](https://github.com/qualcomm/qcom-reusable-workflows/blob/main/.github/workflows/qcom-preflight-checks-reusable-workflow.yml) that runs a series of preflight checks on your proposed contribution.

For more info check https://github.com/qualcomm/qcom-actions/blob/main/README.md

PR #470, opened 2026-07-13
fix(debos/flash): update qcom-dtb-metadata commit to latest and filter FIT image via --soc

neervish:
## Summary
Bump the qcom-dtb-metadata pin to the latest commit (d30e9cdb766f08b05160ededf53556ed665d0c0b) and pass `--soc` to
`build-dtb-image.sh` to create the fitimage for the targets supported in this repo.

This pulls [PR #119] (https://github.com/qualcomm-linux/qcom-dtb-metadata/pull/119) to support glymur-crd-camx.dtbo
Taking above PR #119 makes dtb-multidtb.bin size grow beyond 4MB as many new SoC/board support have been added on qcom-dtb-metadata.

To fix that, this also takes [PR #120] (https://github.com/qualcomm-linux/qcom-dtb-metadata/pull/120) which adds --soc/--board filtering to build-dtb-image.sh, and pass:

      --soc {{ $fit_soc | uniq | join " " }}

restricting the FIT image to only the silicon families this repo builds boards for.
This brings dtb-multidtb.bin back under 4MB.
fit_soc expands into the silicon families this repo supports.

This also takes [PR #122] (https://github.com/qualcomm-linux/qcom-dtb-metadata/pull/122) which fixes a whitespace/indentation bug in the fdt-kodiak-el2.dtbo block that broke --prune's tab-anchored regex and caused build failures.

## Result
dtb-multidtb.bin back to 4MB with glymur camx dtbo support.

## Exception
qcm2290 boards are excluded since qcom-dtb-metadata has no entry for qcm2290 under /soc; passing it to --soc fails with "[ERROR] Invalid --soc 'qcm2290'".
Raised a BUG here https://github.com/qualcomm-linux/qcom-dtb-metadata/issues/126 to add support for qcm2290 soc.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -29,10 +29,15 @@ actions:
     sha256sum: a81f39859a6dbe00cda2bc995890186f8733244ed6885d061573aaea1b39e0f4
     unpack: true
 
+# soc_id identifies the SoC/SKU each board is built on, using the same
+# names as qcom-dtb-metadata's /soc node (qcom-metadata.dts), so --soc
+# can select the FIT image configs for the boards this repo supports.
+# Values are derived from IPCAT and verified against the kernel's
+# include/dt-bindings/arm/qcom,ids.h QCOM_ID_* constants.
 {{- $boards := list }}
 {{- $boards = append $boards (dict
     "name" "glymur-crd"
-    "silicon_family" "glymur"
+    "soc_id" "glymur"
     "ptool_platforms" (list "glymur-crd/nvme" "glymur-crd/spinor")
     "boot_binaries_download" (dict
         "description" "Glymur boot binaries"
@@ -55,7 +60,7 @@ actions:
 {{- if eq $build_qcs615 "true" }}
 {{- $boards = append $boards (dict
     "name" "qcs615-ride"
-    "silicon_family" "qcs615"
+    "soc_id" "qcs615"
     "ptool_platforms" (list "qcs615-ride/ufs")
     "boot_binaries_download" (dict
         "description" "QCS615 boot binaries"
@@ -78,7 +83,7 @@ actions:
 {{- if eq $build_qcm6490 "true" }}
 {{- $boards = append $boards (dict
     "name" "qcs6490-rb3gen2-vision-kit"
-    "silicon_family" "qcm6490"
+    "soc_id" "qcs6490"
     "ptool_platforms" (list "qcs6490-rb3gen2/ufs")
     "boot_binaries_download" (dict
         "description" "QCM6490 boot binaries"
@@ -101,7 +106,7 @@ actions:
 {{- if eq $build_qcs8300 "true" }}
 {{- $boards = append $boards (dict
     "name" "qcs8300-ride"
-    "silicon_family" "qcs8300"
+    "soc_id" "qcs8300"
     "ptool_platforms" (list "qcs8300-ride-sx/ufs")
     "boot_binaries_download" (dict
         "description" "QCS8300 boot binaries"
@@ -122,7 +127,7 @@ actions:
 )}}
 {{- $boards = append $boards (dict
     "name" "monaco-evk"
-    "silicon_family" "qcs8300"
+    "soc_id" "qcs8275"
     "ptool_platforms" (list "iq-8275-evk/emmc" "iq-8275-evk/ufs")
     "boot_binaries_download" (dict
         "description" "QCS8300 boot binaries"
@@ -143,7 +148,7 @@ actions:
 )}}
 {{- $boards = append $boards (dict
     "name" "monaco-arduino-monza"
-    "silicon_family" "qcs8300"
+    "soc_id" "qcs8300"
     "ptool_platforms" (list "qcs8275-monza/emmc")
     "boot_binaries_download" (dict
         "description" "QCS8300 boot binaries"
@@ -166,7 +171,7 @@ actions:
 {{- if eq $build_qcs9100 "true" }}
 {{- $boards = append $boards (dict
     "name" "qcs9100-ride-r3"
-    "silicon_family" "qcs9100"
+    "soc_id" "qcs9100"
     "ptool_platforms" (list "qcs9100-ride-sx/ufs")
     "boot_binaries_download" (dict
         "description" "QCS9100 boot binaries"
@@ -187,7 +192,7 @@ actions:
 )}}
 {{- $boards = append $boards (dict
     "name" "lemans-evk"
-    "silicon_family" "qcs9075"
+    "soc_id" "qcs9075"
     "ptool_platforms" (list "iq-9075-evk/ufs")
     "boot_binaries_download" (dict
         "description" "QCS9100 boot binaries"
@@ -210,7 +215,7 @@ actions:
 {{- if eq $build_rb1 "true" }}
 {{- $boards = append $boards (dict
     "name" "qrb2210-rb1"
-    "silicon_family" "qcm2290"
+    "soc_id" "qcm2290"
     "ptool_platforms" (list "qrb2210-rb1/emmc")
     "boot_binaries_download" (dict
         "description" "RB1 rescue image"
@@ -225,7 +230,7 @@ actions:
 {{- end }}
 {{- $boards = append $boards (dict
     "name" "qrb2210-arduino-imola"
-    "silicon_family" "qcm2290"
+    "soc_id" "qcm2290"
     "ptool_platforms" (list "qrb2210-unoq/emmc-16GB")
     "boot_binaries_download" (dict
         "description" "UNO Q boot binaries"
@@ -322,7 +327,7 @@ actions:
 
 {{- range $board := $boards }}
       ### board: {{ $board.name }}
-      ### silicon family: {{ $board.silicon_family }}
+      ### soc id: {{ $board.soc_id }}
 
       # mirror template variables in shell variables to keep the action body
       # in shell syntax
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -23,10 +23,10 @@ actions:
 
   - action: download
     description: Download qcom-dtb-metadata (DTB image build scripts and ITS/metadata files)
-    url: https://github.com/qualcomm-linux/qcom-dtb-metadata/archive/b07b6e7792b8c1a24d9758353fdefbe9591eb5a3.tar.gz
+    url: https://github.com/qualcomm-linux/qcom-dtb-metadata/archive/d30e9cdb766f08b05160ededf53556ed665d0c0b.tar.gz
     name: qcom-dtb-metadata
     filename: qcom-dtb-metadata.tar.gz
-    sha256sum: a81f39859a6dbe00cda2bc995890186f8733244ed6885d061573aaea1b39e0f4
+    sha256sum: a43ccf98149658352a74c54cac1455c35b44e7d48798dc0072540604ec475ae9
     unpack: true
 
 # soc_id identifies the SoC/SKU each board is built on, using the same
@@ -312,6 +312,16 @@ actions:
           ;;
       esac
 
+      # Derive the list of SoCs supported.
+      # qcm2290 is excluded here because qcom-dtb-metadata's /soc node has no
+      # entry for it yet: https://github.com/qualcomm-linux/qcom-dtb-metadata/issues/126
+      {{- $fit_soc := list }}
+      {{- range $board := $boards }}
+      {{- if ne $board.soc_id "qcm2290" }}
+      {{- $fit_soc = append $fit_soc $board.soc_id }}
+      {{- end }}
+      {{- end }}
+
       # Generate dtb.bin (multi-DTB FIT image) from the qcom/ subtree of dtbs.tar.gz.
       if [ -f "${ARTIFACTDIR}/dtbs.tar.gz" ]; then
           QCOM_DTB_METADATA="$(ls -d "${ROOTDIR}/../qcom-dtb-metadata.tar.gz.d/qcom-dtb-metadata-"*)"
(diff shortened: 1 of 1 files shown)


lool (commented):
@neervish The change looks clean, but I don't like that there is a secret place where the list of supported SoCs/families is hardcoded.

Can we derive this from the list of boards we're building for instead? The silicon family should be in the board "database" already

neervish:
> Neeraj (@neervish) The change looks clean, but I don't like that there is a secret place where the list of supported SoCs/families is hardcoded.
> 
> Can we derive this from the list of boards we're building for instead? The silicon family should be in the board "database" already

@lool  Done, updated this in a new commit.

  Used fit_soc_aliases for two boards where silicon_family doesn't match the FIT's compatible string in qcom-fitimage.its:
    **-** qcs6490-rb3gen2-vision-kit: silicon_family "**qcm6490**", but FIT uses compatible = "qcom,**qcs6490**-iot" -> alias "qcs6490"
    **-** monaco-evk: silicon_family "**qcs8300**", but FIT uses compatible = "qcom,**qcs8275**-iot" -> alias "qcs8275"

  For other boards this isn't needed — e.g. qcs9100-ride-r3's
  silicon_family "**qcs9100**" already matches compatible = "qcom,**qcs9100**-qam"
  directly.

  qcm2290 boards are excluded since qcom-dtb-metadata has no entry for
  qcm2290 under /soc; passing it to --soc fails with
  "[ERROR] Invalid --soc 'qcm2290'".

  $fit_soc is now derived from $boards on every build, so newly added boards are automatically covered via their silicon_family — no more manual --soc list edits needed.

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -307,6 +309,17 @@ actions:
           ;;
       esac
 
+      # Derive the list of SoCs supported
+      {{- $fit_soc := list }}
+      {{- range $board := $boards }}
+      {{- if ne $board.silicon_family "qcm2290" }}
```
lool: why single out agatti?

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -79,6 +79,7 @@ actions:
 {{- $boards = append $boards (dict
     "name" "qcs6490-rb3gen2-vision-kit"
     "silicon_family" "qcm6490"
+    "fit_soc_aliases" (list "qcs6490")
```
lool: can we take a step back and avoid managing two sets of silicon family names here?

are there other projects than qcom-dtb-metadata that use the same list of silicon family names as in the --soc parameter? I'd like to anchor on some stable common definition rather than have 5 definitions of the silicon family...

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -123,6 +124,7 @@ actions:
 {{- $boards = append $boards (dict
     "name" "monaco-evk"
     "silicon_family" "qcs8300"
+    "fit_soc_aliases" (list "qcs8275")
```
lool: ditto, very unfortunate naming :-(

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -315,7 +328,7 @@ actions:
           tar -C build/dtbs -xzf "${ARTIFACTDIR}/dtbs.tar.gz" qcom
           "${QCOM_DTB_METADATA}/build-dtb-image.sh" \
               --dtb-src build/dtbs/qcom \
-              --soc glymur qcm6490 qcs6490 qcs8300 qcs8275 qcs9075 qcs9100 qcs615 \
+              --soc {{ $fit_soc | uniq | join " " }} \
```
lool, line 333: This looks better, thanks!

lool:
Sorry, missed your notes below; reply below too:

> Used fit_soc_aliases for two boards where silicon_family doesn't match the FIT's compatible string in qcom-fitimage.its: **-** qcs6490-rb3gen2-vision-kit: silicon_family "**qcm6490**", but FIT uses compatible = "qcom,**qcs6490**-iot" -> alias "qcs6490" **-** monaco-evk: silicon_family "**qcs8300**", but FIT uses compatible = "qcom,**qcs8275**-iot" -> alias "qcs8275"
> 
> For other boards this isn't needed — e.g. qcs9100-ride-r3's silicon_family "**qcs9100**" already matches compatible = "qcom,**qcs9100**-qam" directly.

I'd like to avoid having parallel naming schemes. For DTB names, I followed the linux kernel names. Is there a solid anchor we can follow for the silicon family names that both qcom-deb-images and qcom-dtb-metadata could follow?

> qcm2290 boards are excluded since qcom-dtb-metadata has no entry for qcm2290 under /soc; passing it to --soc fails with "[ERROR] Invalid --soc 'qcm2290'".

Could you please log that bug in qcom-dtb-metadata ("Please add support for Agatti/qcm2290" I guess) and add that link as a comment above the exception in your changes.

> $fit_soc is now derived from $boards on every build, so newly added boards are automatically covered via their silicon_family — no more manual --soc list edits needed.

Thanks, that's in the spirit of what I requested; I want to have a clear shared definition of the silicon family names now, but that's good progress!

debos-recipes/qualcomm-linux-debian-flash.yaml
neervish: > can we take a step back and avoid managing two sets of silicon family names here?
> 
> are there other projects than qcom-dtb-metadata that use the same list of silicon family names as in the --soc parameter? I'd like to anchor on some stable common definition rather than have 5 definitions of the silicon family...

@lool 
Checked meta-qcom , it does not uses qcom-dtb-metadata at all.
Checked qcom-distro-images , it does uses qcom-dtb-metadata's --soc filtering but it has the same "hardcoded list" issue which you had flagged.

So I think we can follow qcom-dtb-metadata's own /soc names.
As silicon_family is not used anywhere in qcom-deb-images, we're free to rename its values to match.
We can just rename the mismatched silicon_family values to match qcom-dtb-metadata's /soc names exactly:(qcm6490→qcs6490 for qcs6490-rb3gen2-vision-kit and (qcs8300→qcs8275 for monaco-evk).
Drop fit_soc_aliases entirely since it'd no longer be needed.

Does this work or do you have any other suggestion ?

lool: That's good research @neervish, thanks. Could we base it on something like the linux kernel DTB data and use that for qcom-dtb-metadata? I am worried we went with slightly volatile and not necessarily consistent marketing names in the initial set in qcom-dtb-metadata, and I'm keen to anchor this in a place that can't be challenged (I was bitten with board names in the past and moved to track dtb names because of this ;))

neervish: > That's good research Neeraj (@neervish), thanks. Could we base it on something like the linux kernel DTB data and use that for qcom-dtb-metadata? I am worried we went with slightly volatile and not necessarily consistent marketing names in the initial set in qcom-dtb-metadata, and I'm keen to anchor this in a place that can't be challenged (I was bitten with board names in the past and moved to track dtb names because of this ;))

Hi @lool , i further checked this and got to know that the /soc names taken in qcom-dtn-metdata maps to msm-id for every dtb.

Lets consider example of the qcs6490-rb3gen2.dtb  , this dtb is used for the soc qcs6490 and board is iot which maps to msm-id qcs6490 { msm-id = <0x000001f2>; };(**498** in decimal.) as per qcom-metadata.dts.
**The name qcs6490 for above msm-id is decided based on IPCAT.**

**IPCAT shows this msm-id as QCS6490 – Kodiak IOT Without Modem | 498**

Similarly for monaco-evk.dtb , this is used for qcs8275 soc and maps to msm-id qcs8275 {msm-id = <0x000002a3>; };(**675** in decimal) as per qcom-metadata.dts.

**From** the **IPCAT** we can see this msm-id maps to this ==> **Qcs8275** – industrial Non-Safe | CHIPINFO_ID_QCS_MONACO_INDUSTRIAL | **675**


**So the name qcs8275 for above msm-id is decided based on IPCAT**

**Also the above msm-ids names matches with QCOM_ID_* constant in the kernel's include/dt-bindings/arm/qcom,ids.h exactly**
  qcm6490 { msm-id = <0x1f1>; };  →  QCOM_ID_QCM6490   497
  qcs6490 { msm-id = <0x1f2>; };  →  QCOM_ID_QCS6490   498
  qcs8300 { msm-id = <0x2a2>; };  →  QCOM_ID_QCS8300   674
  qcs8275 { msm-id = <0x2a3>; };  →  QCOM_ID_QCS8275   675

so we can not change these names in qcom-dtb-metadata as these are decided based on above information.

So for us we will have to use the soc names which are present in qcom-dtb-metadata. 
So we can use existing silicon_family itself by modifying the names of the two boards to match qcom-dtb-metadata soc.
**or** add a new field (e.g. fit_soc) per board specifically for --soc filtering, anchored to qcom-dtb-metadata's names, and leave silicon_family

i prefer the rename since silicon_family has no other consumers.
Please confirm so that i can go ahead with the changes.

lool: @neervish that's perfect:
- you confirmed we have a solid way of allocating this in IPCAT
- and further there's an upstream kernel anchor for these ids

I wish we had a doc explaining the expectations for boards yaml entries, but in the absence of that, could you please add some comments above the boards database documenting the rationale for the silicon family naming scheme? And align the boards definitions to use it consistently as you proposed? So we'd drop silicon_family and fit_soc_aliases and use soc_id or something like that.

debos-recipes/qualcomm-linux-debian-flash.yaml
neervish: Please check the latest summary

debos-recipes/qualcomm-linux-debian-flash.yaml
neervish: @lool I've pushed the requested changes in latest commit - dropped silicon_family/fit_soc_aliases in favor of soc_id, aligned all boards to their correct values, and added a comment documenting the naming rationale.
Also raised bug in qcom-dtb-metadata for qcm2290 board , captured this as an exception in PR Summary.

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -23,16 +23,21 @@ actions:
...
     description: Download qcom-dtb-metadata (DTB image build scripts and ITS/metadata files)
-    url: https://github.com/qualcomm-linux/qcom-dtb-metadata/archive/b07b6e7792b8c1a24d9758353fdefbe9591eb5a3.tar.gz
+    url: https://github.com/qualcomm-linux/qcom-dtb-metadata/archive/d30e9cdb766f08b05160ededf53556ed665d0c0b.tar.gz
     name: qcom-dtb-metadata
     filename: qcom-dtb-metadata.tar.gz
-    sha256sum: a81f39859a6dbe00cda2bc995890186f8733244ed6885d061573aaea1b39e0f4
+    sha256sum: a43ccf98149658352a74c54cac1455c35b44e7d48798dc0072540604ec475ae9
     unpack: true
 
+# silicon_family is renamed to soc_id, with values matching qcom-dtb-metadata's
```
lool: don't mention the old name, just explain the function of the soc_id field

debos-recipes/qualcomm-linux-debian-flash.yaml
```diff
@@ -307,6 +312,14 @@ actions:
           ;;
       esac
 
+      # Derive the list of SoCs supported
+      {{- $fit_soc := list }}
+      {{- range $board := $boards }}
+      {{- if ne $board.soc_id "qcm2290" }}
```
lool, line 320: Add a comment above on why qcm2290 is special cased; you can link to the qcom-dtb-metadata issue

lool (changes requested):
Hi Neeraj, now the resulting changes are almost exactly what I want to merge; just two minor tweaks to the comments in flash.yaml, and you mainly need to adjust your git history: you've incrementally added git commits on top of the previous ones, when actually what projects want to merge are the resulting commits.

Your style of development made my review easier, but the history we want merged in the main tree should be one commit per logical change.

I would propose the following 3 commits:
- one commit updating to latest qcom-dtb-metadata (the url/SHA update)
- one commit to change/add the silicon family field to align on the new standard you've identified; use a "!" in your commit message line to reflect that it's a breaking change (actually the field wasn't used, but it's still a data format change and we'll rely on the field in the future)
- one commit changing the logic to call --soc

but arguably you could squash 1) and 3) together and put them last

debos-recipes/qualcomm-linux-debian-flash.yaml
neervish: Done

debos-recipes/qualcomm-linux-debian-flash.yaml
neervish: Done

neervish:
> Hi Neeraj, now the resulting changes are almost exactly what I want to merge; just two minor tweaks to the comments in flash.yaml, and you mainly need to adjust your git history: you've incrementally added git commits on top of the previous ones, when actually what projects want to merge are the resulting commits.
> 
> Your style of development made my review easier, but the history we want merged in the main tree should be one commit per logical change.
> 
> I would propose the following 3 commits:
> 
> * one commit updating to latest qcom-dtb-metadata (the url/SHA update)
> * one commit to change/add the silicon family field to align on the new standard you've identified; use a "!" in your commit message line to reflect that it's a breaking change (actually the field wasn't used, but it's still a data format change and we'll rely on the field in the future)
> * one commit changing the logic to call --soc
> 
> but arguably you could squash 1) and 3) together and put them last

Done, restructured the git history into 2 commits per your suggestion.
  1. refactor(debos/flash)!: rename silicon_family to soc_id — renames the field and fixes the soc_id values for qcs6490-rb3gen2-vision-kit and monaco-evk to match qcom-dtb-metadata's /soc node names.
  2. fix(debos/flash): select FIT DTBs via --soc using soc_id — bumps qcom-dtb-metadata and adds the --soc filtering logic.

lool (approved):
Latest commits look great; thanks!

Note to self: I've reviewed the qcom-dtb-metadata log and diff between old rev and new rev; it's mainly the new feature we need, board additions, and also a change for el2 dtbs that we wanted.

PR #471, opened 2026-07-13
feat(ci): verify Syft installer signature before running

anujm1:
Make sure we verify the Syft install script using cosign before installation [1].

[1] https://oss.anchore.com/docs/installation/syft/

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -75,8 +75,10 @@ jobs:
 
       # mtools is needed for the flash recipe
       # file, device-tree-compiler and u-boot-tools are needed by qcom-dtb-metadata
+      # gettext-base is needed by cosign-installer, which uses envsubst internally
+      # curl is needed by cosign-installer and the Syft install script
       - name: Install debos and dependencies of the recipes and local tests
-        run: apt -y install debian-archive-keyring debos file make mmdebstrap mtools python3-pexpect python3-pytest qemu-efi-aarch64 qemu-system-arm xmlstarlet python3-defusedxml device-tree-compiler u-boot-tools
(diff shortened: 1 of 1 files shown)


.github/workflows/debos.yml
```diff
@@ -190,13 +190,17 @@ jobs:
       - name: Unpack rootfs to generate SBOM
         run: mkdir -v rootfs && tar -C rootfs -xf rootfs.tar
 
+      - name: Install cosign
+        # Required so the syft installer can verify the release signature (-v).
+        uses: sigstore/cosign-installer@v4.1.2
```
lool: Can you use a sha here, and add the actual version it's meant to point as a comment? thanks

.github/workflows/debos.yml
```diff
@@ -190,13 +190,17 @@ jobs:
...
+
       # Syft is not packaged in Debian; it's available as a binary tarball or
       # as container image from upstream; it's available on arm64 and x86
       - name: Install Syft
         run: |
           set -ux
           apt -y install curl
-          curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh
+          curl -sSfL https://raw.githubusercontent.com/anchore/syft/main/install.sh | sh -s -- -v
```
lool, line 204: Please split this change into its own commit, I don't think it's related?

lool (changes requested):
This seems not to build, it looks like you need to apt install gettext-base for this action to work with our minimal debian containers

.github/workflows/debos.yml
anujm1: I think they are related and should be one commit. cosign is being installed so `-v` switch can use it to validate the signature of binary. I can still split it if that's preferred here.

.github/workflows/debos.yml
anujm1: I will change this. Thank you for reviewing.

.github/workflows/debos.yml
lool: Ah they are indeed, I should have followed the link you provided; thanks

lool (approved):

PR #475, opened 2026-07-17
feat(debos/flash): add hamoa- and purwa-iot-evk

lool:
Add board entries for the Hamoa and Purwa IQ-X EVKs. Both use the
iq-x7181-evk ptool platform (ufs + spinor), share the HAMOA boot
binaries and the IQ-X.1.2 EVK CDT, and build a multidtb FIT; they differ
only in name, soc_id (hamoa / purwa) and dtb.

The IQ-X EVK CDT ships only as a .tar.gz, whereas every existing board's
CDT is a .zip. Make the CDT-unpack step dispatch on the archive
extension (tar for .tar.gz/.tgz, unzip otherwise) instead of assuming
.zip, so these boards can unpack their CDT.


lool:
Superseded by #476

PR #476, opened 2026-07-17
Add recent Qualcomm Linux boards to the flash recipe

lool:
Add support in qcom-deb-images for recently added Qualcomm platforms.

This required updated ptool (for shikra-evk SKUs) and otherwise just metadata additions.

- **refactor(debos/flash): bump qcom-ptool to Python package**
- **feat(debos/flash): build one multidtb FIT per SoC**
- **fix(debos/flash): shorten boards flash generation**
- **fix(debos/flash): skip multidtb SoCs with no FIT**
- **feat(debos/flash): add hamoa- and purwa-iot-evk**
- **feat(debos/flash): add shikra-evk CQM/CQS/IQS SKUs**
- **feat(debos/flash): add sm8750-mtp**
- **feat(debos/flash): add kaanapali-mtp**

### DIFF
--- a/scripts/gen-ptool.sh
+++ b/scripts/gen-ptool.sh
@@ -19,9 +19,13 @@ CDT_FILENAME="$3"
 BUILDID="$4"
 # disk storage, emmc, nvme, spinor or ufs
 DISK_TYPE="$5"
-# dtb type: multidtb (shared ../dtb-multidtb.bin at ARTIFACTDIR level) or
-# combineddtb (local dtb-combineddtb.bin in flash dir); default is combineddtb
+# dtb type: multidtb (shared ../dtb-multidtb-<soc>.bin at ARTIFACTDIR level)
+# or combineddtb (local dtb-combineddtb.bin in flash dir); default is
+# combineddtb
 DTB_TYPE="${6:-combineddtb}"
+# SoC id, used to pick the per-SoC ../dtb-multidtb-<soc>.bin; only required for
+# the multidtb dtb type
+SOC="${7:-}"
 
 PARTITIONS_CONF="${QCOM_PTOOL}/platforms/${PLATFORM}/partitions.conf"
 
@@ -47,7 +51,11 @@ esac
 
 case "$DTB_TYPE" in
   multidtb)
-    dtb="../dtb-multidtb.bin"
+    if [ -z "$SOC" ]; then
+      echo "multidtb dtb type requires a SoC id (arg 7)"
+      exit 1
+    fi
+    dtb="../dtb-multidtb-${SOC}.bin"
     ;;
   combineddtb)
     dtb="dtb-combineddtb.bin"
--- a/scripts/gen-ptool.sh
+++ b/scripts/gen-ptool.sh
@@ -74,19 +74,24 @@ partition_map="${partition_map},dtb_b=${dtb}"
 partition_map="${partition_map},efi=${esp}"
 partition_map="${partition_map},rootfs=${rootfs}"
 
+# qcom-ptool ships as the "qcom_ptool" Python package; run its scripts as
+# modules with the tree root on PYTHONPATH so intra-package imports resolve
+# without a pip install
+export PYTHONPATH="${QCOM_PTOOL}${PYTHONPATH:+:${PYTHONPATH}}"
+
 # generate ptool-partitions.xml from partitions.conf
-"${QCOM_PTOOL}/gen_partition.py" -i "${PARTITIONS_CONF}" \
+python3 -m qcom_ptool.gen_partition -i "${PARTITIONS_CONF}" \
     -o ptool-partitions.xml \
     -m "${partition_map}"
 
 # generate contents.xml from ptool-partitions.xml and contents.xml.in
 CONTENTS="${QCOM_PTOOL}/platforms/${PLATFORM}/contents.xml.in"
 if [ -e "$CONTENTS" ]; then
-    "${QCOM_PTOOL}/gen_contents.py" -p ptool-partitions.xml \
+    python3 -m qcom_ptool.gen_contents -p ptool-partitions.xml \
         -t "$CONTENTS" \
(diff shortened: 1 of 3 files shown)


lool:
Supersedes #475

neervish:
**---JFI:** With these new boards added , the soc_ids also gets added to the --soc list passed to build-dtb-image.sh, so their DTB entries get packed into dtb-multidtb.bin / qclinux_fit.img .
**This will increase that FIT image's size compared to today and can go beyond 4MB.**

As of now with existing boards we are getting qclinux_fit.img size as 3.6 MB.

lool:
@neervish so perhaps time to switch to per SoC FIT images, what do you think?

lool:
So there was an incompatible change in ptool, but it was a fairly simple adjustment to call the utilities through the python package instead of calling scripts. I made it in the same commit as the ptool update, the other commits are identical.

NB: we might want to hold off on merging this PR as it might push our common dtb.bin above our 4MiB budget. I might have to switch to per board dtb.bins instead, which was a version we explored when we landed multidtb.bin support.

obbardc:
> NB: we might want to hold off on merging this PR as it might push our common dtb.bin above our 4MiB budget. I might have to switch to per board dtb.bins instead, which was a version we explored when we landed multidtb.bin support.

Is it worth having a unit test or check to validate that dtb.bin is within the size budget ?

lool:
> > NB: we might want to hold off on merging this PR as it might push our common dtb.bin above our 4MiB budget. I might have to switch to per board dtb.bins instead, which was a version we explored when we landed multidtb.bin support.
> 
> Is it worth having a unit test or check to validate that dtb.bin is within the size budget ?

Ah I had the same idea and actually added one, but then dropped it as it was duplicating the current behavior: right now we generate a 4MiB FAT, and then copy the FIT in it, so the build fails when trying to do that mcopy. I actually think that's the normal behavior, and if we added a check, we'd have two places to maintain for that 4MiB limit.

lool:
Ok, so initial version didn't build when there was no dtb for a potentially supported SoC; added logic to skip soc-multidtb.bin generation in that case.

lool:
I've reordered commit to make more sense

lool:
@obbardc FYI with this PR I managed to generate a pathologically long action that broke MAX_ARG_STRLEN (128kiB):
```
2026/07/22 19:10:54 Action `Generate flash directories for eMMC and UFS boards` failed at stage Run, error: fork/exec /usr/bin/sh: argument list too long
```
see https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/29947413901/job/89016208523?pr=476 for a log

I don't think it's ok for my action to be so long, but I wonder if debos could use a temp file instead; might ease debugging in some cases too

lool:
I dealt with the overly long shell script by having a single shell loop that parses a table.

This also makes it easy to move this logic out of debos recipes in the future.

lool:
Had to fix another edge case with Hamoa: can't generate the FIT image when any of the dtbo is missing... so skip these boards in that case.

obbardc:
> > > NB: we might want to hold off on merging this PR as it might push our common dtb.bin above our 4MiB budget. I might have to switch to per board dtb.bins instead, which was a version we explored when we landed multidtb.bin support.
> > 
> > 
> > Is it worth having a unit test or check to validate that dtb.bin is within the size budget ?
> 
> Ah I had the same idea and actually added one, but then dropped it as it was duplicating the current behavior: right now we generate a 4MiB FAT, and then copy the FIT in it, so the build fails when trying to do that mcopy. I actually think that's the normal behavior, and if we added a check, we'd have two places to maintain for that 4MiB limit.

I think what you have is fine if the build fails if we try to overflow 4MiB limit; e.g. no need to for a separate unit test.

obbardc:
> Christopher Obbard (@obbardc) FYI with this PR I managed to generate a pathologically long action that broke MAX_ARG_STRLEN (128kiB):
> 
> ```
> 2026/07/22 19:10:54 Action `Generate flash directories for eMMC and UFS boards` failed at stage Run, error: fork/exec /usr/bin/sh: argument list too long
> ```
> 
> see https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/29947413901/job/89016208523?pr=476 for a log
> 
> I don't think it's ok for my action to be so long, but I wonder if debos could use a temp file instead; might ease debugging in some cases too

I tracked this under https://github.com/go-debos/debos/issues/739

lool:
Via Srinivas: "Neeraj verified this PR on Glymur and it is working as expected"

scripts/gen-ptool.sh
```diff
@@ -74,19 +74,24 @@ partition_map="${partition_map},dtb_b=${dtb}"
 partition_map="${partition_map},efi=${esp}"
 partition_map="${partition_map},rootfs=${rootfs}"
 
+# qcom-ptool ships as the "qcom_ptool" Python package; run its scripts as
```
obbardc, line 77: nitpick: seems like in future we may want to switch to shipping ptool as a debian package? happy to do that.

neervish:
> Via Srinivas: "Neeraj verified this PR on Glymur and it is working as expected"

Tried this PR #476 changes , It now generates 6 separate multidtb bins one per SoC(glymur, hamoa, purwa, shikra-cqm/cqs/iqs) in qcom-deb-images.
Actual FIT content is much smaller per-SoC now — e.g. glymur's(**dtb-multidtb-glymur.bin**) qclinux_fit.img is only ~278K.

obbardc (approved):
LGTM; just one comment about future of ptool.

scripts/gen-ptool.sh
lool: @obbardc Yup! that's exactly one of the motivations / suggestions that led to the ptool refactoring!

lool: Switching to a deb will take a little bit of time though, so I'll go ahead and merge this now

PR #477, opened 2026-07-17
fix(ci/glymur-crd): add more generous pattern

lool:
Serial console seems to lose some characters on glymur-crd, even
during boot ROM. Workaround the lack of detection for the early boot
for now.

### DIFF
--- a/ci/lava/glymur-crd/boot.yaml
+++ b/ci/lava/glymur-crd/boot.yaml
@@ -44,6 +44,14 @@ actions:
       - "new password"
       - sudo su
     method: minimal
+    parameters:
+      # glymur-crd's serial console drops bytes during the fast boot output
+      # burst, so the single-line default kernel-start-message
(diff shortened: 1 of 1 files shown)


gagath (approved):

gagath:
@lool Please create a ticket in this repo so we can remove this workaround when the lab situation is fixed.

lool:
@gagath At first, I was under the impression it was a transient short-term thing we'd fix quickly in a few days/weeks, but after discussing with Tools & Testing team, they don't think it's a stale open console; they will check USB cable though.

This gets me worried that we're just burying the issue, so perhaps we should just live with these failures until we design a test that will expose them enough that we address them, perhaps a serial stressor of some kind.

So I guess perhaps we should drop that PR entirely

gagath:
I have created https://github.com/qualcomm-linux/qcom-deb-images/issues/488 to track this if the serial issues appear again so we have something to refer to. Closing this PR and hoping the lab team will find the root cause of this problem.

PR #486, opened 2026-07-22
ci: Force wifi device for wifi smoke tests on qcs615-ride

mattface:
qcs615-ride wireless device is wlp1s0, so add the flag to the wlan-smoke-test lava test action.


mattface:
example [lava job](https://lava.infra.foundries.io/scheduler/job/329005)

lool (approved):
I'm happy to merge this fix right now; should we send a patch to lava-test-plans to pick up the right device automatically? e.g. first wifi device or something

Or perhaps this is a bogus device name in the first place?

mattface:
i have a feeling this is image dependent, so will need to be a per-project setting in lava-test-plans

PR #489, opened 2026-07-22
ci: scope LAVA test-artifact download per suite and run attempt

mattface:
The publish-test-results job downloads all of the run-level artifacts with no filter. LAVA action artifacts persist across re-run attempts and are shared by both the forky and trixie suite actions (same run id), so a re-run made actions/download-artifact fetch stale artifacts from superseded attempts and 404 on the orphaned ones.

Name artifacts so a filtered download fetches only the relevant, current set.

The same fix is required in test-on-pr.yaml

### DIFF
--- a/.github/workflows/lava-test.yml
+++ b/.github/workflows/lava-test.yml
@@ -64,6 +64,12 @@ jobs:
                       | map({
                           board: split("/")[0],
                           target: .,
+                          # filename-safe id per target, e.g. qrb2210-rb1-boot
+                          slug: (
+                              .
+                              | sub("\\.yaml$"; "")
+                              | gsub("/"; "-")
+                          ),
                           name: (
                               .
                               | sub("\\.yaml$"; "")
@@ -133,6 +139,13 @@ jobs:
           fail_action_on_failure: false
           save_result_as_artifact: true
           save_job_details: true
+          # Stable per-suite/per-board name: the action deletes the previous
+          # same-named artifact before uploading, so a re-run overwrites in
+          # place and only the latest result per board ever exists.
+          result_file_name: test-results-${{ inputs.suite }}-${{ matrix.slug }}
+          # Job-details artifacts always keep the (changing) LAVA job id and are
+          # never overwritten, so scope them by suite + run attempt instead.
+          test_job_file_name_prefix: ${{ inputs.suite }}-${{ github.run_attempt }}-
 
   publish-test-results:
     name: Publish Tests Results
@@ -140,10 +153,20 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Download Artifacts
+      # Artifacts live at the run level and persist across re-run attempts, so
+      # scope the download to this suite's current results only. Without this,
+      # download-artifact tries to fetch stale artifacts from superseded
+      # attempts (and the other suite) and 404s on the orphaned ones.
+      - name: Download test results (this debian suite; latest per board)
         uses: actions/download-artifact@v8.0.1
         with:
           path: artifacts
+          pattern: test-results-${{ inputs.suite }}-*
+      - name: Download job details (this debian suite, this run attempt)
+        uses: actions/download-artifact@v8.0.1
+        with:
+          path: artifacts
+          pattern: ${{ inputs.suite }}-${{ github.run_attempt }}-test-job-*
 
       - name: List files
         run: |
@@ -157,7 +180,7 @@ jobs:
 
       - name: Publish Test Job Details
         run: |
-          for json_file in $(find ${{ github.workspace }} -name "test-job-*.json")
+          for json_file in $(find ${{ github.workspace }} -name "*test-job-*.json")
           do
               DEVICE_TYPE=$(cat "$json_file" | jq -r ".requested_device_type")
               URL=$(cat "$json_file" | jq -r ".url")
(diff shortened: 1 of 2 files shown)


mattface:
Note this is largely impossible to test

lool:
Loïc ran Claude which said:

## Code Review: PR #489 — `ci: scope LAVA test-artifact download per suite and run attempt`

### Overview

The `publish-test-results` job in `lava-test.yml` (and its counterpart in `test-on-pr.yml`) called `actions/download-artifact` with **no filter**, pulling *all* run-level artifacts. Because LAVA artifacts persist across re-run attempts and are shared by both the `trixie` and `forky` suite jobs (which run under the **same run id** in `build-on-push.yml` / `build-daily.yml`), a re-run made `download-artifact` fetch stale artifacts from superseded attempts and 404 on orphaned ones.

The fix gives artifacts scoped, predictable names and switches the download to `pattern:` filters. I verified the mechanism against `foundriesio/lava-action@v13`'s `submit.js`, and the design is sound:

- **`result_file_name`** → the action does `deleteArtifact(name)` then `uploadArtifact(name)`, so a **stable** per-board name (`test-results-<suite>-<slug>`) overwrites in place — exactly one current copy per board.
- **`test_job_file_name_prefix`** → the job-details artifact name always embeds the changing LAVA job id (`<prefix>test-job-<jobId>`) and is **never** deleted/overwritten, so these accumulate. Scoping by `suite + run_attempt` is the correct way to isolate the current attempt's set.

The reasoning is accurate and the two artifact types are handled differently for the right reasons.

### Strengths

- **Excellent comments.** Each non-obvious choice (why results use a stable name but job-details use `run_attempt`; the cross-file coupling note) is documented inline. This is the hard part of CI code and it's done well.
- **Correct `find` pattern update.** Changing `-name "test-job-*.json"` → `-name "*test-job-*.json"` correctly accounts for the new prefix (`<suite>-<attempt>-test-job-*.json`).
- The `slug` jq addition mirrors the existing `name` transform cleanly and is filename-safe.

### Issues & Suggestions

**1. `slug` is more permissive than the actual paths (minor / defensive).**
`name` uses `gsub("/"; " ")` and `slug` uses `gsub("/"; "-")` after stripping `.yaml`. Given today's flat `board/test.yaml` layout (all 7 templates are exactly 2 levels), slugs like `qrb2210-rb1-boot` are unique. But `result_file_name` has no `run_attempt`, so uniqueness across boards is load-bearing — two targets that slugged identically would silently overwrite each other's results. Not a bug today; worth a one-line comment that the slug must stay unique per board.

**2. Comment says "latest per board" but scoping is per *suite*, not per board (wording).**
The step name `Download test results (this debian suite; latest per board)` is accurate, but the *uniqueness* guarantee ("latest per board") comes from the stable `result_file_name`, not the download pattern. Fine as-is; just noting the two mechanisms are easy to conflate when reading.

**3. Hard-coded `trixie` coupling in `test-on-pr.yml` (already flagged in the PR, acceptable).**
The new comment explicitly calls out that `suite: trixie` must stay in sync with the download patterns. This is the right call given `test-on-pr.yml`'s `publish-test-results` is a top-level job (post-`workflow_run`) that can't read the reusable workflow's input. The explicit `suite: trixie` input (even as a documented no-op) plus the sync comment is a reasonable trade-off.

**4. Behavior on zero matches (verify, low risk).**
If a suite legitimately produces no matching artifacts (e.g. all boards excluded), `download-artifact` with a non-matching `pattern` logs a warning and succeeds (it does not fail), and the downstream `find` loops simply produce no output — same as before. This matches the prior no-filter behavior closely enough.

**5. Consistency: `test-on-pr.yml` was not given the `slug` treatment — correct.**
`test-on-pr.yml`'s `publish` job only consumes the patterns; the `slug`/naming is produced in `lava-test.yml`. The split is correct; no duplication needed.

### Correctness / Risk Assessment

- **No security impact** — permissions unchanged; only artifact names and download filters.
- **No performance concern** — filtered downloads are strictly cheaper than the prior download-everything.
- **Backward compatible** — first run after merge produces newly-named artifacts; the download patterns match them. No migration issue since artifacts are per-run.
- **Test coverage** — this is CI plumbing; correctness is validated by the workflow itself. The real proof is a green re-run, which is the scenario the PR targets.

### Verdict

**Approve.** This is a well-reasoned, well-documented fix that correctly distinguishes the two artifact lifecycles (overwrite-in-place vs. accumulate-by-job-id). The only substantive latent risk is slug uniqueness for `result_file_name` (#1), which is safe under the current flat layout but load-bearing enough to deserve a guard comment. Everything else is polish.

One question for the author, if you want to future-proof: would it be worth having `test-on-pr.yml` derive its patterns from a shared constant/output rather than duplicating the literal `trixie`? Not blocking — the sync comment covers it — but it's the one spot that could silently drift.

lool (approved):
Hey, I think the approach made sense, and would address the recent failure I reported:
https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/29899983158/job/88914475840

This would deal with versioning artifacts across rebuild attempts; the only thing I spotted was the hardcoding of trixie, which is a wider thing we need to tackle, but I think this great for now!

mattface:
Since Claude somewhat wrote that PR, it would have been quite embarrassing for it to not approve it :joy:

PR #490, opened 2026-07-23
fix(build-linux-deb): add missing dep build-essential

gagath:


lool:
I'm curious, why would build-essential be needed?

gagath:
> I'm curious, why would build-essential be needed?

build-essential:native is required to build any package, including the upstream linux package. Without it, the invocation of `make bindep-pkg` will eventually call dpkg-buildpackage which will fail if `build-essential:native` is not installed on the system.

(added this to the commit description)

gagath:
Not stale, please review this.

PR #491, opened 2026-07-23
ci: update wifi device name on some boards to wlp1s0

mattface:
The devices lemans-evk, monaco-evk and qcs8300-ride enumerate their wifi device as wlp1s0, so we need to set the LAVA test action to use this instead of default wlan0. Same change as for qcs615-ride.

### DIFF
--- a/ci/lava/lemans-evk/boot.yaml
+++ b/ci/lava/lemans-evk/boot.yaml
@@ -60,6 +60,7 @@ actions:
       parameters:
         SKIP_INSTALL: "True"
         WAIT_TIME: 60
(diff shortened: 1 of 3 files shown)


obbardc (approved):
LGTM; would it be good if the test could autodetermine the device as a follow-up ?

lool (approved):
Thanks @obbardc and @mattface!

lool:
> LGTM; would it be good if the test could autodetermine the device as a follow-up ?

https://github.com/Linaro/test-definitions/pull/631

PR #494, opened 2026-07-24
ci(linux): publish unstripped vmlinux as artifact

lool:
The kernel deb build only publishes the .deb packages, whose
linux-image ships a stripped Image. That leaves no way to get good
oops/panic backtraces from kernel crashes.

make bindeb-pkg leaves an unstripped linux/vmlinux in the source tree,
carrying the kernel symbol table.  Copy it into the artifacts
directory so it flows to both the EFS and S3 uploads, giving
crash-debugging parity with the vmlinux artifact meta-qcom's Yocto
CI publishes.


obbardc (approved):
LGTM!

PR #497, opened 2026-07-27
feat(debos/flash)!: update Glymur boot binaries to 00079

jayantsa1:
Known fixes:

EFI Variables Changes & Multi-DTB Support

Extended EFI variable handling to support runtime read/write operations aligned with UEFI BDS boot flow.

AOSS Sleep Issue Fixes

Addressed AOSS (Always-On SubSystem) sleep entry/exit failures that were preventing the platform from achieving deep low-power states on Glymur LNX.

FastRPC Related Fix

Resolved a FastRPC communication failure that was causing intermittent DSP session establishment errors.


lool (changes requested):
Hi,
the commit message body is ok, but the commit message title does not follow the established pattern of conventional commits; we don't have an automated check of that, and I know you need this change soon, but please address this in your next PR.

Also, the changes are Glymur specific, but this is not conveyed in the PR/commit title.

PR #499, opened 2026-07-27
ci: glymur: exclude glymur-crd.dtb hardcode

phanendra-qc:
Remove hardcoded glymur dtb passage to efi with below Boot bins PR bringing support of multi-dtb. 

Depends-on : https://github.com/qualcomm-linux/qcom-deb-images/pull/497


lool (changes requested):
@phanendra-qc that's awesome to see!

Two things:
- can you please reword your commit message to follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), in the style of other commit messages in this repo
- can you please add a commit to drop the workflows/daily CI specific to glymur CRD? I believe the custom DTB is the last reason we needed those

PR #501, opened 2026-07-28
ci: exclude glymur-crd.dtb hardcode.

phanendra-qc:
- As PR#497 adding support of multi-dtb

### DIFF
--- a/.github/workflows/linux-daily-glymur.yml
+++ b/.github/workflows/linux-daily-glymur.yml
@@ -21,7 +21,6 @@ jobs:
     with:
       kernel_name: glymur
       build_linux_deb_extra_args: '--qcom-next --ref qcom-next-7.2-rc3-20260722 prune.config qcom.config'
-      debos_extra_args: '-t dtb:qcom/glymur-crd.dtb'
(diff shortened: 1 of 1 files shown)


lool (changes requested):
Hi

Why did you open another PR? We should have continued in #499, now this PR review starts without the context of the other one.

1) the first line of the commit message kind of tries to follow Convention Commits, the trailing dot is weird though, it just eats one char for now value, please drop

2) The commit message body is not good; it mentions https://github.com/qualcomm-linux/qcom-deb-images/pull/497 without explaining what the PR is about, so people have to look it up. Having a reference is nice, but that should come after conveying the "why" in a human readable way. On top of that, it would be nice if GH made that a proper link I can click, but I'm not sure why it's not, perhaps you need a space between PR and #497.

3) Privately, you shared that the current qcom-next images do NOT boot without hardcoding the dtb, so we need to get to the bottom of that first, before changing this.

4) we discussed in the other PR how we should take the opportunity to remove the glymur CI entirely if it has no purpose anymore.

phanendra-qc:
closed as moving with qcom-next workflow

Issue #498, opened 2026-07-27
snapshot not used when running mmdebstrap


basak-qcom:
With the snapshot support introduced in #422, I see that mmdebstrap is run without the snapshot as defined by the caller. mmdebstrap automatically adds -updates and -security for anything that isn't unstable or testing. Both forky and trixie will therefore move forward for baseline packages even though a snapshot is selected.

I'm currently working on #495 but filing this as a separate issue.

obbardc:
The mmdebstrap docs say:

> If I<SUITE> does not refer to "unstable" or "testing", then
> I<SUITE>-updates and I<SUITE>-security mirrors are automatically added.

which I think is a trap; mmdebstrap only adds those suites when it is given no mirror argument at
all (see the logic in `get_sourceslist_by_suite()`). I guess this could be clarified in the `mmdebstrap` documentation?

Since the debos mmdebstrap action always passes mirrors to mmdebstrap and our recipe sets `mirrors: ['http://deb.debian.org/debian']`, the  `-updates`/`-security` sources are not written during the mmdebstrap call; those suites are only added later when we write `*.sources` ourselves.

So all we need to do here is change `mirrors` to the snapshot URL:

```yaml
mirrors: ['https://snapshot.debian.org/archive/debian/{{ $snapshot }}/']
apt-opts: ['Acquire::Check-Valid-Until "false"']
```

The `Check-Valid-Until` is also required otherwise an older snapshot fails with `Release file ... is expired (invalid since 1159d ...)`.

Also `apt-opts` are written into `/etc/apt/apt.conf.d/99mmdebstrap` inside the target. If it isn't deleted after bootstrapping, we would ship images with `Check-Valid-Until` disabled globally for on-device `apt` which could cause a security regression. Since we write `Check-Valid-Until: no` in the `snapshot_*.sources` later, the generated `99mmdebstrap` can simply be removed once the bootstrap is done.

PR #505, opened 2026-07-28
Add snapshot support for qli apt repository

basak-qcom:
Extend the existing snapshot support to function when using the qli apt
repository in addition to Debian trixie, trixie-security and trixie-backports.

This does not address #498 or #293.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,8 +154,8 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
 - `kernelpackage`: **deprecated**, superseded by `kernelpackages`; still
   accepted as a single package name for backwards compatibility.
 - `suite`: Debian suite to use, defaults to `trixie`.
-- `snapshot`: use a Debian snapshot archive for a reproducible build (`YYYYMMDD`
-  or `YYYYMMDDTHHMMSSZ`); logged to `/etc/buildinfo` as `SNAPSHOT=<date>`.
+- `snapshot`: use a Debian snapshot archive for a reproducible build
+  (`YYYYMMDDTHHMMSSZ`); logged to `/etc/buildinfo` as `SNAPSHOT=<date>`.
   Live mirrors are restored in the final image to allow upgrades.
 
 For the image recipe:
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
 - `snapshot`: use snapshot apt archives for a reproducible build
   (`YYYYMMDDTHHMMSSZ`); logged to `/etc/buildinfo` as `SNAPSHOT=<date>`.
   Live mirrors are restored in the final image to allow upgrades.
+  See [docs/snapshot.md](docs/snapshot.md) for full usage and internals.
 
 For the image recipe:
 
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
 - `kernelpackage`: **deprecated**, superseded by `kernelpackages`; still
   accepted as a single package name for backwards compatibility.
 - `suite`: Debian suite to use, defaults to `trixie`.
-- `snapshot`: use a Debian snapshot archive for a reproducible build
+- `snapshot`: use snapshot apt archives for a reproducible build
   (`YYYYMMDDTHHMMSSZ`); logged to `/etc/buildinfo` as `SNAPSHOT=<date>`.
   Live mirrors are restored in the final image to allow upgrades.
 
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -103,10 +103,10 @@ actions:
     command: |
       set -eux
 {{- if $snapshot }}
-      if echo "{{ $snapshot }}" | grep -qE '^([0-9]{8}|[0-9]{8}T[0-9]{6}Z)$'; then
+      if echo "{{ $snapshot }}" | grep -qE '^[0-9]{8}T[0-9]{6}Z$'; then
         SNAPSHOT="{{ $snapshot }}"
       else
-        echo "ERROR: invalid snapshot '{{ $snapshot }}'. Use: YYYYMMDD | YYYYMMDDTHHMMSSZ" >&2
+        echo "ERROR: invalid snapshot '{{ $snapshot }}'. Use: YYYYMMDDTHHMMSSZ" >&2
         exit 1
       fi
       touch /etc/buildinfo && chmod 644 /etc/buildinfo
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -141,18 +141,6 @@ actions:
 {{- end }}
       EOF
 
-{{- if $snapshot }}
-      # Derive snapshot_debian.sources from debian.sources by rewriting mirror URLs
-      sed \
-        -e "s#http://deb\.debian\.org/debian/#https://snapshot.debian.org/archive/debian/${SNAPSHOT}/#g" \
-        -e "s#http://deb\.debian\.org/debian-security/#https://snapshot.debian.org/archive/debian-security/${SNAPSHOT}/#g" \
-        -e "/^Enabled:/a\\# TODO: remove Check-Valid-Until once Debusine-based snapshots are available" \
-        -e "/^Enabled:/a\\Check-Valid-Until: no" \
-        /etc/apt/sources.list.d/debian.sources \
-        > /etc/apt/sources.list.d/snapshot_debian.sources
-      apt-snapshot-toggle enable
-{{- end }}
-
 {{- if and (ne $suite "unstable") (ne $suite "testing") }}
 
 {{- /* prefer installing these packages from backports */}}
@@ -174,9 +162,6 @@ actions:
     chroot: true
     command: |
       set -eux
-{{- if $snapshot }}
-      SNAPSHOT=$(grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2)
-{{- end }}
       tee /etc/apt/sources.list.d/debian-backports.sources <<EOF
       # Debian backports
       Types: deb
@@ -187,17 +172,6 @@ actions:
       Signed-By: /usr/share/keyrings/debian-archive-keyring.gpg
       EOF
 
-{{- if $snapshot }}
-      # Derive snapshot_debian-backports.sources from debian-backports.sources by rewriting mirror URLs
-      sed \
-        -e "s#http://deb\.debian\.org/debian/#https://snapshot.debian.org/archive/debian/${SNAPSHOT}/#g" \
-        -e "s#http://deb\.debian\.org/debian\$#https://snapshot.debian.org/archive/debian/${SNAPSHOT}/#" \
-        -e "/^Enabled:/a\\# TODO: remove Check-Valid-Until once Debusine-based snapshots are available" \
-        -e "/^Enabled:/a\\Check-Valid-Until: no" \
-        /etc/apt/sources.list.d/debian-backports.sources \
-        > /etc/apt/sources.list.d/snapshot_debian-backports.sources
-      apt-snapshot-toggle enable
-{{- end }}
       tee /etc/apt/preferences.d/debian-backports.pref <<EOF
       # for binary packages built from these source packages, score the version from
       # Debian backports higher as to get hardware enabled or better hardware support
@@ -256,9 +230,6 @@ actions:
     chroot: true
     command: |
       set -eux
-{{- if $snapshot }}
-      SNAPSHOT=$(grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2)
-{{- end }}
       tee /etc/apt/sources.list.d/qli.sources <<EOF
       # Qualcomm Linux repository
       Types: deb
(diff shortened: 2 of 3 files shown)


basak-qcom:
I've tested this locally. When running just the rootfs recipe with `-t snapshot:20260720T0000Z`, I see that the snapshot URL is being used when apt updates, but then snapshots are disabled at the end, as expected. `apt-snapshot-toggle enable` re-enables the snapshots, as expected. After rebuilding from clean all the way through to the image and loopback mounting `disk-sdcard.img2`, I see that `sources.list.d/` is as expected (without snapshots) and `SNAPSHOT=` is set in `/etc/buildinfo`.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -256,12 +256,16 @@ actions:
     chroot: true
     command: |
       set -eux
+{{- if $snapshot }}
+      SNAPSHOT=$(grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2)
```
obbardc: It isn't clear how the `SNAPSHOT` variable is set in `/etc/buildinfo` just by looking at this commit. Perhaps you could explain that in the commit message.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -256,12 +256,16 @@ actions:
...
+{{- end }}
       tee /etc/apt/sources.list.d/qli.sources <<EOF
       # Qualcomm Linux repository
       Types: deb
       URIs: https://deb.debusine.qualcomm.com/qualcomm/qli
       Suites: {{ $suite }}
       Components: main contrib non-free-firmware non-free
+      Enabled: yes
```
obbardc, line 239: What does `Enabled: yes` mean? Just by reading this it doesn't sound like it's related to snapshots at all; perhaps it needs to be in a separate, previous commit?

README.md
```diff
@@ -154,8 +154,8 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
 - `kernelpackage`: **deprecated**, superseded by `kernelpackages`; still
   accepted as a single package name for backwards compatibility.
 - `suite`: Debian suite to use, defaults to `trixie`.
-- `snapshot`: use a Debian snapshot archive for a reproducible build (`YYYYMMDD`
-  or `YYYYMMDDTHHMMSSZ`); logged to `/etc/buildinfo` as `SNAPSHOT=<date>`.
+- `snapshot`: use a Debian snapshot archive for a reproducible build
```
obbardc: I think it could be clearer that this field is specific to Qualcomm's Debusine instance.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
obbardc: AH; ok; it is set already before in the recipe if building with a `snapshot` debos template variable. I guess that wasn't very clear and I had to read the recipe to understand that. I think clearing that up would make this commit stronger.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -277,17 +248,56 @@ actions:
...
+          name=$(basename "$src" .sources)
+          # skip already-derived snapshot files
+          case "$name" in snapshot_*) continue ;; esac
+          # table: per-source mirror URL (assert) -> snapshot rewrite (sed exprs);
+          # sources with no snapshot archive fall through and are skipped
+          case "$name" in
+            debian)
+              assert='http://deb.debian.org/debian'
+              set -- \
+                -e "s#http://deb\.debian\.org/debian/#https://snapshot.debian.org/archive/debian/${SNAPSHOT}/#g" \
```
obbardc, line 270: Since you are using `$SNAPSHOT` against `https://snapshot.debian.org`, which I assume uses the same timestamp format as debusine, the previous commits where you change the snapshot timestamp format should be more generic and refer to both debusine and snapshots.d.o.

obbardc (changes requested):
Looks good overall; just some small nitpicks.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: Enabled: yes/no is a standard field in sources entry that was introduced with this new format; we use it in the snapshot implementation to flip flop between snapshot vs regular sources easily.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -273,6 +277,17 @@ actions:
        =bX6V
        -----END PGP PUBLIC KEY BLOCK-----
       EOF
+
+{{- if $snapshot }}
+      # Derive snapshot_qli.sources from qli.sources by rewriting mirror URLs
+      sed \
+        -e "s#https://deb\.debusine\.qualcomm\.com/qualcomm/qli#&/${SNAPSHOT}#" \
+        -e "/^Enabled:/a\\# TODO: remove Check-Valid-Until once Debusine-based snapshots are available" \
```
lool: I think we don't need the Check-Valid-Until for Debusine repositories?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -277,17 +248,56 @@ actions:
...
 {{- if $snapshot }}
-      # Derive snapshot_qli.sources from qli.sources by rewriting mirror URLs
-      sed \
-        -e "s#https://deb\.debusine\.qualcomm\.com/qualcomm/qli#&/${SNAPSHOT}#" \
-        -e "/^Enabled:/a\\# TODO: remove Check-Valid-Until once Debusine-based snapshots are available" \
-        -e "/^Enabled:/a\\Check-Valid-Until: no" \
-        /etc/apt/sources.list.d/qli.sources \
-        > /etc/apt/sources.list.d/snapshot_qli.sources
+  - action: run
+    description: Derive snapshot_*.sources from live APT sources
```
lool, line 255: Thanks for consolidating that, this bugged me too!

Perhaps this would be easier to maintain as a standalone script though?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
obbardc: I understand the purpose of the field, I just don't think it's explained well enough in the commit message; it's just touched upon :-)

basak-qcom: Resolved (see general comment).

debos-recipes/qualcomm-linux-debian-rootfs.yaml
basak-qcom: Sure, I can pull all of this out into a separate script. Probably the toggling script should be pulled out as well. Ideally neither the derivation script nor the toggle script belong in the rootfs at all. Do we have a mechanism to do that in debos? Is this feasible?

This feels like it's out of scope of this PR though. I'm already improving on the state of things, but don't want to block on making things perfect just to add one entry to the implementation that was already there. If we intend to do it, I can create an issue to track it as future work.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
basak-qcom: I believe there is only one universal snapshot timestamp format, which is YYYYMMDDTHHMMZ. I edited the commit message to make it clear that generality is the intent.

README.md
basak-qcom: Done.

basak-qcom (commented):
Thank you for the reviews!

I accept Chris' points about it not being clear what all these different pieces do, and that this should be better documented. However, my PR and its commits merely extend what was there before, so I don't think an explanation of what was there before belong in _these_ commits. I also think that commit messages aren't a great place to explain what is there, only why they're being changed. Instead, I think documentation that helps people understand the code or how to use it belongs in the source tree itself. So I've added documentation for this in a separate commit that adds a docs/snapshot.md that explains all of it. This comes after my commits, since it's not worth writing documentation for the old stuff and then update that as I update the code at this stage. But for the future, any commit message that touches snapshots could refer to and/or update docs/snapshot.md at the same time, which I hope resolves Chris' concerns about documentation.

Diffing against my previous branch I see only documentation changes, so no need to retest.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
obbardc: Thanks. I can't resolve the thread for some reason; but your change resolves my original concern.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
obbardc: Thanks. I can't resolve the thread for some reason; but your change resolves my original concern.

lool:
@obbardc Hey, can you give another review to latest changes from Robie? I'm good on my side

debos-recipes/qualcomm-linux-debian-rootfs.yaml
basak-qcom: Discussed with Loic. We believe that Debusine's model is that it does not generate repository indexes at snapshot serving time. Instead they are stored as artifacts, so the artifact retention system is used to serve historically generated indexes. Thus they won't be resigned, and therefore Check-Valid-Until won't work in Debusine's case either and we must ignore it.

obbardc: I think it is worth writing this in a comment under the `{{- if $snapshot }}` line.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -273,6 +277,17 @@ actions:
        =bX6V
        -----END PGP PUBLIC KEY BLOCK-----
       EOF
+
+{{- if $snapshot }}
+      # Derive snapshot_qli.sources from qli.sources by rewriting mirror URLs
```
obbardc: nitpick time: I think it is quite fragile to be doing sed replacements over the `qli.sources` file; to me overwriting the file with the full snapshot sources file would be cleaner. Or even having it templated so it writes either sources template based on if $snapshot is enabled or not.

But I think that could be a follow-up, it's a really minor nitpick.

obbardc: OH; you do this in a later commit anyway - fine. Will leave the review comment here anyway so you can see my thinking.

docs/snapshot.md
```diff
@@ -0,0 +1,236 @@
...
+dated Qualcomm Linux archive), so that a build performed today and a build
+performed months from now resolve to the *same* package versions.
+
+The snapshot date used is recorded in the image at `/etc/buildinfo` as
+`SNAPSHOT=<date>`, so any image can be traced back to the archive state it was
+built from.
+
+## Using snapshots
+
+### The `snapshot` option
```
obbardc, line 29: Can you help me check my understanding?

When you set `snapshot` to a snapshot timestamp, it enables snapshots for that snapshot date on both the debian and qli repos; fine.

But won't the snapshot timestamp not be different for the debian && qli repos as the snapshots are decoupled?

If so, maybe it makes sense to split the variable into two; `snapshot_debian || snapshot` && `snapshot_qli`.

Apologies if I am getting this wrong; I possibly don't understand how the QLI repo does snapshotting. Maybe they're the same date; if so it may make sense to highlight that in the docs?

obbardc (commented):
Implementation looks fine to me, just couple of clarifications needed.

docs/snapshot.md
basak-qcom: I expect the snapshot timestamp to be global. Both Debian and Debusine accept any snapshot timestamp. It doesn't have to be the timestamp that matches some kind of publication event. The infrastructure will resolve that to the apt repository state that existed at the time of the timestamp specified, which in practice means looking backwards from that timestamp to what was the most recent set of indexes published at that time.

At least this is my understanding.

I'm not sure if that answers your question?

obbardc: > The infrastructure will resolve that to the apt repository state that existed at the time of the timestamp specified, which in practice means looking backwards from that timestamp to what was the most recent set of indexes published at that time.

That answers my question perfectly! I think it'd be good if you could note that in the documentation though.

README.md
obbardc: Great; thread can be resolved (still cannot do that).

debos-recipes/qualcomm-linux-debian-rootfs.yaml
obbardc: I agree; improving it in a follow-up PR would be good. Can you create a separate issue to track it and close this thread?

obbardc (approved):
Just some minor nitpicks.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
basak-qcom: I think this is already covered in the explanation right by the `Check-Valid-Until: no` insertion code:

> rewrite the mirror URL to the dated snapshot archive and disable Check-Valid-Until (snapshots are static, so the Release becomes stale)

debos-recipes/qualcomm-linux-debian-rootfs.yaml
basak-qcom: Created #534

docs/snapshot.md
basak-qcom: Done, and I enhanced the general explanation of why Check-Valid-Until must be disabled at the same time.

PR #510, opened 2026-07-29
Add armhf image support

lumag:
Qualcomm has a number of 32-bit ARM platforms that cannot run the arm64 images this project produces, so there is currently no way to get a ready-made Debian image for them. This PR makes the architecture a first-class parameter of the debos recipes, so that armhf images can be built alongside the arm64 ones.

Making room for a second architecture means the artifacts have to say which one they belong to, so every build artifact is now named after the architecture it was built for (`rootfs-arm64.tar`, `disk-ufs-armhf.img`, ...). This renames the arm64 artifacts, but only inside the build tree: the artifacts published by CI keep exactly the names they have today, the flashing pipeline and its rawprogram XMLs are untouched, and the old names stay available as symlinks so that existing scripts and instructions keep working.

armhf platforms don't boot via EFI, so the armhf images deliberately drop the ESP and systemd-boot and consist of a single root partition; the kernel defaults to Debian's `linux-image-armmp`, since the arm64 kernel packages don't apply. CI builds and publishes the armhf artifacts in the distro build workflows (PR/push/daily), but not in the kernel-validation pipelines, whose EFS kernels are arm64-only.

Nothing checked that a produced image can actually boot, so the qemu tests grew an armhf boot test, and it immediately paid for itself: the armhf images could not mount their root filesystem at all. Their initramfs contains none of the storage drivers, because initramfs-tools asks dracut-install to install whole module directories, dracut-install expands those with glibc's `fts(3)`, and fts reads directories with the 32-bit `readdir()`, which fails with `EOVERFLOW` on the 64-bit directory cookies ext4 hands out. On a 32-bit architecture every such directory therefore looks empty and is skipped without an error. This is Debian [#1110545](https://bugs.debian.org/1110545), with [#1079080](https://bugs.debian.org/1079080) for the failure going unreported. Naming the few modules the boot test needs works around it here, but an armhf image still lacks the MMC and UFS drivers it would need on real hardware: that has to be fixed in Debian, and the workaround says so.

The first commits are independent fixes made along the way: containerized `make` builds could not open `/dev/kvm` as the non-root container user, which broke the KVM fakemachine backend those builds force; the `overlays` workflow input was interpolated straight into a `run:` script, which both zizmor and Semgrep flag (also sent separately as #511, so this commit disappears if that lands first); and a build left several kinds of files untracked.

Validated locally: both armhf images build with `make USE_CONTAINER=yes`, carry an armhf userland with the armmp kernel and the expected single-partition GPT layout with the 32-bit ARM root partition type, and the boot test takes the rebuilt image all the way to a login prompt.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -27,6 +27,13 @@ on:
         description: Skip QEMU tests
         type: boolean
         default: false
+      build_armhf:
+        description: >
+          Also build an armhf rootfs and disk images; armhf always uses the
+          recipe's default kernel (Debian's linux-image-armmp) since the
+          kernelpackages input refers to arm64-only kernels
+        type: boolean
+        default: false
 
     outputs:
       artifacts_url:
@@ -148,6 +155,24 @@ jobs:
               --print-recipe \
               debos-recipes/qualcomm-linux-debian-rootfs.yaml
 
+      - name: Build armhf rootfs with debos
+        if: ${{ inputs.build_armhf }}
+        env:
+          OVERLAYS: ${{ inputs.overlays }}
+        run: |
+          set -ux
+          # NB: no kernelpackages/aptlocalrepo here; the default-ci and EFS
+          # kernels are arm64 only, so armhf sticks to the recipe's default
+          # kernel (Debian's linux-image-armmp)
+          debos \
+              -t architecture:armhf \
+              -t overlays:"$OVERLAYS" \
+              -t xfcedesktop:true \
+              -t "buildid:${BUILD_ID}" \
+              ${DEBOS_EXTRA_ARGS} \
+              --print-recipe \
+              debos-recipes/qualcomm-linux-debian-rootfs.yaml
+
       - name: Build UFS and SD card images with debos
         run: |
           set -ux
@@ -160,6 +185,15 @@ jobs:
           # builds both disk images and the unsuffixed compatibility symlinks
           # the flashing pipeline expects
           make arm64 FAKEMACHINE_OPTS="-b qemu" EXTRA_DEBOS_OPTS="--print-recipe ${DEBOS_EXTRA_ARGS}"
+
+      - name: Build armhf UFS and SD card images with debos
+        if: ${{ inputs.build_armhf }}
+        run: |
+          set -ux
+          # rootfs-armhf.tar was built by the armhf rootfs step above, so make
+          # only runs the image recipe here (same pattern as for arm64)
+          make disk-ufs-armhf.img FAKEMACHINE_OPTS="-b qemu" EXTRA_DEBOS_OPTS="--print-recipe ${DEBOS_EXTRA_ARGS}"
+          make disk-sdcard-armhf.img FAKEMACHINE_OPTS="-b qemu" EXTRA_DEBOS_OPTS="--print-recipe ${DEBOS_EXTRA_ARGS}"
 
       - name: Build flashable files with debos
         run: |
@@ -179,6 +213,13 @@ jobs:
           gzip -c disk-ufs-arm64.img >"${dir}/${PREFIX}-disk-ufs.img.gz"
           gzip -c disk-sdcard-arm64.img >"${dir}/${PREFIX}-disk-sdcard.img.gz"
           cp -av dtbs-arm64.tar.gz "${dir}/${PREFIX}-dtbs.tar.gz"
+          # armhf artifacts are only present when the build_armhf input is set
+          if [ -f rootfs-armhf.tar ]; then
+              gzip -c rootfs-armhf.tar >"${dir}/${PREFIX}-rootfs-armhf.tar.gz"
+              gzip -c disk-ufs-armhf.img >"${dir}/${PREFIX}-disk-ufs-armhf.img.gz"
+              gzip -c disk-sdcard-armhf.img >"${dir}/${PREFIX}-disk-sdcard-armhf.img.gz"
+              cp -av dtbs-armhf.tar.gz "${dir}/${PREFIX}-dtbs-armhf.tar.gz"
+          fi
           # create tarballs with support for all UFS and all eMMC boards
           scripts/bundle-flash-dirs.sh "${dir}" "${PREFIX}"
           # generate sha256sums for all artifacts
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -104,6 +104,8 @@ jobs:
           )
 
       - name: Build rootfs with debos
+        env:
+          OVERLAYS: ${{ inputs.overlays }}
         run: |
           set -ux
           kernel_packages="$KERNEL_PACKAGES"
@@ -137,7 +139,7 @@ jobs:
           esac
 
           debos \
-              -t overlays:'${{ inputs.overlays }}' \
+              -t overlays:"$OVERLAYS" \
               -t xfcedesktop:true \
               -t aptlocalrepo:${PWD}/local-apt-repo \
               -t kernelpackages:"$kernel_packages" \
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -157,8 +157,9 @@ jobs:
           # qemu backend also requires to set scratchsize, otherwise the
           # whole build is done from memory and the out of memory killer
           # gets triggered
-          make disk-ufs.img FAKEMACHINE_OPTS="-b qemu" EXTRA_DEBOS_OPTS="--print-recipe ${DEBOS_EXTRA_ARGS}"
-          make disk-sdcard.img FAKEMACHINE_OPTS="-b qemu" EXTRA_DEBOS_OPTS="--print-recipe ${DEBOS_EXTRA_ARGS}"
+          # builds both disk images and the unsuffixed compatibility symlinks
+          # the flashing pipeline expects
+          make arm64 FAKEMACHINE_OPTS="-b qemu" EXTRA_DEBOS_OPTS="--print-recipe ${DEBOS_EXTRA_ARGS}"
 
       - name: Build flashable files with debos
         run: |
@@ -171,11 +172,13 @@ jobs:
           # create a directory for the current run
           dir="debos-artifacts"
           mkdir -v "${dir}"
-          # compress output files directly into the staging directory
-          gzip -c rootfs.tar >"${dir}/${PREFIX}-rootfs.tar.gz"
-          gzip -c disk-ufs.img >"${dir}/${PREFIX}-disk-ufs.img.gz"
-          gzip -c disk-sdcard.img >"${dir}/${PREFIX}-disk-sdcard.img.gz"
-          cp -av dtbs.tar.gz "${dir}/${PREFIX}-dtbs.tar.gz"
+          # compress output files directly into the staging directory; the
+          # published arm64 artifacts keep the names they had before the build
+          # artifacts gained an architecture suffix
+          gzip -c rootfs-arm64.tar >"${dir}/${PREFIX}-rootfs.tar.gz"
+          gzip -c disk-ufs-arm64.img >"${dir}/${PREFIX}-disk-ufs.img.gz"
+          gzip -c disk-sdcard-arm64.img >"${dir}/${PREFIX}-disk-sdcard.img.gz"
+          cp -av dtbs-arm64.tar.gz "${dir}/${PREFIX}-dtbs.tar.gz"
           # create tarballs with support for all UFS and all eMMC boards
           scripts/bundle-flash-dirs.sh "${dir}" "${PREFIX}"
           # generate sha256sums for all artifacts
@@ -192,7 +195,7 @@ jobs:
           destination: ${{ github.repository_owner }}/${{ github.event.repository.name }}/${{ github.run_id }}-${{ github.run_attempt }}/
 
       - name: Unpack rootfs to generate SBOM
-        run: mkdir -v rootfs && tar -C rootfs -xf rootfs.tar
+        run: mkdir -v rootfs && tar -C rootfs -xf rootfs-arm64.tar
 
       - name: Install cosign
         # Required so the syft installer can verify the release signature (-v).
--- a/Makefile
+++ b/Makefile
@@ -64,8 +64,15 @@ all: arm64
 .PHONY: arm64
 arm64: disk-ufs-arm64.img disk-sdcard-arm64.img $(COMPAT_LINKS)
 
+# armhf images are not built by default; ask for them with `make armhf`
+.PHONY: armhf
+armhf: disk-ufs-armhf.img disk-sdcard-armhf.img
+
 rootfs-arm64.tar dtbs-arm64.tar.gz: debos-recipes/qualcomm-linux-debian-rootfs.yaml
 	$(DEBOS_CMD) -t architecture:arm64 $<
+
+rootfs-armhf.tar dtbs-armhf.tar.gz: debos-recipes/qualcomm-linux-debian-rootfs.yaml
+	$(DEBOS_CMD) -t architecture:armhf $<
 
 DISK_UFS_ARM64_IMAGES := disk-ufs-arm64.img \
 	disk-ufs-arm64.img1 \
@@ -74,12 +81,25 @@ DISK_UFS_ARM64_IMAGES := disk-ufs-arm64.img \
 $(DISK_UFS_ARM64_IMAGES): debos-recipes/qualcomm-linux-debian-image.yaml rootfs-arm64.tar
 	$(DEBOS_CMD) -t architecture:arm64 $<
 
+# armhf images have no ESP, so there is no second partition to extract
+DISK_UFS_ARMHF_IMAGES := disk-ufs-armhf.img \
+	disk-ufs-armhf.img1
+
+$(DISK_UFS_ARMHF_IMAGES): debos-recipes/qualcomm-linux-debian-image.yaml rootfs-armhf.tar
+	$(DEBOS_CMD) -t architecture:armhf $<
+
 DISK_SDCARD_ARM64_IMAGES := disk-sdcard-arm64.img \
 	disk-sdcard-arm64.img1 \
 	disk-sdcard-arm64.img2
 
 $(DISK_SDCARD_ARM64_IMAGES): debos-recipes/qualcomm-linux-debian-image.yaml rootfs-arm64.tar
 	$(DEBOS_CMD) -t architecture:arm64 -t imagetype:sdcard $<
+
+DISK_SDCARD_ARMHF_IMAGES := disk-sdcard-armhf.img \
+	disk-sdcard-armhf.img1
+
+$(DISK_SDCARD_ARMHF_IMAGES): debos-recipes/qualcomm-linux-debian-image.yaml rootfs-armhf.tar
+	$(DEBOS_CMD) -t architecture:armhf -t imagetype:sdcard $<
 
 rootfs.tar: rootfs-arm64.tar
 	ln -sf $< $@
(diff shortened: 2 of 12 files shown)


.github/workflows/debos.yml
```diff
@@ -146,6 +153,22 @@
...
+      - name: Build armhf rootfs with debos
+        if: ${{ inputs.build_armhf }}
+        run: |
+          set -ux
+          # NB: no kernelpackages/aptlocalrepo here; the default-ci and EFS
+          # kernels are arm64 only, so armhf sticks to the recipe's default
+          # kernel (Debian's linux-image-armmp)
+          debos \
+              -t architecture:armhf \
+              -t overlays:'${{ inputs.overlays }}' \
```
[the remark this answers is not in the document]
lumag: Fixed

.github/workflows/debos.yml
```diff
@@ -146,6 +153,22 @@
...
+              -t xfcedesktop:true \
+              -t "buildid:${BUILD_ID}" \
+              ${DEBOS_EXTRA_ARGS} \
+              --print-recipe \
+              debos-recipes/qualcomm-linux-debian-rootfs.yaml
+
```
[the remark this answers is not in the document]
lumag: Also fixed

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -1,16 +1,31 @@
...
 {{- $snapshot := or .snapshot "" }}
 
-architecture: arm64
+# only arm64 and armhf are supported architectures
+{{- if and (ne $architecture "arm64") (ne $architecture "armhf") }}
+# Call an unexisting template to err out of the recipe
+{{ template "ERR" }}
+{{- end }}
+
+# suffix appended to artifact names for non-default architectures, so that
```
obbardc: This is overly complex; the usual debos recipe pattern is something like this:

```
{{- $rootfs = printf "rootfs-%s.tar" $architecture }}
{{- $image = printf "image-%s-%s.img" $imagetype $architecture }}
```

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -30,6 +45,8 @@ actions:
     imagesize: {{ $imagesize }}
     partitiontype: gpt
     partitions:
+{{- if ne $architecture "armhf" }}
+      # armhf images don't boot via EFI and carry no ESP
```
obbardc: I think this comment could lead some to think "generic armhf doesn't boot via EFI"; I think it'd be better if you said "On qualcomm platforms..."

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -30,6 +45,8 @@ actions:
     imagesize: {{ $imagesize }}
     partitiontype: gpt
     partitions:
```
obbardc, line 70: This whole section is a bit unreadable due to the inline if statements; usual way I've done this is have separate actions completely:

```
{{- if eq $architecture "arm64" }}
  - action: image-partition
    description: Create an image file
    ...
{{- else if eq $architecture "armhf" }}
  - action: image-partition
    description: Create an image file
    ...
{{- end }}
```

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -138,13 +169,15 @@ actions:
...
+      # generate systemd unit to grow the root partition with above UUID;
+      # the root partition is the first one on armhf (no ESP) and the second
+      # one otherwise
       cat >/etc/systemd/system/debos-grow-rootfs.service <<EOF
       [Unit]
       Description=Grow root partition and filesystem (debos)
       [Service]
       Type=oneshot
-      ExecStartPre=-sh -c 'growpart /dev/\`lsblk -n -o pkname $ROOTFS_DEV\` 2'
+      ExecStartPre=-sh -c 'growpart /dev/\`lsblk -n -o pkname $ROOTFS_DEV\` {{ if eq $architecture "armhf" }}1{{ else }}2{{ end }}'
```
obbardc: This doesn't really fill me with much joy. Could you expose the magic numbers as a variable and set it before the action? Something like:

```
{{- if eq $architecture "arm64" }}
{{ $rootfs_device_index = 2 }}
{{- else if eq $architecture "armhf" }}
{{ $rootfs_device_index = 1 }}
{{- end }}
```

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -1,9 +1,16 @@
+{{- $architecture := or .architecture "arm64" }}
 {{- $xfcedesktop := or .xfcedesktop "false" }}
 {{- $gnomedesktop := or .gnomedesktop "false" }}
 {{- $localdebs := or .localdebs "none" }}
 {{- $aptlocalrepo := or .aptlocalrepo "none" }}
+# default kernel package is architecture-dependent: Debian's arm64 kernel for
```
obbardc: nitpicks: whitespace around this if statement would help for readability; I also think possibly moving the if statement (and `$kernelpackages`) to the end of the variable declaration section would be nicer; IMO it's not great to have if statements amongst global variable declarations :-) .

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -21,7 +28,20 @@
...
-architecture: arm64
+# only arm64 and armhf are supported architectures
+{{- if and (ne $architecture "arm64") (ne $architecture "armhf") }}
+# Call an unexisting template to err out of the recipe
+{{ template "ERR" }}
+{{- end }}
+
+# suffix appended to output artifact names for non-default architectures, so
+# that artifacts for multiple architectures can coexist (e.g. rootfs-armhf.tar)
+{{- $archsuffix := "" }}
```
obbardc: Same comment as the other recipe.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -21,7 +28,20 @@
 {{- $suite = "unstable" }}
 {{- end }}
 
-architecture: arm64
+# only arm64 and armhf are supported architectures
+{{- if and (ne $architecture "arm64") (ne $architecture "armhf") }}
```
obbardc: Same comment as the other recipe.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -1,16 +1,31 @@
+{{- $architecture := or .architecture "arm64" }}
 {{- $dtb := or .dtb "firmware" }}
 {{- $imagesize := or .imagesize "6GiB" }}
 {{- $imagetype := or .imagetype "ufs" }}
-{{- $image := printf "disk-%s.img" $imagetype }}
 {{- $snapshot := or .snapshot "" }}
 
-architecture: arm64
+# only arm64 and armhf are supported architectures
+{{- if and (ne $architecture "arm64") (ne $architecture "armhf") }}
```
obbardc: Consider `{{- if not (has $architecture (list "arm64" "armhf")) }}` (untested)

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -707,12 +729,12 @@ actions:
       tar \
           -C "${latest_kernel}" \
           --transform='s|^\./||' \
-          -cvzf "$ARTIFACTDIR/dtbs.tar.gz" \
+          -cvzf "$ARTIFACTDIR/dtbs{{ $archsuffix }}.tar.gz" \
```
obbardc: consider just using `$architecture` (or having it as a template variable at the top of the recipe); it's clearer to me generally.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -707,12 +729,12 @@ actions:
...
           .
 
   - action: pack
     description: Create root filesystem tarball
-    file: rootfs.tar
+    file: rootfs{{ $archsuffix }}.tar
```
obbardc: Use a `$rootfs` variable like my comment on the rootfs recipe.

Makefile
```diff
@@ -51,25 +51,50 @@ http_proxy ?= $(shell apt-config dump --format '%v%n' Acquire::http::Proxy)
 export http_proxy
 
 .PHONY: all
-all: disk-ufs.img disk-sdcard.img
+all: arm64 armhf
```
obbardc: Do we really want to build armhf images every time ? I see them being quite niche and may be easier to opt-in.

Makefile
```diff
@@ -51,25 +51,50 @@ http_proxy ?= $(shell apt-config dump --format '%v%n' Acquire::http::Proxy)
...
+
+.PHONY: arm64
+arm64: disk-ufs.img disk-sdcard.img
+
+.PHONY: armhf
+armhf: disk-ufs-armhf.img disk-sdcard-armhf.img
 
 rootfs.tar dtbs.tar.gz: debos-recipes/qualcomm-linux-debian-rootfs.yaml
```
obbardc: I think it makes sense to suffix these filenames with the architecture; but that could be a breaking change.

Makefile
```diff
@@ -51,25 +51,50 @@ http_proxy ?= $(shell apt-config dump --format '%v%n' Acquire::http::Proxy)
...
 	$(DEBOS_CMD) $<
 
+# armhf artifacts carry a -armhf suffix (from the architecture template
```
obbardc: I don't think this comment adds much value.

Makefile
```diff
@@ -51,25 +51,50 @@ http_proxy ?= $(shell apt-config dump --format '%v%n' Acquire::http::Proxy)
...
+	$(DEBOS_CMD) -t architecture:armhf $<
+
 DISK_UFS_IMAGES := disk-ufs.img \
 	disk-ufs.img1 \
 	disk-ufs.img2
 
 $(DISK_UFS_IMAGES): debos-recipes/qualcomm-linux-debian-image.yaml rootfs.tar
 	$(DEBOS_CMD) $<
 
+# armhf images have no ESP, so the root filesystem is the sole (first)
```
obbardc: I don't think this comment adds much value either; it's repeating what is the recipe.

README.md
```diff
@@ -104,6 +104,12 @@ To build flashable assets for all supported boards, follow these steps:
     # (optional) if you want SD card images or support for eMMC boards, run
     # this as well:
     make disk-sdcard.img
+
+    # (optional) armhf (32-bit ARM) variants of the above; artifacts are
```
obbardc: Not sure how much value it adds to say the artifacts can coexist.

README.md
```diff
@@ -149,8 +159,9 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
 - `overlays`: a `,`-separated list of rootfs overlays to add from
   `debos-recipes/overlays/`. See the *Supported overlays* section below.
 - `kernelpackages`: a `,`-separated list of kernel packages to install from
-  apt; defaults to `Debian’s linux-image-arm64`. Can (and should) be set to
-  `none` if you are providing local kernel package instead.
+  apt; defaults to Debian’s `linux-image-arm64` on arm64 and to Debian’s
```
obbardc: Use `'` instead of `’` to match the rest of the file.

README.md
```diff
@@ -160,6 +171,13 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
 
 For the image recipe:
 
+- `architecture`: Debian architecture of the image, either `arm64` (the
+  default) or `armhf`; must match the architecture used for the root
+  filesystem recipe. Non-default architectures consume the matching rootfs
+  tarball (e.g. `rootfs-armhf.tar`) and append a `-<architecture>` suffix to
+  the image name (e.g. `disk-ufs-armhf.img`); the root partition GPT type
+  UUID is set according to the architecture. armhf images don't boot via
```
obbardc: again; I think it makes sense to make it clear that it is specific Qualcomm devices that don't boot via EFI.

obbardc (changes requested):
Some thoughts.

obbardc:
Also; I see no tests for these newly-created images. What's your plan for that?

debos-recipes/qualcomm-linux-debian-image.yaml
lumag: Done

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lumag: ack

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lumag: ack

Makefile
lumag: done

Makefile
```diff
@@ -27,8 +27,9 @@ ifeq ($(USE_CONTAINER),auto)
 endif
 
 ifeq ($(USE_CONTAINER),yes)
-	# Only pass --device /dev/kvm if KVM is available on the host
-	KVM_DEVICE := $(if $(wildcard /dev/kvm),--device /dev/kvm)
+	# Only pass --device /dev/kvm if KVM is available on the host; also add
```
obbardc, line 30: I think it'd be a good idea to create a separate PR with this commit (while leaving it in this branch probably); I think it is something very useful which we can merge much sooner than this PR.

.github/workflows/debos.yml
```diff
@@ -104,6 +104,8 @@ jobs:
           )
 
       - name: Build rootfs with debos
+        env:
```
obbardc, line 107: This commit can be dropped; it was fixed in https://github.com/qualcomm-linux/qcom-deb-images/commit/797f2f4b35b0478284f57e5407b2f3d501393a43

README.md
```diff
@@ -89,23 +89,34 @@ To build flashable assets for all supported boards, follow these steps:
...
     # (optional) if you want SD card images or support for eMMC boards, run
     # this as well:
-    make disk-sdcard.img
+    make disk-sdcard-arm64.img
+
+    # both of the above, plus the unsuffixed compatibility symlinks:
+    #make arm64
     ```
 
+    Build artifacts are named after the architecture they were built for
```
obbardc, line 112: Not sure this reads well; how about something like this?

`Build artifacts include the architecture they were built for...`

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -15,6 +15,13 @@
 {{- $rootfs := printf "rootfs-%s.tar" $architecture }}
 {{- $image := printf "disk-%s-%s.img" $imagetype $architecture }}
 
+# On Qualcomm platforms, armhf images have no ESP, so the root filesystem is
```
obbardc, line 18: nitpick: `On Qualcomm platforms, armhf images have no ESP so the root filesystem is`

also; move the `armhf` comment before the `if armhf` statement.


We also need a separate comment to describe the `rootpartnum` variable and why it is set to 2.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -99,6 +127,9 @@ actions:
       apt-get update
 {{- end }}
 
+{{- if ne $architecture "armhf" }}
```
obbardc, line 130: I think I'd rather reverse this check and have it as `if eq arm64` && reverse the comment's intention; but that's a minor nitpick.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -347,11 +355,11 @@ actions:
       # WPA / WPA2 / WPA3 client support
       - wpasupplicant
 
-{{- if eq $suite "trixie" }}
+{{- if and (eq $suite "trixie") (eq $architecture "arm64") }}
 {{- if ne $overlays "none" }}
 {{- range $overlay := split "," $overlays }}
 {{- if eq $overlay "qsc-deb-releases" }}
-  # fastrpc-tests is only available in trixie-overlay
+  # fastrpc-tests is only available in trixie-overlay, and only for arm64
```
obbardc, line 362: nit: `# fastrpc-tests is only available for arm64 in trixie-overlay` possibly

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -424,8 +432,10 @@ actions:
     packages:
       - alsa-utils
       - clinfo
-      # used to debug clock issues
+{{- if eq $architecture "arm64" }}
+      # used to debug clock issues; only built for arm64 in Debian
```
obbardc, line 436: also; we need to discuss this out-of-band; should/can we build debugcc for armhf in debian? Not sure if that's possible. Create a bug report CCing me if so (or remind catch me on slack).

README.md
```diff
@@ -107,15 +107,18 @@ To build flashable assets for all supported boards, follow these steps:
...
 
     Build artifacts are named after the architecture they were built for
-    (`rootfs-arm64.tar`, `disk-ufs-arm64.img`, ...). arm64 was once the only
+    (`rootfs-arm64.tar`, `disk-ufs-armhf.img`, ...). arm64 was once the only
     architecture and its artifacts had no architecture in their name, so
     `make arm64` also creates symlinks under those names (`rootfs.tar`,
     `disk-ufs.img`, ...) for the flashing pipeline and for existing scripts.
     The artifacts published by CI are unaffected: they keep their historical
-    names.
+    names for arm64, and only armhf ones carry a `-armhf` suffix.
```
obbardc, line 121: nit: `names for arm64. Only armhf artifacts have a `-armhf` suffix.`

.github/workflows/debos.yml
```diff
@@ -27,6 +27,13 @@ on:
         description: Skip QEMU tests
         type: boolean
         default: false
+      build_armhf:
```
obbardc, line 30: I wonder if `architecture` may be better variable, comma separated? May need some better shuffling with the recipe calling later in the recipe though; but it seems more future-proof to me.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -661,6 +661,35 @@ actions:
...
+  #   https://bugzilla.kernel.org/show_bug.cgi?id=205957 - the ext4 side
+  # Modules requested by name are unaffected, so name the ones the qemu boot
+  # test needs. Drivers for real hardware (MMC, UFS) are still missing, so
+  # this is a workaround for testing, not a fix.
+  - action: run
+    description: Name the modules the qemu boot test needs (see Debian #1110545)
+    chroot: true
+    command: |
+      set -eux
+      cat >>/etc/initramfs-tools/modules <<'EOF'
```
obbardc, line 682: An overlay of that file may be cleaner than creating the file in a script, not sure though as these recipes seem to avoid overlays.

Makefile
```diff
@@ -122,6 +122,10 @@ test: disk-ufs-arm64.img
 	# rootfs/ is a build artifact, so should not be scanned for tests
 	py.test-3 --ignore=rootfs
 
+.PHONY: test-armhf
+test-armhf: disk-ufs-armhf.img
```
obbardc, line 126: this isn't plumbed into the github actions ? (I didn't actually check if the existing qemu test was; so apologies if not)

Makefile
```diff
@@ -122,6 +122,10 @@ test: disk-ufs-arm64.img
...
+	py.test-3 --ignore=rootfs -k armhf
```
obbardc, line 127: also; are there no boards we can test in LAVA? If not, keep this thread open so we can create an issue about it for future improvements.

obbardc (changes requested):
thanks for fixing the previous review; just added some minor changes/thoughts/nitpicks.

obbardc:
@lumag the debos builds in CI fails as well; but I guess you already got notified about that :-).

debos-recipes/qualcomm-linux-debian-rootfs.yaml
obbardc: debugcc/armhf is now in unstable/testing/trixie-backports; this hunk can be removed now :-)

.github/workflows/debos.yml
lumag: I'd rather not do it. Lists are harder to manage in GH workflows.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lumag: I don't think so. Ideally it should go away if dracut is fixed. Also, this would conflict with the overlays nature of being selectable components.

Makefile
lumag: It is.

Makefile
lumag: No. We discussed adding few ifc6410 boards, but I don't think it ever succeeded.

Makefile
lumag: I mean, the `test-armhf` is not wired (following the current `make test` design), but the armhf images are tested.

PR #514, opened 2026-07-30
ci: revert EnricoMi action to version tag

lool:
The enterprise allowed-actions policy matches
EnricoMi/publish-unit-test-result-action by version tag, so pinning
it to a bare commit SHA (PR #503, commit c07b8d9) made GitHub reject
the workflow at startup:

  The action EnricoMi/publish-unit-test-result-action@d0a4676... is
  not allowed ... all actions must be from a repository owned by
  your enterprise, created by GitHub...

Only build-on-push.yml (push to main) reaches this action via
lava-test.yml, so PR #503's own checks never exercised it and the
startup_failure surfaced only after merge. See run 30561677856.

Restore the @v2 tag and suppress the zizmor/semgrep mutable-tag
findings inline. test-on-pr.yml carries the same pin and would fail the
same way on its next run, so revert it too.

### DIFF
--- a/.github/workflows/lava-test.yml
+++ b/.github/workflows/lava-test.yml
@@ -190,7 +190,12 @@ jobs:
           ls -R $GITHUB_WORKSPACE
 
       - name: Publish Test Results
-        uses: EnricoMi/publish-unit-test-result-action@d0a4676d0e0b938bc201470d88276b7c74c712b3 # v2.24.0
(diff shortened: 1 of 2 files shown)


lool:
I've pinged GitHub admins to add the SHA of this action to our allowed list.

lool:
Self-landed to unbreak CI

PR #516, opened 2026-07-31
ci: fix wifi device name on glymur-crd to wlP4p1s0

lool:
Devices of type glymur-crd enumerate their wifi device as wlP4p1s0, so
we need to set the LAVA test action to use this instead of default
wlan0.

### DIFF
--- a/ci/lava/glymur-crd/boot.yaml
+++ b/ci/lava/glymur-crd/boot.yaml
@@ -70,6 +70,7 @@ actions:
       parameters:
         SKIP_INSTALL: "True"
         WAIT_TIME: 60
+        DEVICE: wlP4p1s0
 context:
   test_character_delay: 10
 device_type: glymur-crd


ci/lava/glymur-crd/boot.yaml
```diff
@@ -70,6 +70,7 @@ actions:
       parameters:
         SKIP_INSTALL: "True"
         WAIT_TIME: 60
+        DEVICE: wlp1s0
```
obbardc: Without all the context, I think the change in this PR is independent of https://github.com/Linaro/test-definitions/pull/631 but if I am wrong, it'd good to link to that change.

obbardc (approved):

gagath (approved):

ci/lava/glymur-crd/boot.yaml
lool: It's more of the same fixes we did locally

Issue #260, opened 2026-02-24
EL2 boot for KVM support


gagath:
Booting in EL2 in necessary in order to use the hardware virtualization capacities of the processor. A dedicated device-tree is available in the upstream kernel for this purpose, but the boot chain is more complicated than when using the regular device-tree.

There seems to be a need to use [slbounce](https://github.com/TravMurav/slbounce) and [qebspil](https://github.com/stephan-gh/qebspil) in order to perform the EL2 boot from UEFI, however there are no known other public documentation on how to achieve EL2 boot on Hamoa systems. This should be documented somewhere for users who want to achieve KVM virtualization on qcom hardware.

lool:
Just bumped onto https://github.com/qualcomm-linux/meta-qcom/pull/1631 where EL2 is being enabled on some IoT platforms

obbardc:
I'm happy to package slbounce and friends if needed, but I've never managed to get them working (due to not trying yet), and as you say it could go away at any time from a bios update

PR #528, opened 2026-08-03
feat(debos/rootfs): add default user to kvm group

lool:
/dev/kvm is owned root:kvm with mode 0660, so the default "debian"
user could not use KVM acceleration without root on deployed images.
Add the user to the kvm group, and pre-emptively create the group in
case udev - which normally ships the sysusers config creating it -
is absent, mirroring the existing sudo group handling.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -368,10 +368,12 @@ actions:
       set -eux
       # pre-emptively create a sudo group if sudo isn't installed
       getent group sudo >/dev/null 2>&1 || groupadd --system sudo
+      # pre-emptively create a kvm group if udev isn't installed
+      getent group kvm >/dev/null 2>&1 || groupadd --system kvm
       # some useful groups for desktop scenarios, but also to run payloads
       # from the serial console, over SSH, or in containers - where the desktop
       # session has not updated ACLs to the device nodes
-      groups="adm,audio,render,sudo,users,video"
+      groups="adm,audio,kvm,render,sudo,users,video"
       # fastrpc group only exists when fastrpc-support is installed (Debian)
       if getent group fastrpc >/dev/null 2>&1; then
         groups="${groups},fastrpc"


obbardc (approved):
Looks fine, did you KVM works on hardware now or just assumed to work ? I am fine with the latter for a change.

lool:
> Looks fine, did you KVM works on hardware now or just assumed to work ? I am fine with the latter for a change.

it might work in some cases, so it gives a way to use it when it does

obbardc:
> > Looks fine, did you KVM works on hardware now or just assumed to work ? I am fine with the latter for a change.
> 
> it might work in some cases, so it gives a way to use it when it does

Sorry, I reread my original message and it didn't make any sense. I was supposed to ask if you validated the change actually works and allows the user to use kvm (of course on hardware where it is supported, ha ha) :-)

lool:
I did not validate this yet, I guess we probably need some useful kvm capable userspace; I'm not sure if normal tooling is acceptable or not for our default image

obbardc:
> I did not validate this yet

~~Just an offer, I can test it with qemu locally on RB3gen2 if you'd like? I have some cycles tomorrow.~~

Skip that; it runs gunyah so no KVM for me!

> I guess we probably need some useful kvm capable userspace; I'm not sure if normal tooling is acceptable or not for our default image

I don't imagine it's worth creating a separate virtualisation image variant or installing qemu etc in the default images in the short term unless there is some use-case I don't yet know about :D.

lool:
> ~Just an offer, I can test it with qemu locally on RB3gen2 if you'd like? I have some cycles tomorrow.~
> 
> Skip that; it runs gunyah so no KVM for me!

Latest boot firmware was supposed to address that!

> I don't imagine it's worth creating a separate virtualisation image variant or installing qemu etc in the default images in the short term unless there is some use-case I don't yet know about :D.

right, so how would we validate /dev/kvm works? I guess we need something like qemu-system to run a test VM

For end-developers, I guess everyone has their taste on what a good hypervisor should be

obbardc:
> > ~Just an offer, I can test it with qemu locally on RB3gen2 if you'd like? I have some cycles tomorrow.~
> > Skip that; it runs gunyah so no KVM for me!
> 
> Latest boot firmware was supposed to address that!

how curious; let's eventually create an issue about that once I have tested some more, I am using the latest boot firmware and qcom-next kernel...

My offer is still there to test this if I can get the rb3gen2 working :-).


> > I don't imagine it's worth creating a separate virtualisation image variant or installing qemu etc in the default images in the short term unless there is some use-case I don't yet know about :D.
> 
> right, so how would we validate /dev/kvm works? I guess we need something like qemu-system to run a test VM
> For end-developers, I guess everyone has their taste on what a good hypervisor should be

yeah, I think it'd have to be a manual install from `apt` for validation for now, soon we can think about what the virtualisation stack should look like and if we want to install it by default/have a separate variant (I have taken a note about this).

lool:
I checked on a rb3gen2 with latest boot firmware + latest qcom-next, and with the default assets it's not getting kvm support:
```
debian@debian:~$ sudo dmesg |grep kvm
[    0.138891] kvm [1]: HYP mode not available
```
so we probably need to flip to another set of boot assets to confirm

obbardc:
> I checked on a rb3gen2 with latest boot firmware + latest qcom-next, and with the default assets it's not getting kvm support:
> 
> ```
> debian@debian:~$ sudo dmesg |grep kvm
> [    0.138891] kvm [1]: HYP mode not available
> ```
> 
> so we probably need to flip to another set of boot assets to confirm

FYI I got the same error message.

Without knowing much of the recent backstory, I checked meta-qcom: I wonder if Linux KVM support requires the U-Boot flow and https://github.com/qualcomm-linux/meta-qcom/blob/master/recipes-bsp/u-boot/files/gunyah-exit.cfg ?

Perhaps it makes sense to track this in an issue instead?

lool:
Sure, an issue would be fine; I've flipped this PR back to draft until we find one way on rb3gen2/one platform to test kvm on, and some relevant userspace test.

obbardc:
FYI https://github.com/qualcomm-linux/qcom-deb-images/issues/260 may align to this PR.

PR #533, opened 2026-08-06
fix(rootfs): precreate fastrpc group

lool:
The fastrpc group is no longer created when fastrpc-support is
installed; it is now created at boot time from the systemd
sysusers.d fragment shipped by the package. As a result, the
build-time "getent group fastrpc" check failed and the debian user
was never added to the fastrpc group.

Precreate the group at build time by running systemd-sysusers on
/usr/lib/sysusers.d/fastrpc.conf (guarded on its presence) before
creating the debian user, so the existing group-membership logic
works again.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -361,6 +361,19 @@ actions:
       # /etc/hosts is created by netbase
       sed -i "1s/^/127.0.1.1	debian\n/" /etc/hosts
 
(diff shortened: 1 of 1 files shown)


obbardc (approved):

basak-qcom (approved):
I agree that this is the most appropriate fix and it looks good. Since we don't have tests, please validate the image manually before merging.

lool:
This change worked, I was able to boot the trixie image in qemu and the debian user was in the fastrpc group.

I failed to test it on a real device, and discovered other issues with our forky images which I'll send a separate PR for.

PR #535, opened 2026-08-06
fix(debos/rootfs): add fastrpc-tests unconditionally

lool:
fastrpc-tests is no longer limited to the trixie qsc-deb-releases
overlay; instead it's available from QLI trixie and forky
repositories which are enabled unconditionally.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -361,20 +361,14 @@ actions:
       # WPA / WPA2 / WPA3 client support
       - wpasupplicant
 
-{{- if eq $suite "trixie" }}
-{{- if ne $overlays "none" }}
-{{- range $overlay := split "," $overlays }}
-{{- if eq $overlay "qsc-deb-releases" }}
-  # fastrpc-tests is only available in trixie-overlay
+{{- if ne $suite "unstable" }}
+  # fastrpc-tests is available from QLI overlays for trixie and forky
   - action: apt
-    description: Install fastrpc-tests (trixie only)
+    description: Install fastrpc-tests from QLI overlay
     recommends: true
     packages:
-      # fastrpc support. fastrpc-tests pulls in the required stack
+      # fastrpc-tests pulls in fastrpc-support
       - fastrpc-tests
-{{- end }}
-{{- end }}
-{{- end }}
 {{- end }}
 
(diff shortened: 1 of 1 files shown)


debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -361,21 +361,13 @@ actions:
       # WPA / WPA2 / WPA3 client support
       - wpasupplicant
 
-{{- if eq $suite "trixie" }}
-{{- if ne $overlays "none" }}
-{{- range $overlay := split "," $overlays }}
-{{- if eq $overlay "qsc-deb-releases" }}
-  # fastrpc-tests is only available in trixie-overlay
+  # fastrpc-tests is available from QLI overlays for trixie and forky
```
obbardc, line 365: I think this needs to be gated behind `{{- if ne $suite "unstable" }}` to match the action which sets up the QLI overlay (see the action with description `Create Qualcomm Linux APT sources`).

obbardc (changes requested):
Just one small thought to match the other QLI instance.

Does this mean we are slowly getting away from the QArtifactory deb repo ?! :D 


BTW, in https://github.com/qualcomm-linux/qcom-deb-images/pull/504 I have a commit which disables the QLI overlay when building vanilla Debian. I'll rebase that on top of this PR once it's merged.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: agreed and good catch – fixed!

a larger problem I didn't discuss before sending this is that we should be installing fastrpc-support when fastrpc-tests is not available, but then I realized that it wouldn't be available for a vanilla Debian build that is forky or higher. And that we didn't really have a vanilla Debian build. Hence https://github.com/qualcomm-linux/qcom-deb-images/issues/536 and my recent fastrpc backport to trixie. :)

obbardc: yeah; i will take a look at the vanilla debian build.

obbardc: tracked your specific fastrpc concern in https://github.com/qualcomm-linux/qcom-deb-images/issues/536#issuecomment-5209772361

obbardc (approved):

PR #538, opened 2026-08-07
feat(ci/lava): add monaco-arduino-monza

### LINKED ISSUE #537: Monza CI support
Now that a couple of monza (VENTUNO Q) boards have been added to the production LAVA instance, we should use them in CI.

lool:
Add LAVA templates for monaco-arduino-monza, identical to
qrb2210-rb1 which is also using eMMC storage, but setting the proper
expected wifi device name.

As soon as we add a CI template, it gets picked up by default unless
the workflow uses an explicit include list, so verify that this
works with various kernels.

Tested with:
- arduino (PASS) https://lava.infra.foundries.io/scheduler/job/353015
- mainline (PASS) https://lava.infra.foundries.io/scheduler/job/353161
- qcom-next (FAIL) https://lava.infra.foundries.io/scheduler/job/353160

Update excludes for Debian builds and qcom-next; don't exclude for
mainline/linux-next; arduino daily was pre-emptively including this
platform too.

Fixes: #537

### DIFF
--- a/.github/workflows/build-on-push.yml
+++ b/.github/workflows/build-on-push.yml
@@ -38,3 +38,5 @@ jobs:
       LAVATOKEN: ${{ secrets.LAVATOKEN }} # lava-test.yml
     with:
       url: ${{ needs.build-daily.outputs.artifacts_url }}
+      # https://github.com/qualcomm-linux/qcom-deb-images/issues/537
+      boards_exclude: '["monaco-arduino-monza"]'
--- a/ci/lava/monaco-arduino-monza/boot.yaml
+++ b/ci/lava/monaco-arduino-monza/boot.yaml
@@ -1,0 +1,74 @@
+actions:
+- deploy:
+    rootfs_image: disk-sdcard.img2
+    overlay_path: /
+    qcomflash:
+      headers:
+        Authorization: Q_S3_TOKEN
+      url: "{{BUILD_DOWNLOAD_URL}}/{{SUITE}}-flash-emmc.tar.gz"
+      apply-overlay: true
+    timeout:
+      minutes: 20
+    to: qdl
+- boot:
+    firehose_program: prog_firehose_ddr.elf
+    rawprogram: rawprogram0.xml
+    patch: patch0.xml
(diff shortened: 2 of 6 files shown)


lool:
rebased after latest CI changes

lool:
Ideally, we'd have a single map of which boards are expected to work where so that we can refer to it from multiple workflows, notably for the 3 qcom-next bases ones.

.github/workflows/build-on-push.yml
```diff
@@ -38,3 +38,5 @@ jobs:
       LAVATOKEN: ${{ secrets.LAVATOKEN }} # lava-test.yml
     with:
       url: ${{ needs.build-daily.outputs.artifacts_url }}
+      # https://github.com/qualcomm-linux/qcom-deb-images/issues/537
+      boards_exclude: '["monaco-arduino-monza"]'
```
obbardc, line 42: Should we also exclude this board in the Linux-next and linux-mainlinr workflows?

obbardc (approved):

gagath (approved):

.github/workflows/build-on-push.yml
lool: No I think most boards should be listed on these to get the state of upstream somewhere

ci/lava/monaco-arduino-monza/boot.yaml
```diff
@@ -0,0 +1,74 @@
+actions:
+- deploy:
+    rootfs_image: disk-sdcard.img2
+    overlay_path: /
+    qcomflash:
+      headers:
+        Authorization: Q_S3_TOKEN
```
mwasilew, line 7: IT asked to move to Q_QLI_S3_TOKEN. I'll send a new patch to fix that.

mwasilew (commented):
I'm a bit late as the patch was merged. I'll send a fix today.

ci/lava/monaco-arduino-monza/boot.yaml
obbardc: That's used in all of the CI templates, I guess a PR to fix them all would be good.

PR #546, opened 2026-08-10
chore(debos): prefer wireplumber from backports

rchiluka29:
WirePlumber 0.5.12 is available in trixie-backports and is needed for AudioReach validation on Glymur. Add the wireplumber source package to the existing backports preference list so wireplumber and libwireplumber-0.5-0 are selected together during image builds.

PipeWire is intentionally not added here. Glymur testing showed QCOM PipeWire 1.4.2-1~qcom1 with WirePlumber 0.5.12 keeps playback working, while Debian backports PipeWire 1.4.9 regressed playback. WirePlumber 0.5.12 only requires pipewire >= 1.0.2, which the QCOM PipeWire package satisfies.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -156,7 +156,8 @@ actions:
       "src:linux:any"
       "src:linux-signed-arm64:any"
       "src:mesa:any"
-      "src:u-boot-efi-dtb" }}
+      "src:u-boot-efi-dtb"
+      "src:wireplumber:any" }}
 
   - action: run
     description: Add Debian {{ $suite }}-backports APT configuration


lool:
Hi, can you explain how it regressed playback?

rchiluka29:
> Hi, can you explain how it regressed playback?

Edited:
 I re-tested this on Glymur and the earlier PipeWire playback concern is not reproducing now.

  Basic validation passed with WirePlumber 0.5.12 using:
  - QCOM PipeWire 1.4.2-1~qcom1
  - Debian PipeWire 1.4.2-1
  - Debian backports PipeWire 1.4.9-1~bpo13+2

  I updated the PR to include PipeWire in the backports source-package preference as well. With the current qsc-deb-releases pin, the QCOM PipeWire package still wins. If that changes later, the Debian PipeWire paths above have also been sanity-tested.

lool:
Hi @rchiluka29 can you more specifically describe which feature is present in the wireplumber in trixie-backports / which issue the trixie version has so that we can connect with the functional requirements?

Your PR currently still includes pipewire from backports.

I'm also worried about the issue you mention with pipewire from backports; in the case we would need that backport in the future, could you describe the issue/symptoms?

Thanks

rchiluka29:
> Hi rchiluka29 can you more specifically describe which feature is present in the wireplumber in trixie-backports / which issue the trixie version has so that we can connect with the functional requirements?
> 
> Your PR currently still includes pipewire from backports.
> 
> I'm also worried about the issue you mention with pipewire from backports; in the case we would need that backport in the future, could you describe the issue/symptoms?
> 
> Thanks


The functional issue is with WirePlumber startup on Glymur AudioReach.  With Debian trixie WirePlumber 0.5.8, we observed a boot-time failure while loading the AudioReach 90-device-detection.conf. 
The failure is in the component/config loading path: required components may be loaded in parallel instead of waiting for their dependencies, which can make WirePlumber fail during startup.

WirePlumber 0.5.12 from trixie-backports avoids this issue with 0.5.12, PAL sinks/sources enumerate and basic speaker playback / handset-mic record validation passes.

The PipeWire concern I mentioned earlier is not reproducible now. To keep this PR focused, I can drop the PipeWire change and keep only the WirePlumber backport preference.

lool:
If you don't need newer pipewire right now or in the near future, let's drop it from this PR, thanks

rchiluka29:
> If you don't need newer pipewire right now or in the near future, let's drop it from this PR, thanks

Done, I dropped the PipeWire change and kept this PR focused on WirePlumber only.

obbardc (approved):

PR #550, opened 2026-08-10
feat(debos/flash): reduce multi-DTB FIT image size to 2 MiB

neervish:
## Summary
dtb-multidtb-<soc>.bin was always built at build-dtb-image.sh's 4 MiB default.
glymur-crd's ptool partitions (dtb_a/dtb_b) are being reduced from 4 MiB to 2 MiB, so glymur's image needs to shrink to match.

Rather than special-casing glymur, reduce the default to 2 MiB for every SoC that generates a multidtb FIT image: actual FIT content for all current SoCs (glymur, hamoa, purwa, shikracqm, shikracqs, shikraiqs) is well under 2 MiB, and ptool partitions that stay at 4 MiB or larger are unaffected, since flashing a smaller image into a larger partition is harmless.

## Test
Built with these changes; confirmed `dtb-multidtb-<soc>.bin` is now 2.0 MiB for the SoCs (glymur, hamoa, purwa, shikracqm, shikracqs, shikraiqs)

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -534,6 +534,7 @@ actions:
               if "${QCOM_DTB_METADATA}/build-dtb-image.sh" \
                   --dtb-src build/dtbs/qcom \
                   --soc {{ $soc }} \
+                  --size 2 \
                   --out "${dtb_bin}" \
                   --prune; then
                   echo "Multi-DTB FIT image generated: ${dtb_bin}"


lool (commented):
Hi @neervish, thanks for the PR

Would all platform fit comfortably within 2MiB? If so, I'd prefer reducing to 2MiB across the board rather than introducing a parameter and one more difference between families.

neervish:
> Hi Neeraj (@neervish), thanks for the PR
> 
> Would all platform fit comfortably within 2MiB? If so, I'd prefer reducing to 2MiB across the board rather than introducing a parameter and one more difference between families.

Hi @lool, Thanks for the feedback!

Since dtb-multidtb-<soc>.bin is generated once per soc_id (not per board/platform), I read "reducing to 2MiB across the board" as reducing it for every SoC that currently uses multidtb generation -> **Please confirm** 

Also checked the actual FIT content size for all 6 SoCs currently using multidtb

SoC | FIT Image Size
-- | --
Glymur | ~278   KB
Hamoa | ~310   KB
Purwa | ~457   KB
Shikra (cqm) | ~121   KB
Shikra (cqs) | ~121   KB
Shikra (iqs) | ~119   KB

All well within 2 MiB, Even if we add more variants and respective overlays it should be less than 2MiB.

JFYI The immediate need is glymur-crd, whose ptool partitions (dtb_a/dtb_b) are being reduced from 4 MiB → 2 MiB.
In ptool dtb partitions of hamoa/purwa are at 4096KB and shikra at 65536KB with no planned reduction as of now.
Their content fits in 2 MiB either way, so happy to hardcode --size 2 for all of them if that's what you'd prefer — just want to confirm before dropping the current SoC wise dtb size changes, since only glymur currently has a confirmed reason to shrink.

lool:
> > Hi Neeraj (Neeraj (@neervish)), thanks for the PR
> > Would all platform fit comfortably within 2MiB? If so, I'd prefer reducing to 2MiB across the board rather than introducing a parameter and one more difference between families.
> 
> Hi Loïc Minier (@lool), Thanks for the feedback!
> 
> Since dtb-multidtb-.bin is generated once per soc_id (not per board/platform), I read "reducing to 2MiB across the board" as reducing it for every SoC that currently uses multidtb generation -> **Please confirm**

Yes! :)

I do hope most SoCs will eventually move to multidtb, ideally we'd get rid of other implementations and align on one, so it would be nice to estimate that too, but that's low prio.

> Also checked the actual FIT content size for all 6 SoCs currently using multidtb

Thanks!

> All well within 2 MiB, Even if we add more variants and respective overlays it should be less than 2MiB.

Awesome, let's do it, let's switch our FIT payload to 2MiB, even if the part size is actually 4MiB on some platforms. Flashing will zero out the rest or leave it intact, I don't think we care. No need to patch part size IMO.

lool (changes requested):
let's do 2MiB for all FIT images

neervish:
> let's do 2MiB for all FIT images

Done — updated the PR, now using a flat --size 2 for all multidtb SoCs.

lool (approved):
Approved if CI passes

PR #554, opened 2026-08-12
feat(qcom-next): update to tag qcom-next-7.2-rc5-20260812

rnaramre:
Move to the latest linux-qcom-next release tag qcom-next-7.2-rc5-20260812, which corresponds to kernel version v7.2-rc5.


lool (approved):
Tag update looks good; I passed, but we don't really try building the new kernel, so we need to look at tomorrow CI, until we switch this to pkg-linux-qcom

PR #558, opened 2026-08-12
ci: allow manual runs of the workflows from any branch

obbardc:
The daily and weekly workflows have a `workflow_dispatch` trigger, so can be manually ran, but have a guard on the main branch; so they cannot be manually ran on development branches. This means changes to the workflows cannot be tested before they are merged: currently, the only way to find out whether a workflow change works is to merge it and wait for the next cron run.

Remove the branch guard when the workflows are manually ran from the `workflow_dispatch` trigger to allow branches pushed to the upstream repo to be tested.

Also document how to run the workflows manually in CONTRIBUTING.md.

---

To verify this, I ran:
```
$ gh workflow run linux-daily-qcom-next.yml --ref wip/obbardc/dispatch-workflows-from-branch
✓ Created workflow_dispatch event for linux-daily-qcom-next.yml at wip/obbardc/dispatch-workflows-from-branch
```
which created [this workflow](https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/31603984133). I'd like to wait for this workflow to complete before merging.

### DIFF
--- a/.github/workflows/build-daily-debian.yml
+++ b/.github/workflows/build-daily-debian.yml
@@ -20,8 +20,13 @@ permissions: {}
 
 jobs:
   build-daily-debian:
-    # don't run cron from forks of the main repository or from other branches
-    if: github.repository == 'qualcomm-linux/qcom-deb-images' && github.ref == 'refs/heads/main'
+    # don't run cron from forks of the main repository or from other branches;
+    # manual runs are allowed from any branch of the main repository so
+    # that changes to the workflows can be tested before they are merged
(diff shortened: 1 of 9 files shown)


lool (approved):
the GH workflow guard is ugly, but I can't think of anything you can do abou tit

obbardc:
> the GH workflow guard is ugly

@lool yeah, I did think about reversing the check so we only check repo/branch on the trigger, but it didn't match the rest of the repo where we prefer to deny-first.

PR #562, opened 2026-08-12
ci: drop duplicated workflow frequency

obbardc:
The scheduled workflows carry the job cadence in both the filenames
and workflow/job names, for instance: `linux-weekly-mainline.yml` and
`Weekly mainline linux build`. The schedule is set by the cron
expression in the workflow, so keeping the schedule information in
multiple places is just repeating ourselves.

Instead, rename the workflow files after what they build/test and
remove the duplicated schedule information across the repository.


lool (approved):
This looks correct, and I agree with the spirit of the changes

PR #575, opened 2026-08-14
ci: only publish EFS latest symlinks when building main branch

obbardc:
Both the kernel and the U-Boot workflows allow manual runs with `workflow_dispatch` from any branch of the main repository, so that workflow changes can be tested before they are merged. Those runs were also updating the EFS symlinks `/efs/${KERNEL_NAME}` and `/efs/u-boot-rb1-latest` to their build, which every consumer workflow resolves to pick up the latest artifacts.

A WIP branch build therefore can silently repoint the latest artifacts at its own artifacts, which could cause changes which are not ready to leak into stable image builds.

Move the symlink update into its own step guarded on the main branch. Branch builds still publish to /efs/${BUILD_ID} and still upload private artifacts, they just no longer update the latest symlinks.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -168,7 +168,7 @@ jobs:
           export BUILD_DIR="/efs/${BUILD_ID}"
           mkdir -vp "${BUILD_DIR}"
           cp -av artifacts/* "${BUILD_DIR}"
-          # perhaps help NFS sync
+          # flush artifacts to NFS
           sync
 
       # only publish EFS symlink to the latest build artifacts if building the
@@ -179,7 +179,7 @@ jobs:
           set -ux
           # create or update symlink
           ln -fnsv "${BUILD_ID}" "/efs/${KERNEL_NAME}"
-          # perhaps help NFS sync
+          # flush the symlink to NFS
           sync
 
       - name: Upload private artifacts
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -168,6 +168,15 @@ jobs:
           export BUILD_DIR="/efs/${BUILD_ID}"
           mkdir -vp "${BUILD_DIR}"
           cp -av artifacts/* "${BUILD_DIR}"
+          # perhaps help NFS sync
+          sync
+
+      # only publish EFS symlink to the latest build artifacts if building the
+      # main branch
+      - name: Publish EFS symlink
+        if: github.ref == 'refs/heads/main'
+        run: |
+          set -ux
           # create or update symlink
           ln -fnsv "${BUILD_ID}" "/efs/${KERNEL_NAME}"
           # perhaps help NFS sync
--- a/.github/workflows/u-boot.yml
+++ b/.github/workflows/u-boot.yml
@@ -83,7 +83,7 @@ jobs:
           export BUILD_DIR="/efs/${BUILD_ID}"
           mkdir -vp "${BUILD_DIR}"
           cp -av artifacts/* "${BUILD_DIR}"
-          # perhaps help NFS sync
+          # flush artifacts to NFS
           sync
 
       # only publish EFS symlink to the latest build artifacts if building the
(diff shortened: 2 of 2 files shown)


.github/workflows/linux.yml
```diff
@@ -168,6 +168,15 @@ jobs:
           export BUILD_DIR="/efs/${BUILD_ID}"
           mkdir -vp "${BUILD_DIR}"
           cp -av artifacts/* "${BUILD_DIR}"
+          # perhaps help NFS sync
+          sync
```
basak-qcom, line 172: What's the purpose of this change, please? I don't follow how it relates to the commit message.

.github/workflows/u-boot.yml
```diff
@@ -83,6 +83,15 @@ jobs:
           export BUILD_DIR="/efs/${BUILD_ID}"
           mkdir -vp "${BUILD_DIR}"
           cp -av artifacts/* "${BUILD_DIR}"
+          # perhaps help NFS sync
+          sync
```
basak-qcom, line 87: Same question as above.

basak-qcom (commented):
lgtm, one question inline

.github/workflows/linux.yml
obbardc: the sync call is already there in the `Upload results to EFS` action; I just duplicated it for my addition `Publish EFS symlink` and `git diff` makes that a bit confusing :-).

I will try to make that that a bit more clear in my commit message though.

.github/workflows/u-boot.yml
obbardc: for record, same answer as above, the sync call is already there in the Upload results to EFS action; I just duplicated it for my addition Publish EFS symlink and git diff makes that a bit confusing :-).

I will try to make that that a bit more clear in my commit message though.

.github/workflows/linux.yml
obbardc: Done and added another commit to reword the comment.

.github/workflows/u-boot.yml
obbardc: Done and added another commit to reword the comment.

lool (commented):
I wonder if we have a bigger problem here and we should somehow only give some EFS write secret to some workflows that only run from our main branch, and only read secrets elsewhere. But we don't really have such separation.

Can't I trivially revert that in a PR?

One thing that bugs me is testing for branch "main", without also testing for the org + repo name; the "main" branch is only meaningful in the context of our repo?

obbardc:
> I wonder if we have a bigger problem here and we should somehow only give some EFS write secret to some workflows that only run from our main branch, and only read secrets elsewhere. But we don't really have such separation.
> 
> Can't I trivially revert that in a PR?
> 
> One thing that bugs me is testing for branch "main", without also testing for the org + repo name; the "main" branch is only meaningful in the context of our repo?

@lool one thing that has just occurred to me, once we have #593 and #595 we no longer need to have `latest` symlinks for any .deb artifacts anymore... Currently the `latest` symlink is created as we have separate kernel/build workflows and we reuse the nightly kernels (and latest U-boot image for RB1) in each image build.

After those PRs are merged, the bindeb kernel workflow & resultant image workflow will then be married together (e.g. for manual debugging) and the image workflow can simply use the full efs artifacts path the kernel pushed to (e.g. with buildid). Then I think it makes sense to just drop the `latest` symlinks...


But that doesn't answer your question about the read/write tokens though.

lool:
> > I wonder if we have a bigger problem here and we should somehow only give some EFS write secret to some workflows that only run from our main branch, and only read secrets elsewhere. But we don't really have such separation.
> > Can't I trivially revert that in a PR?
> > One thing that bugs me is testing for branch "main", without also testing for the org + repo name; the "main" branch is only meaningful in the context of our repo?
> 
> Loïc Minier (@lool) one thing that has just occurred to me, once we have #593 and #595 we no longer need to have `latest` symlinks for any .deb artifacts anymore... Currently the `latest` symlink is created as we have separate kernel/build workflows and we reuse the nightly kernels (and latest U-boot image for RB1) in each image build.
> 
> After those PRs are merged, the bindeb kernel workflow & resultant image workflow will then be married together (e.g. for manual debugging) and the image workflow can simply use the full efs artifacts path the kernel pushed to (e.g. with buildid). Then I think it makes sense to just drop the `latest` symlinks...
> 
> But that doesn't answer your question about the read/write tokens though.

Agreed, good thinking, we might even be able to get rid of EFS entirely if we can download the S3 artifacts

PR #578, opened 2026-08-14
ci: build & test images using an APT snapshot

### LINKED ISSUE #564: Add CI coverage for snapshot functionality
Snapshot functionality is currently not covered by CI in this repository.

There are now several code paths and branching points related to snapshots, currently completly untested. We should add CI coverage for snapshots.

obbardc:
Snapshot builds pin the APT sources to a dated archive (see
docs/snapshot.md). They have their own code paths in the rootfs and
image recipes which nothing else exercises, so build an image with them
regularly, in a workflow of their own.

It runs weekly and on demand rather than daily or on pull requests:
these are full image builds and snapshot.debian.org is slower and more
rate-limited than the regular mirrors. Keeping it out of build.yml is
what lets it have that cadence and gives its artifacts a destination of
their own and doesn't pollute the nightly builds with additional build jobs or artifacts.

This only ever answers "do the snapshot code paths still work?": no LAVA
job boots the images, the build going green and the QEMU tests ran by
debos.yml are the coverage we're after. The artifacts are uploaded all
the same, so that a failure can be investigated. The default image variant is
enough, as the snapshot code paths don't depend on it.

The snapshot timestamp is derived from the committer date of the commit being
built. Any well-formed timestamp resolves to the publication which was live at that
moment, so the value only has to be a date the archives still serve, not
one which coincides with a publication. The commit date is such a date
and needs no maintenance: it moves forward on its own as the repository
is worked on, while two runs of the same commit still pin the same
publication, so a failure points at a change in our recipes rather than
at a change in the archives.

Fixes: #564

## TODO

- Should we remove `flash` build as it is a waste of time/space?
- Remove dummy testing commits
- can't download image in `test-snapshot` job:
```
+ curl -fL --retry 3 https://qli-prod-artifacts.qualcomm.com/qcom-prd-gh-artifacts/qualcomm-linux/qcom-deb-images/*/forky-disk-ufs.img.gz
+ gunzip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 403
```
- CI check to check that buildinfo contains the expected SNAPSHOT= line
- CI check to verify that we've correctly cleaned up so apt-snapshot-toggle is removed (see https://github.com/qualcomm-linux/qcom-deb-images/issues/582) and snapshots are disabled on all sources.list and sources.list.d entries.
- snapshot job: upload `snapshot.txt` into artifacts
- snapshot tests: instead of separate test file `qemu_snapshot_tests.py`, add tests into `qemu_test.py` and only pass vars in debos job. this will allow us to run snapshot tests on non-snapshot enabled image too, and remove `extra_tests` from debos.yml. also, rename `test_env`. 

## Verification

TODO: See [sample build-snapshot.yml workflow run]()

### DIFF
--- a/.github/workflows/build-snapshot.yml
+++ b/.github/workflows/build-snapshot.yml
@@ -24,6 +24,8 @@ on:
           date of the commit being built.
         type: string
         default: ''
+  # WIP: Temporarily allow on branch pushes.
+  push:
 
 # grant nothing by default; every job below declares the scopes it needs
 permissions: {}
@@ -50,10 +52,10 @@ jobs:
     # don't run cron from forks of the main repository or from other branches;
     # manual runs are allowed from any branch of the main repository so
     # that changes to the workflows can be tested before they are merged
-    if: >-
-      github.repository == 'qualcomm-linux/qcom-deb-images' &&
-      (github.event_name == 'workflow_dispatch' ||
-      github.ref == 'refs/heads/main')
+    #if: >-
+    #  github.repository == 'qualcomm-linux/qcom-deb-images' &&
+    #  (github.event_name == 'workflow_dispatch' ||
+    #  github.ref == 'refs/heads/main')
     name: Determine the snapshot timestamp
     runs-on: ubuntu-latest
     permissions:
@@ -100,10 +102,10 @@ jobs:
     # don't run cron from forks of the main repository or from other branches;
     # manual runs are allowed from any branch of the main repository so
     # that changes to the workflows can be tested before they are merged
-    if: >-
-      github.repository == 'qualcomm-linux/qcom-deb-images' &&
-      (github.event_name == 'workflow_dispatch' ||
-      github.ref == 'refs/heads/main')
+    #if: >-
+    #  github.repository == 'qualcomm-linux/qcom-deb-images' &&
+    #  (github.event_name == 'workflow_dispatch' ||
+    #  github.ref == 'refs/heads/main')
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/build-snapshot.yml
+++ b/.github/workflows/build-snapshot.yml
@@ -123,3 +123,9 @@ jobs:
       kernelpackages: efs/qcom-next
       # the default variant is enough: the snapshot code paths don't depend on
       # which variant is built
+      #
+      # the QEMU tests debos.yml runs are the same for every image; telling
+      # them the timestamp this build was pinned to is what turns their
+      # snapshot checks from "no snapshot was used" into "this exact snapshot
+      # was". they run in the build job, against the image still on that runner
+      qemu_test_env: EXPECTED_SNAPSHOT=${{ needs.snapshot-date.outputs.snapshot }}
--- a/.github/workflows/build-snapshot.yml
+++ b/.github/workflows/build-snapshot.yml
@@ -1,0 +1,125 @@
+# Builds QLI Debian images from a dated APT snapshot with APT sources pinned to
+# snapshot.debian.org and the Qualcomm Linux archive (see docs/snapshot.md).
+#
+# The snapshot timestamp is derived from the HEAD commit's timestamp and can be
+# manually set when manually triggering the workflow.
+
+name: Build snapshot
+
+on:
+  # Weekly rather than daily and never on pull requests. These are full image
+  # builds and snapshot.debian.org is slower and more rate-limited than the
+  # regular mirrors. Run the job on Saturday when the daily builds are not
+  # competing for the builder runners.
+  schedule:
+    - cron: '30 3 * * 6'
+  # Allow manual runs; optionally accept a snapshot timestamp to build against
+  # rather than deriving it from the last commit timestamp.
+  workflow_dispatch:
+    inputs:
+      snapshot:
+        description: >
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -54,6 +54,52 @@ jobs:
...
+      suite: ${{ matrix.suite }}
+      # both the rootfs and the image recipe re-pin their own APT sources, and
+      # debos_extra_args is passed to both.
+      #
+      # any well-formed timestamp resolves to the publication which was live at
+      # that moment, so this is a fixed date rather than something computed: it
+      # keeps the build reproducible and makes a failure a change in our
+      # recipes rather than a change in the archives. bump it when it gets old
+      # enough that the archives no longer serve it well.
+      debos_extra_args: -t snapshot:20260810T082104Z
```
basak-qcom: I agree with keeping the build reproducible. But consider this scenario. I file a PR that picks up a new package. The checks against that PR passes, and the PR is merged. Then the next day this job fails, because the snapshot is too old. Someone will then have to context switch back to bump the timestamp.

To avoid this, how about using the timestamp of the latest commit instead? This would nicely avoid the above scenario I think. The build would still be reproducible. It would be required that the package is available at the snapshot of the commit. This may not always be strictly true, for example if I prepare something in advance of publishing something and then land the PR after. But hopefully that's good enough and an improvement over the above scenario most of the time.

Is there a scenario I haven't through of that would break with this change?

.github/workflows/build.yml
```diff
@@ -54,6 +54,52 @@ jobs:
...
+      # nothing consumes these images, so don't spend the time compressing
+      # them nor the storage keeping them around
+      upload_artifacts: false
```
basak-qcom: I'm unsure about this. If we get a test failure, debugging it would be assisted by having the build available. Theoretically it might be reproducible, but maybe that's the bug we'd be debugging. Does that make this a premature optimisation, or is there really serious enough pressure on the artifact store that we can't keep the artifacts around for a month or so?

basak-qcom (commented):
I think this is a worthwhile improvement. Thank you! A couple of comments inline for discussion.

basak-qcom:
Ah sorry I just saw the reference to #564 after clicking submit. If we're closing that issue, then should we add some snapshot-specific tests? I suggest:

1) Check that buildinfo contains the expected SNAPSHOT= line
2) Verify that we've correctly cleaned up so `apt-snapshot-toggle` is removed (see #582) and snapshots are disabled on all sources.list and sources.list.d entries.

basak-qcom:
A third, harder test might be to ensure that we don't have packages that are installed that are newer than the snapshot selected. That might not be worth it right now though for the effort it would take.

.github/workflows/build.yml
obbardc: I like the idea of generating the snapshot date based on the latest commit! I too can't see any huge negatives for that and if anything it makes it easier to keep in sync, since we don't have to manually update the snapshot date used in CI :-). I'll rework the PR for that.

But for your other part about PRs, remember that for now `build.yml` is a nightly build and not ran on any PR context, so should be fine. I wonder if we should test snapshot support pre-merge too, this brings up a similar question to #585 - perhaps we should consider amalgamating the different build workflows. Let's discuss that on #585.

.github/workflows/build.yml
obbardc: This was simply a premature optimisation and I am not attached to it. I agree with your thoughts about keeping the artifacts.

I will modify the PR to upload the artifacts, possibly under a separate directory `build_snapshot` or similar (to keep parity with the name).

I will keep the commit where I allowed disabling artifact upload separate for a later PR perhaps :-)

obbardc:
> Should we add some snapshot-specific tests? I suggest:
> 1. Check that buildinfo contains the expected SNAPSHOT= line

Thanks, I agree! I will do that and either extend the existing qemu test or just do something like check the files in the rootfs. Not decided yet. We could also check the sources.list contains the snapshot ID, it's another trivial check.

> 2. Verify that we've correctly cleaned up so `apt-snapshot-toggle` is removed (see [Snapshot state is conditionally removed, leaving a trap where it could fail to be removed #582](https://github.com/qualcomm-linux/qcom-deb-images/issues/582)) and snapshots are disabled on all sources.list and sources.list.d entries.

Agree.

> 3. Harder test might be to ensure that we don't have packages that are installed that are newer than the snapshot selected. That might not be worth it right now though for the effort it would take.

I agree, but I think that check could be a bit too much effort to implement right now, but I will create a follow-up issue about it once I've done 1 & 2.

.github/workflows/build.yml
obbardc: I moved snapshot build into a separate workflow && generated the timestamp based on the latest commit. Thanks!

.github/workflows/build.yml
obbardc: Done! I am not sure what happens to the artifacts if a job fails to build though; I think nothing gets uploaded (pending verification). Do you think that will matter?

PR #385, opened 2026-04-30
ci: test forky in linux.yml workflow

gagath:
The newly supported forky suite is only used in the build-on-pr and build-on-push workflows. Update the linux worklow so that the daily builds can also test forky.

PR #579, opened 2026-08-14
ci: test forky in test-on-pr.yml workflow

obbardc:
This is a draft as it may be superseded by #628 

---

`Build on PR` builds images for both trixie and forky, but only
trixie was ever booted on the boards. Run the LAVA tests over both
suites, like build-debian.yml and linux.yml already do.

Both builds are published to the same URL, with the suite as a
filename prefix, so the build URL retrieved from the build workflow
covers both and only the suite passed to lava-test.yml changes.

The publish job still only reports the trixie results; the next
commit extends it to both suites.

This is a follow-up from https://github.com/qualcomm-linux/qcom-deb-images/pull/385

## Verification
None yet...

`workflow_run` triggered workflows always execute the version of the file on the repo's default branch. So `test-on-pr.yml` on a PR branch can't be ran against the "new" workflow changes.

see sample workflow: https://github.com/qualcomm-linux/qcom-deb-images/actions/runs/31854327928

### DIFF
--- a/.github/workflows/test-on-pr.yml
+++ b/.github/workflows/test-on-pr.yml
@@ -61,6 +61,10 @@ jobs:
           echo "url=${BUILD_URL}" >> $GITHUB_OUTPUT
 
   test:
+    strategy:
+      fail-fast: false
+      matrix:
+        suite: [trixie, forky]
     uses: ./.github/workflows/lava-test.yml
     permissions:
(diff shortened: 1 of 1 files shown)


obbardc:
May be superseded by #628 ; will close this once that PR has been reviewed.

PR #589, opened 2026-08-20
feat(debos/rootfs): add URM (userspace-resource-manager) to rootfs

kartnema:
Userspace Resource Manager(URM) is a lightweight userspace daemon
which monitors system resources and enforces policies using
Linux kernel interfaces such as cgroups and sysfs.

Various workloads on different platforms can be supported by URM,
through the addition of appropriate Resource and Signal
configurations, without requiring any code changes. If a workload
requires advanced handling, it can be implemented cleanly through
the URM-plugin framework.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -401,6 +401,13 @@ actions:
     packages:
       # fastrpc-tests pulls in fastrpc-support
       - fastrpc-tests
+
+  # userspace-resource-manager is available from QLI overlays for trixie and forky
+  - action: apt
+    description: Install userspace-resource-manager from QLI overlay
+    recommends: true
+    packages:
+      - userspace-resource-manager
 {{- end }}
 
   - action: run


debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -388,6 +388,8 @@ actions:
       - tqftpserv
       # dynamic device nodes handling
       - udev
+      # Dynamic Resource Management Framework
+      - userspace-resource-manager
```
obbardc: Since this package isn't available from Debian repositories, this will fail when we build images without the QLI repository enabled (e.g. `debos -t qliaptrepo:false`).

Instead, can you add another action to install this package below [`Install fastrpc-tests from QLI overlay`](https://github.com/qualcomm-linux/qcom-deb-images/blob/main/debos-recipes/qualcomm-linux-debian-rootfs.yaml#L399), e.g:

```
{{- if and (eq $qliaptrepo "true") (ne $suite "unstable") }}
  # fastrpc-tests is available from QLI overlays for trixie and forky
  - action: apt
    description: Install fastrpc-tests from QLI overlay
    recommends: true
    packages:
      # fastrpc-tests pulls in fastrpc-support
      - fastrpc-tests

  - action: apt
    description: Install userspace-resource-manager from QLI overlay
    recommends: true
    packages:
      - userspace-resource-manager
{{- end }}
```

obbardc (changes requested):
We should only install package only when the QLI APT repo is enabled.

Also, I see there are other packages `userspace-resource-manager-tests` and `userspace-resource-manager-extensions`, do we need to install those in the image? 

Also, do we need to run the test suite in some LAVA CI test at all? Perhaps you could open up an issue about that, so we don't forget.

kartnema:
Thanks for the comments @obbardc 


> We should only install package only when the QLI APT repo is enabled.
> 
> Also, I see there are other packages `userspace-resource-manager-tests` and `userspace-resource-manager-extensions`, do we need to install those in the image?
> 
> Also, do we need to run the test suite in some LAVA CI test at all? Perhaps you could open up an issue about that, so we don't forget.


- My understanding was that pkg-urm has already been released into QLI post reviews (https://jira-dc4.qualcomm.com/jira/browse/OSTT-1234), additionally I can see our packages as part of "qli" (earlier they were in qli-staging . Is this correct? If so, then we only need to worry about adding some checks like you indicated to build URM only if QLI APT is enabled.

- URM core functionality is covered by the main package itself; the userspace-resource-manager-extensions package provide additional customizations through plugins and configurations (which hook into URM to override / extend behavior). userspace-resource-manager-extensions is not mandatory (as URM can function standalone without extensions), however it will be good to have if available. Something similar to "RRECOMMENDS" in Yocto/Bitbake environments, would be suitable. 
userspace-resource-manager-tests should not be installed by default.

- userspace-resource-manager-tests covers the required test cases which need to be run, I have included them as part of our suite in linux-qcom-testkit, along with package dependencies. Refer: https://github.com/qualcomm-linux/qcom-linux-testkit/blob/main/Runner/suites/Performance/userspace-resource-manager/run.sh. Please let me know the steps so that it can be included as part of Lava CI.

obbardc (approved):

PR #593, opened 2026-08-21
refactor(debos/flash)!: switch RB1 && UnoQ to EDK2-based boot firmware

obbardc:
Switch QRB2210-based devices (RB1 and UnoQ) to use the EDK2-based boot firmware. This removes the need for a custom U-Boot binary to be built and shipped with the images. 

Tested on Arduino UnoQ locally and RB1 in CI.

---
## Outstanding points (PR is draft until these are resolved)
- [ ] Test image on RB1
- [ ] Move the commit removing `u-boot-efi-dtb` to a separate PR?
- [ ] Move the commit removing `qbootctl` to a separate PR?

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -111,9 +111,6 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
-      - name: Copy U-Boot for RB1 from EFS
-        run: cp -av "/efs/u-boot-rb1-latest/rb1-boot.img" .
-
       # mtools is needed for the flash recipe
       # file, device-tree-compiler and u-boot-tools are needed by qcom-dtb-metadata
       - name: Install debos and dependencies of the recipes and local tests
@@ -215,7 +212,7 @@ jobs:
       - name: Build flashable files with debos
         run: |
           set -ux
-          make flash FAKEMACHINE_OPTS="-b qemu" EXTRA_DEBOS_OPTS="-t u_boot_rb1:rb1-boot.img -t buildid:${BUILD_ID} --print-recipe ${DEBOS_EXTRA_ARGS} ${DEBOS_VARIANT_ARGS}"
+          make flash FAKEMACHINE_OPTS="-b qemu" EXTRA_DEBOS_OPTS="-t buildid:${BUILD_ID} --print-recipe ${DEBOS_EXTRA_ARGS} ${DEBOS_VARIANT_ARGS}"
 
       - name: Stage debos artifacts for publishing
         run: |
--- a/.github/workflows/linux-mainline.yml
+++ b/.github/workflows/linux-mainline.yml
@@ -1,9 +1,8 @@
 name: mainline linux build
 
 on:
-  # run weekly on Monday at 2:30pm, after the daily builds and the weekly u-boot
-  # build have finished for the day, so this doesn't add its LAVA jobs on top of
-  # theirs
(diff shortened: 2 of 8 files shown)


lool:
@obbardc Thanks for starting this; a few bits:
- yes, https://github.com/qualcomm-linux/qcom-dtb-metadata/issues/126 is related but you use a "combined dtb" style for both boards in the board database instead of a FIT image, and it will work for now – we'll only miss fancy overlays, but some other boards miss it it too
- yes, we could remove qbootctl, these were the only official platforms that needed it, but it's a bit orthogonal, and Dmitry might be adding support for platforms that need it, so we can do it in a followup PR, taking the time to discuss
- yes, we can drop u-boot-efi-dtb, mark this as a breaking change as arduino-deb-images might depend on it, but they can continue adding it in their layer; you can also defer to a separate PR
- don't bother with EFS cleanup, there's an artifact aging process that should eventually garbage collect this
- I would only deprecate the U-Boot scripts and drop them from docs; you can also do this in a separate PR

I would convert both agatti boards in one PR – RB1 and UNOQ, otherwise the current scope of your PR is fine.

lool:
We should also test on UNOQ, UNOQ still has u-boot in its zip with boot binaries; I'm not sure how our changes:
- affect current boot
- affect Arduino's expectations (but that can be managed in arduino-deb-images)

There are 4 boot stacks on Agatti right now:
- old Android: foundation that we built upon, that we want to get away from
- u-boot: that we currently chainload but that could be standalone
- new QLI boot binaries: 90% complete, not sure what's missing, what you're moving to
- opensource upstream edk2/atf/optee: currently booting, but not sure how upstream the patches are and how complete the port is

Generally happy to progress from 1/2 to 3, your PR seems to be going in the right direction. The challenge is with UNOQ and upgrade path, we want to make sure we don't break:
- normal boot (u-boot-efi-dtb to help loading the latest DTB, qbootctl to set the boot as successful etc.)
- DTB overlay functionality: Arduino has some scripts to patch the DTB in ESP for e.g. hats
- upgrades: will partition layout be compatible before/after

obbardc:
@lool

> We should also test on UNOQ, UNOQ still has u-boot in its zip with boot binaries; I'm not sure how our changes:
> 
> * affect current boot
> * affect Arduino's expectations (but that can be managed in arduino-deb-images)

Just FYI, this PR does work fine on UNOQ, we no longer consume the Arduino boot binaries zip :-).
I did the initial bringup etc locally on UNOQ but didn't do so with the RB1 as I don't have one here, but it works fine in CI for both boards.


> There are 4 boot stacks on Agatti right now:
> 
> * old Android: foundation that we built upon, that we want to get away from
> * u-boot: that we currently chainload but that could be standalone
> * new QLI boot binaries: 90% complete, not sure what's missing, what you're moving to
> * opensource upstream edk2/atf/optee: currently booting, but not sure how upstream the patches are and how complete the port is
> 
> Generally happy to progress from 1/2 to 3, your PR seems to be going in the right direction. The challenge is with UNOQ and upgrade path, we want to make sure we don't break:
> 
> * normal boot (u-boot-efi-dtb to help loading the latest DTB, qbootctl to set the boot as successful etc.)
> * DTB overlay functionality: Arduino has some scripts to patch the DTB in ESP for e.g. hats
> * upgrades: will partition layout be compatible before/after

See above :-).

PR #595, opened 2026-08-21
feat!: install the qcom-next kernel package from QLI APT

### LINKED ISSUE #386: qcom-deb-images: [Glymur Debian]  switch to consume kernel debs generated by pkg-linux-qcom
qcom-deb-images nightly currently building qcom-next kernel Debs using bindeb-pkg 
- no validation/testing of kernel deb's in place before it gets consuming into qcom-deb-images  ( usecase : wlan broken on rc7) 

As per previous discussion, 
> qcom-deb-images should consume kernel debs generated with standard debian metadata, 
        > pkg-linux-qcom(qcom/debian/latest) has debian metadata to produce kernel debs based on qcom->next.
        > validate kernel debs over axiom 
        > Host PROD PPA 

Raise PR to update package name of kernel into rootfs.yaml as linux-image-7.0.0-rc7-qcom  (for corresponding deb package linux-image-7.0.0-rc7-qcom_1-1_arm64.deb) , PR go through validation at qcom-deb-image level & merge.

which ensures qcom-deb-images nightly/PR Build should consume latest kernel deb of qcom-next from PROD PPA & reproducibility of snapshot

obbardc:
Install the qcom-next kernel from the QLI APT repository, instead of from the debs which the `qcom-next linux build` workflow publishes to EFS.

This decouples the image builds from the kernel build: the images become reproducible from APT alone, the "Build" workflow no longer has to chain onto another workflow to get a kernel and the kernel tag no longer has to be bumped by hand in this repository.

## Dependencies / Papercuts
This PR will remain a draft until these are resolved.
- The current kernel packages in the QLI repository do not create `systemd-boot` menu entries on `forky` after package install (initial images are fine: `systemd-boot.postinst` calls `kernel-install`). Fixed with https://github.com/qualcomm-linux/pkg-linux-qcom/pull/121
* The current kernel packages in the QLI repository installs both the mainline/in-tree drivers and the prebuilt out-of-tree modules at the same time. The implementation is loaded is not deterministic: module discovery and loading can select either the in-tree or out-of-tree driver depending on aliases, probe order, initramfs contents and other boot-time details.
  * Commit `feat(debos/rootfs): disable the out-of-tree drivers` works around this by adding image-specific modprobe configuration that blacklists the out-of-tree modules by default and commit `feat(debos/rootfs): enable the out-of-tree drivers for multimedia` then adds further image-specific configuration to reverse that behaviour for the multimedia image.
  * This should not be controlled by the image recipes. In particular, a persistent `/etc/modprobe.d/` policy is tied to the root filesystem rather than to the kernel/module set that required it and can give incorrect behaviour when changing or upgrading kernels.
  * Instead we should split the out-of-tree modules into a separate package. The default image can then use only the mainline drivers, while users that require the downstream multimedia stack can explicitly install the out-of-tree module package & configuration. Need to drop the two image-specific enable/disable commits and add documentation describing how to switch from the mainline drivers to the out-of-tree modules when the downstream multimedia stack is required. Should be fixed with https://github.com/qualcomm-linux/pkg-linux-qcom/pull/123 - still under test.

## vmlinux extractor validation

Check that unique, unstripped vmlinux files ends up in artifactdir.

- [x] Test new -dbg vmlinux extractor in CI (qcom-next package; with full-size vmlinux)
- [x] Test new -dbg vmlinux extractor in CI (debian build)
- [x] Test new -dbg vmlinux extractor in CI (EFS kernel)

## kernel package validation
- [x] trixie & forky kernel package installs && boots with correct `uname -r`
- [x] trixie & forky `-dbg` package is unstripped (e.g. unique, unstripped vmlinux ends up in artifactdir)
- [x] trixie systemd-boot menu entry present & matching package version
- [x] forky systemd-boot menu entry present & matching package version
- [ ] trixie systemd-boot menu entry present & matching package version on PACKAGE UPGRADE
- [ ] forky systemd-boot menu entry present & matching package version on PACKAGE UPGRADE
- [ ] trixie `kgsl,camx,iris-vpu,audioreach` out-of-tree modules are not auto loaded
- [ ] forky `kgsl,camx,iris-vpu,audioreach` out-of-tree modules are not auto loaded
- [ ] trixie `kgsl,camx,iris-vpu,audioreach` out-of-tree modules CAN be auto loaded (opt-in)
- [ ] forky `kgsl,camx,iris-vpu,audioreach` out-of-tree modules CAN be auto loaded (opt-in)

## Follow-ups
- [All debug kernel images are much larger than those originally built with bindeb-pkg (450M vs 170M)](https://github.com/qualcomm-linux/pkg-linux-qcom/issues/103) 
- [Configuration from DKMS modules isn't shipped in kernel package (make it opt-in?)](https://github.com/qualcomm-linux/pkg-linux-qcom/issues/105) && DKMS modules are always loaded by default.

Fixes: https://github.com/qualcomm-linux/qcom-deb-images/issues/386

### DIFF
--- a/.github/workflows/debos.yml
+++ b/.github/workflows/debos.yml
@@ -17,7 +17,7 @@ on:
           repositories (e.g. linux-image-arm64) or name of a EFS kernel
           package directory (e.g. efs/mainline)
         type: string
-        default: linux-image-arm64
+        default: ''
       debos_extra_args:
         description: Extra arguments to pass to debos (e.g. -t dtb:qcom/some.dtb)
         type: string
--- a/.github/workflows/linux-qcom-next.yml
+++ b/.github/workflows/linux-qcom-next.yml
@@ -7,10 +7,6 @@ on:
   # images 10:30am) so that they don't all submit LAVA jobs at once. The "Build"
   # workflow chains onto this one, so its LAVA jobs land later still, around
   # 3:30am.
-  #
-  # the images the "Build" workflow produces are shared with the test teams and
-  # are expected by 8:00am IST (2:30am UTC); building from 0:30am leaves an hour
-  # of headroom, see issue #566.
   schedule:
     - cron: '30 0 * * *'
   # allow manual runs
--- a/.github/workflows/linux-qcom-next.yml
+++ b/.github/workflows/linux-qcom-next.yml
@@ -1,3 +1,9 @@
+# DEPRECATED: the QLI Debian images install the qcom-next kernel from the QLI
+# APT repository, so nothing consumes the debs this workflow publishes any
+# more. It is kept as a reference build, to compare the kernel built from the
+# qcom-next tree against the one QLI ships, and should be removed once that
+# comparison stops being useful.
+
 name: qcom-next linux build
 
 on:
--- a/.github/workflows/linux-qcom-next.yml
+++ b/.github/workflows/linux-qcom-next.yml
@@ -31,9 +31,6 @@ jobs:
     with:
       kernel_name: qcom-next
       build_linux_deb_extra_args: '--qcom-next prune.config qcom.config'
-      # stop after publishing the debs to EFS; the "Build" workflow
-      # picks them up and builds and tests the images on every suite
-      skip_image_build: true
       # https://github.com/qualcomm-linux/qcom-deb-images/issues/537
       lava_boards_exclude: '["monaco-arduino-monza", "qrb2210-arduino-imola"]'
     secrets:
--- a/.github/workflows/linux-qcom-next.yml
+++ b/.github/workflows/linux-qcom-next.yml
@@ -30,7 +30,7 @@ jobs:
       pull-requests: write
     with:
       kernel_name: qcom-next
-      build_linux_deb_extra_args: '--qcom-next --ref qcom-next-7.2-rc7-20260821 prune.config qcom.config'
+      build_linux_deb_extra_args: '--qcom-next prune.config qcom.config'
       # stop after publishing the debs to EFS; the "Build" workflow
       # picks them up and builds and tests the images on every suite
       skip_image_build: true
--- a/.github/workflows/linux-qcom-next.yml
+++ b/.github/workflows/linux-qcom-next.yml
@@ -1,14 +1,12 @@
 name: qcom-next linux build
 
 on:
-  # run daily at 0:30am. qcom-next is what QLI ships, so it goes first of the
-  # scheduled builds and gets an empty LAVA queue; the rest of the day's builds
-  # are staggered behind it (arduino 6:30am, linux-next 8:30am, vanilla Debian
-  # images 10:30am) so that they don't all submit LAVA jobs at once. The "Build"
-  # workflow chains onto this one, so its LAVA jobs land later still, around
-  # 3:30am.
+  # run daily at 4:30am, after the "Build" workflow started at 0:30am has
+  # submitted its LAVA jobs, so that this build doesn't add its own on top of
+  # theirs; the rest of the day's builds are staggered behind it in turn
+  # (arduino 6:30am, linux-next 8:30am, vanilla Debian images 10:30am)
   schedule:
-    - cron: '30 0 * * *'
+    - cron: '30 4 * * *'
   # allow manual runs
   workflow_dispatch:
 
--- a/.github/workflows/linux-qcom-next.yml
+++ b/.github/workflows/linux-qcom-next.yml
@@ -31,6 +31,9 @@ jobs:
     with:
       kernel_name: qcom-next
       build_linux_deb_extra_args: '--qcom-next prune.config qcom.config'
+      # reference build: we want the debs and the LAVA results to compare
+      # against, but its failures must not gate qcom-deb-images work
+      lava_report_test_results_externally: false
       # https://github.com/qualcomm-linux/qcom-deb-images/issues/537
       lava_boards_exclude: '["monaco-arduino-monza", "qrb2210-arduino-imola"]'
     secrets:
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -425,6 +425,18 @@ actions:
     packages:
       # fastrpc-tests pulls in fastrpc-support
       - fastrpc-tests
+
+  - action: run
+    description: Prefer kernel from QLI overlay
+    chroot: true
+    command: |
+      set -eux
+      tee /etc/apt/preferences.d/qli-kernel.pref <<EOF
+      Package: linux-image-*-qcom-next-* linux-image-qcom-next linux-headers-*-qcom-next-* linux-headers-qcom-next
+      # TODO: prefer pinning on Release/Label; see https://github.com/qualcomm-linux/qcom-deb-images/pull/576/changes#r3787588202
+      Pin: origin "deb.debusine.qualcomm.com"
+      Pin-Priority: 1001
+      EOF
 {{- end }}
 
   - action: run
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -276,6 +276,25 @@ actions:
        =bX6V
        -----END PGP PUBLIC KEY BLOCK-----
       EOF
+      tee /etc/apt/sources.list.d/qli-staging.sources <<EOF
+      # Qualcomm Linux repository
+      Types: deb
+      URIs: https://deb.debusine.qualcomm.com/qualcomm/qli-staging
+      Suites: {{ $suite }}
+      Components: main contrib non-free-firmware non-free
+      Enabled: yes
+      Signed-By:
+       -----BEGIN PGP PUBLIC KEY BLOCK-----
+       .
+       mDMEajuyfxYJKwYBBAHaRw8BAQdASazjfos7KwJJ+G6xdBRzc3v7orITHEY6jKc3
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -149,8 +149,6 @@ A few options are provided in the debos recipes; for the root filesystem recipe:
   apt; defaults to `linux-image-arm64`. Set it to `none` when you are installing
   a local kernel package. See the [Kernel](#kernel) section below for
   more information.
-- `kernelpackage`: **deprecated**, superseded by `kernelpackages`; still
```
lool, line 152: This was only added in July, can't we keep it a bit longer? Not that we have a ton of downstream users, but I'd like to signal that we would care.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -790,6 +790,24 @@ actions:
...
 {{- end }}
 
+  # TODO: Either make the out-of-tree modules opt in or ship this config with the
+  # kernel package instead; see https://github.com/qualcomm-linux/pkg-linux-qcom/issues/105
+  - action: run
+    description: Disable out-of-tree modules
+    chroot: true
+    command: |
+      set -eux
+      cat > /etc/modprobe.d/disable-out-of-tree.conf <<EOF
```
lool, line 800: This filename will become an interface of ours, especially if we refer to it in docs, or provide a CLI utility to flip flop between drivers. How about:
/etc/modprobe.d/qli.conf

debos-recipes/overlays/multimedia-modules/usr/lib/udev/rules.d/50-kgsl.rules
```diff
@@ -0,0 +1 @@
+ACTION=="add", SUBSYSTEM=="kgsl", KERNEL=="kgsl-3d0", OWNER="root", GROUP="video", MODE="0660"
```
lool: that snippet should be provided in the kernel package and copied from the dkms upstream sources

debos-recipes/overlays/multimedia-modules/usr/lib/modprobe.d/audioreach-blacklist.conf
```diff
@@ -0,0 +1,8 @@
+blacklist q6apm-lpass-dais
```
lool, line 1: Leaving a comment here on the first file in usr/lib, but basically these files should be in the kernel package, not in the image recipe.

.github/workflows/linux-qcom-next.yml
```diff
@@ -1,19 +1,18 @@
+# DEPRECATED: the QLI Debian images install the linux-image-qcom-next package
```
lool, line 1: Do we need to clarify the table/update the badges in the README as well?

ci/vmlinux/download-vmlinux.yaml
```diff
@@ -0,0 +1,74 @@
...
+      set -eux
+
+      # Query dpkg for the highest-version installed kernel package.
+      # linux-image-[0-9]* pattern skips metapackages and the
+      # ${db:Status-Abbrev} filter skips packages which are known to dpkg but
+      # not installed.
+      kernel_pkg=$(dpkg-query -W -f='${db:Status-Abbrev} ${Package}\n' \
+              'linux-image-[0-9]*' |
+          sed -n 's/^ii  //p' |
+          sort -V |
```
lool, line 41: Should this be from `linux-version list` or `sort`, to have at least the proper version sorting algorithm?

ci/vmlinux/download-vmlinux.yaml
```diff
@@ -0,0 +1,74 @@
...
+          tail -n1)
+      kernel_pkgver=$(dpkg-query -W -f='${Version}' "$kernel_pkg")
+      kernel_dbgpkg="${kernel_pkg}-dbg"
+
+      # Remove existing artifacts.
+      rm -rf /*.deb
+
+      # Not every kernel has a matching debug package. Warn rather than failing
```
lool, line 49: I wondered if we should break the build or not extract any dbg if there are multiple kernels, but I guess it's fair to extract the dbg vmlinux for the kernel that would boot by default. We could also extract one vmlinux per `uname -r`, I've seen that in debian/ubuntu live images.

ci/vmlinux/download-vmlinux.yaml
```diff
@@ -0,0 +1,74 @@
...
+              "can't extract vmlinux" >&2
+          exit 0
+      fi
+      deb=$(find . -maxdepth 1 -type f -name "${kernel_dbgpkg}_*.deb" \
+          -print -quit)
+
+      # Hand the package over to the extraction below at a known path; the
+      # download can only write inside the root filesystem, which the script
+      # then reads it back out of.
+      mv -v "$deb" /kernel-dbg.deb
```
lool, line 66: where do we rm this downloaded deb?

README.md
```diff
@@ -20,14 +20,12 @@ tuned!
...
-| Reference upstream kernels | [![linux-next][linux-next-badge]][linux-next] [![mainline][mainline-badge]][mainline] | *covered by the kernel build* [^1] |
-| Project-specific builds | [![arduino][arduino-badge]][arduino] | *covered by the kernel build* [^1] |
-
-[^1]: these builds compile the kernel, build an image and boot test the image in a single workflow, so the Kernel build reports all three.
+| Variant | Image build & test status |
+| ------- | --- |
+| Qualcomm Linux | [![QLI images][images-badge]][images] |
+| Debian | [![Debian images][debian-images-badge]][debian-images] |
+| Reference upstream kernels | [![linux-next images][linux-next-badge]][linux-next] [![mainline images][mainline-badge]][mainline] |
+| Project-specific builds | [![Arduino images][arduino-badge]][arduino] |
```
lool, line 28: I'm reluctant to use "Arduino" with a capital A; we are building the kernel branch of the QLI kernel named "early/hwe/arduino", and in parallel Arduino themselves are publishing their images, the real "Arduino images". I can't quite think of a more careful name that is also short. Let's perhaps just leave it as "arduino" or "arduino images"?

lool (changes requested):
Main thing that bugs me is usr files create by the image generation; perhaps we could live with some /etc kludges instead (etc/udev, etc/initramfs, etc/modprobe.d ...)? Or just prio fixing the generated kernels?

ci/vmlinux/download-vmlinux.yaml
obbardc: It is automically cleaned up when debos exists; I added a clarification to the comment.

ci/vmlinux/download-vmlinux.yaml
obbardc: at this point, we should only have one kernel package installed, so I think it makes sense to enforce that rather than selecting the highest installed version... I'll change this to fail the build if there are zero or multiple installed kernels. This also removes the need for version sorting. Will leave it as a separate commit so we can revert if needed.

ci/vmlinux/download-vmlinux.yaml
obbardc: Done.

.github/workflows/linux-qcom-next.yml
```diff
@@ -36,9 +35,9 @@ jobs:
...
     with:
       kernel_name: qcom-next
-      build_linux_deb_extra_args: '--qcom-next --ref qcom-next-7.2-20260826 prune.config qcom.config'
-      # stop after publishing the debs to EFS; the "Build" workflow
-      # picks them up and builds and tests the images on every suite
-      skip_image_build: true
+      build_linux_deb_extra_args: '--qcom-next prune.config qcom.config'
+      # reference build: we want the debs and the LAVA results to compare
+      # against, but its failures must not gate qcom-deb-images work
+      lava_report_test_results_externally: false
```
obbardc: Need to check if this is correct with new CI scheme.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -781,10 +781,16 @@ actions:
...
 # Install the accelerated multimedia stack: out-of-tree kernel drivers and
 # closed source user-space from the Qualcomm Linux APT repository.
+# For now it is only compatible with the packaged qcom-next kernel.
 
 {{- if ne $qliaptrepo "true" }}
 # Call an unexisting template to err out of the recipe
 {{ template "ERR" }}
+{{- end }}
+
+{{- if ne $kernelpackages "linux-image-qcom-next" }}
```
lool: what if I want to build with a local package?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -799,6 +799,11 @@ actions:
       - gcc
       - dkms
 
+  - action: apt
+    description: Install kernel headers
+    packages:
+      - linux-headers-qcom-next
```
lool: We can't hardcode a specific flavor here

obbardc: sure, will build the images with a default `kernelpackages = linux-image-qcom-next,linux-headers-qcom-next` so we always have the kernel headers, and users can build with their own set. Think that's fine to me.

debos-recipes/qualcomm-linux-debian-rootfs.yaml
obbardc: makes sense, will remove

.github/workflows/debos.yml
```diff
@@ -20,7 +17,7 @@ on:
           repositories (e.g. linux-image-arm64) or name of a EFS kernel
           package directory (e.g. efs/mainline)
         type: string
-        default: linux-image-arm64
+        default: 'linux-image-qcom-next,linux-headers-qcom-next'
```
obbardc: I'm more in favour of duplicating this to the `build*.yml` workflows with a nice comment, so readers of the workflows know what kernel they are getting.

PR #596, opened 2026-08-22
fix(debos/flash): revert qcom-ptool after RB1 regression

obbardc:
The reverted commit bumps qcom-ptool to a version which includes a
breaking change to the qrb2210-rb1 partition layout, which switches the
partitions to a new layout for the EDK-II based bootflow. Unfortunately
that is incompatible with the existing U-Boot based bootflow which the
qcom-deb-images currently uses (see switching to EDK-II in https://github.com/qualcomm-linux/qcom-deb-images/pull/593) and
removes the boot_a and boot_b partitions where we currently chain
U-Boot. Revert the change to allow qrb2210-rb1 to boot again.

Also, update qcom-ptool revision to 80e52fc, just before the breaking changes to qrb2210-rb1 were introduced, which still include the change for [platforms/glymur-crd: update spinor partitions for dual-boot support](https://github.com/qualcomm-linux/qcom-ptool/commit/80e52fcde1cbae7d725692dea98a27cb479c5a2c)

cc @neervish - can you re-test this on the glymur-crd ?

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -15,10 +15,10 @@ architecture: arm64
 actions:
   - action: download
     description: Download qcom-ptool
-    url: https://github.com/qualcomm-linux/qcom-ptool/archive/b147606c32b7d29148eca16fbe071f9420dd51a9.tar.gz
+    url: https://github.com/qualcomm-linux/qcom-ptool/archive/80e52fcde1cbae7d725692dea98a27cb479c5a2c.tar.gz
     name: qcom-ptool
     filename: qcom-ptool.tar.gz
-    sha256sum: a9fb80e1e03400d228c784a12c55de66ff7144ec8f07732f0dff332b33ead242
+    sha256sum: 4fe113e20fa8feffdd962612f17cb03e60e6e7f460738bbcf897c7c9c631ad06
     unpack: true
 
   - action: download
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -15,10 +15,10 @@ architecture: arm64
 actions:
   - action: download
     description: Download qcom-ptool
-    url: https://github.com/qualcomm-linux/qcom-ptool/archive/83631c3415402908fedabf8b54dbf2363353094b.tar.gz
+    url: https://github.com/qualcomm-linux/qcom-ptool/archive/b147606c32b7d29148eca16fbe071f9420dd51a9.tar.gz
     name: qcom-ptool
     filename: qcom-ptool.tar.gz
-    sha256sum: 39068987e6e3b8a892822d02e7e5bae75b8b589f3849f8f2e47bebed4f09eb94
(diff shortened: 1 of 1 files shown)


neervish:
@obbardc Tried to validate with this change on glymur, but hitting a 403 Forbidden when downloading the Glymur boot binaries , i see the same error on build workflow as well.

Just flagging this here, Will retry again sometime today.

2026/08/24 04:29:47 Action `Download of Glymur boot binaries` failed at stage Run, error: url 'https://softwarecenter.qualcomm.com/nexus/generic/product/chip/tech-package/Glymur_bootbinaries.1.0/glymur_bootbinaries.1.0-test-device-public/00080/Glymur_bootbinaries.zip' returned status code 403 (Forbidden)
2026/08/24 04:29:47 fakemachine failed with non-zero exitcode: 1
make: *** [Makefile:74: flash] Error 1

neervish:
> Christopher Obbard (@obbardc) Tried to validate with this change on glymur, but hitting a 403 Forbidden when downloading the Glymur boot binaries , i see the same error on build workflow as well.
> 
> Just flagging this here, Will retry again sometime today.
> 
> 2026/08/24 04:29:47 Action `Download of Glymur boot binaries` failed at stage Run, error: url 'https://softwarecenter.qualcomm.com/nexus/generic/product/chip/tech-package/Glymur_bootbinaries.1.0/glymur_bootbinaries.1.0-test-device-public/00080/Glymur_bootbinaries.zip' returned status code 403 (Forbidden) 2026/08/24 04:29:47 fakemachine failed with non-zero exitcode: 1 make: *** [Makefile:74: flash] Error 1

Not facing this issue now.
Built locally with this PR changes, flashed spinor and NVMe on Glymur, Device boots up fine.

lool (changes requested):
It looks like the breakage only affects RB1, UNO Q still boots (surprising TBH), so I'd prefer keeping the glymur-crd partitioning features which we want for our end of month release (dual boot compatible partition table) and accelerate landing of https://github.com/qualcomm-linux/qcom-deb-images/pull/593 instead – it seems in a good shape already

obbardc:
> It looks like the breakage only affects RB1, UNO Q still boots (surprising TBH)

Yeah, the upstream change only introduced the change into RB1: https://github.com/qualcomm-linux/qcom-ptool/pull/109
Uno Q is still pending to be integrated into qcom-ptool.

> so I'd prefer keeping the glymur-crd partitioning features which we want for our end of month release (dual boot compatible partition table) and accelerate landing of #593 instead – it seems in a good shape already

Yep, the second PR bumps to a version of qcom-ptool which has https://github.com/qualcomm-linux/qcom-ptool/pull/158 (and was verified by @neervish above) :-)

So I see this is a separate from #593 - we can merge this now to cause the CI to go green again, and work on #593 in lower prioirty.

lool (approved):
Ok, checked this including the second commit now; the current ptool history is basically:
- glymur-crd change we want, then
- rb1 change that broke us, then
- number of other changes not particularly critical (one qcs615 fix) => we were there, then
- number of other potentially interesting changes

So I'm ok to go back in time to just the glymur-crd dual boot changes we want.

PR #597, opened 2026-08-24
fix(debos/flash): update qcom-dtb-metadata for Glymur Staging support

aathitha:
Update qcom-dtb-metadata for Glymur Staging support.

- Updated qcom-dtb-metadata SHA from d30e9cdb to 70b09af283da4fba2a2251ff73a7c59ec2699471
- Updated sha256sum accordingly to 0c9ef6ed0a07de39c178d42093f8c5c05b28da3fdaeaf506dacf8aa66b0e94b1
- This picks up PR # 133 changes from qcom-dtb-metadata which adds Glymur Staging overlay support
- fdt-glymur-staging.dtbo added in qcom-next-fitimage.its
- Validated on Glymur CRD device and glymur-staging.dtbo present at /usr/lib/linux-image-7.2.0-rc7-g011a82096bee/qcom/glymur-staging.dtbo
- Device boots successfully at end.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-flash.yaml
+++ b/debos-recipes/qualcomm-linux-debian-flash.yaml
@@ -23,10 +23,10 @@ actions:
 
   - action: download
     description: Download qcom-dtb-metadata (DTB image build scripts and ITS/metadata files)
-    url: https://github.com/qualcomm-linux/qcom-dtb-metadata/archive/d30e9cdb766f08b05160ededf53556ed665d0c0b.tar.gz
+    url: https://github.com/qualcomm-linux/qcom-dtb-metadata/archive/70b09af283da4fba2a2251ff73a7c59ec2699471.tar.gz
     name: qcom-dtb-metadata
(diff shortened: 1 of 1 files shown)


lool:
@aathitha I lack context on this new Glymur Staging entry; what is Staging referring to here? and how do you use it in practice? how long will we have it?

lool (approved):
So @nivasnaik was kind of enough to research this:
- for advanced debugging purposes, we want Embedded Trace Router to work
- this requires a DTB overlay over the default DTB, that can be selected by an EFI var
- since this hadn't been merged upstream yet, it was being delivered through a "staging" DTBO

In the meantime, this has been merged upstream, so the staging DTBO shouldn't be needed.

I'd be happy to merge an updated qcom-dtb-metadata, just for the sake of getting support for new platforms (Nord) or features like this staging DTBO, albeit in practice we don't really benefit from any of these changes with this PR alone.

I reviewed the diff between old and new qcom-dtb-metadata, and it was only additions of configs.

Issue #604, opened 2026-08-27
Create a new "performance" debian build for Glymur


ctheegal:
Please maintain a new "performance" debian build flavor for Glymur. 

Kernel defconfigs to enabled: None as of now. Will keep adding in future when required.

Kernel cmdline parameters: 
quiet 
systemd.tty.term.console=linux

PR #607, opened 2026-08-27
feat: add image profiles with initial debug and performance profiles

obbardc:
Add a `profile` variable to the image recipe to select the intended
runtime configuration of an image, plus initial `performance` and `debug`
profiles.

The profile only affects the image recipe, so the same `rootfs.tar` can be
used to build every profile. When the profile isn't `default`, the image
records it in `/etc/buildinfo` as `PROFILE=<profile>`.

- `performance`: appends `quiet systemd.tty.term.console=dumb` to the kernel
  command line, based on meta-qcom's `ci/performance.yml`.
- `debug`: enables ftrace from boot, `qcom_scm.download_mode=1` and the systemd
  hardware watchdog, based on meta-qcom's `ci/debug.yml` and `ci/base.yml`.
  **Not ready for use yet**: it still installs the default kernel, as it
  depends on the pending `qcom-next-debug` kernel package
  (qualcomm-linux/pkg-linux-qcom#69, #568).

Follow-ups:
- Build and test the image profiles in CI (depends on #628 and should be after #405 lands).
- Install the debug kernel in the debug image, once `qcom-next-debug` is available.

Outstanding questions:
- `PROFILE=` is only written to `/etc/buildinfo` for non-default profiles, so
  default images are unchanged. Should `PROFILE=default` be recorded too, so
  the field is always present?
- Should the debug profile move to its own PR, together with installing the
  debug kernel, leaving this one to add the mechanism and the performance
  profile?


debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -158,6 +173,9 @@ actions:
 {{- if $variantid }}
       echo "VARIANT_ID={{$variantid}}" >>/etc/buildinfo
 {{- end }}
+{{- if ne $profile "default" }}
+      echo "BUILD_PROFILE_ID={{$profile}}" >>/etc/buildinfo
```
lool: Unrelated, but looking back at it, I find VARIANT_ID slightly weird instead of VARIANT

Could BUILD_PROFILE_ID be simply PROFILE? Do we fear it will clash with something?

lool:
Loïc asked claude for a review, please find it below

---

Overall this is a clean, well-decomposed series: the mechanism (`cb95eaa`) is separated from the refactor (`724276d`) and from the first user of it (`65b05e5`, `f792a8a`), and each commit stands on its own. The `/etc/buildinfo` round-trip so the image recipe can detect a rootfs built with a different profile is a genuinely good idea — those two debos invocations disagreeing silently is exactly the kind of bug that costs an afternoon.

Things verified rather than assumed, since they're the load-bearing bits:

- `systemd.tty.term.console=linux` is real and does what `65b05e5` claims. Per systemd's `man/kernel-command-line.xml`, `systemd.tty.term.console` is special-cased to also set the `$TERM` used by PID 1 itself when `TERM=` isn't on the cmdline. Available since **v254**, so trixie (257) and forky are both fine.
- The `include:` entry in `build.yml` adds a **5th** combination rather than mutating trixie/weston-multimedia, because `profile: [default]` is already a matrix axis and `include` can't overwrite an existing key. That matches `f792a8a`'s intent.
- `DEBOS_PROFILE_ARGS` via `$GITHUB_ENV` is sound — `build-debos` is a single job, so it reaches the rootfs build, both disk images and the flash bundle.
- The new `PREFIX` expression yields `trixie-weston-multimedia-performance` and leaves existing artifact names byte-identical for default profiles. No churn for downstream consumers.

## Substantive

**1. The performance profile gets no hardware boot validation.** `f792a8a` builds it only for trixie + `weston-multimedia`, but `ci/lava/*/boot.yaml` hard-codes `{{BUILD_DOWNLOAD_URL}}/{{SUITE}}-flash-emmc.tar.gz` — i.e. the *default* variant only. So the one profile whose entire purpose is changing the kernel command line never boots on real hardware in CI; it only gets `ci/qemu_test.py`. And `quiet` plus a forced `TERM` is precisely the class of change that breaks serial-console harnesses. If the profile were applied to the `default` variant instead, it would land in the LAVA-tested artifact for the same one extra job. Worth reconsidering which combination to spend the job on.

Related caveat: forcing `TERM=linux` on a serial console means systemd emits `linux`-style escape sequences to that port. Harmless today (the QEMU test matches agetty's `debian login:`, not systemd output), but if this profile ever reaches a LAVA-tested artifact, the `auto_login` and `prompts` regexes are then matching a stream with colour codes in it.

**2. A whole extra rootfs for a cmdline-only difference.** For `performance` the rootfs differs from default by exactly one line in `/etc/buildinfo`, yet the cross-check forces a full extra rootfs build plus SBOM, grype scan, both disk images and the flash bundle. That's defensible as forward-looking (`debug` presumably *will* change rootfs contents), but it reads oddly against `f792a8a`'s stated aim of keeping CI jobs minimal. Worth a sentence in the commit message confirming it's deliberate.

**3. `rootfs.tar` staleness makes the new guard's advice unactionable.** In the `Makefile`, `rootfs.tar` depends only on the recipe file, so a changed `-t profile:` doesn't trigger a rebuild. The README example is exactly the command that hits this:

```
make EXTRA_DEBOS_OPTS="-t profile:performance" rootfs.tar disk-ufs.img
```

On a tree with an existing default `rootfs.tar`, the new check in `qualcomm-linux-debian-image.yaml` correctly fires — good — but then says "rebuild rootfs.tar with `-t profile:performance`", which is what the user just did. It needs to say `rm -f rootfs.tar` (or `make clean`) first.

**4. The image recipe never validates the profile *name*,** only agreement with the rootfs. So `-t profile:perfromance` reports "rootfs.tar was built with profile 'default' but this image is being built with profile 'perfromance'" instead of "unknown profile". Adding the same `case` block there would give the accurate error.

## Robustness / duplication

**5. The valid-profile list lives in three places** — the `case` in `debos.yml`, the `case` in the rootfs recipe, and README's *Supported profiles*. The CI copy earns its keep by failing fast before a runner is burned, but a cross-reference comment in each would help, since the rootfs recipe is the actual authority.

**6. Dead branch** in the image recipe: `requested_profile="${requested_profile:-default}"` can never fire, because `{{- $profile := or .profile "default" }}` guarantees non-empty. (The `rootfs_profile="${rootfs_profile:-default}"` below it *can* fire and should stay — that's the normal default-rootfs path.)

**7. The rootfs validation runs unconditionally,** unlike the "Validate snapshot timestamp" precedent it's modelled on, which is wrapped in `{{- if $snapshot }}`. Wrapping in `{{- if .profile }}` would keep `--print-recipe` output unchanged for default builds.

**8. Style drift on reading buildinfo.** The recipe uses `sed -n 's/^BUILD_PROFILE_ID=//p'`; the two existing readers in the rootfs recipe use `grep '^SNAPSHOT=' /etc/buildinfo | cut -d= -f2`. Either is fine, but the `sed` lacks `q`, so a duplicated line would concatenate into a garbage comparison value.

**9. `append-kernel-cmdline: {{ $kernelcmdline }}`** is an unquoted YAML plain scalar. Safe with today's content; a future profile appending anything containing `:` or ` #` breaks the recipe parse in a confusing way. Quoting it now is free.

## Docs and commit messages

**10. `docs/snapshot.md` is wrong:** `# BUILD_PROFILE_ID=...   (if -t profile: was passed)` — it's written only for a *non-default* profile; `-t profile:default` writes nothing. README states this correctly.

**11. The new PREFIX comment** cites `"trixie-multimedia"` (the variant is `weston-multimedia`, so `trixie-weston-multimedia`) and `"trixie-multimedia-debug"`, referencing a `debug` profile that doesn't exist in this series. Use `performance`.

**12. `5d762e1`'s justification is factually inverted.** "Every caller already passes 'default'" — three don't: `build-debian.yml:38`, `build-on-push.yml:20` and `linux.yml:196` all omit `variant`. The change is still behaviour-neutral (both `''` and `default` give `-t xfcedesktop:true` and no PREFIX suffix), but the reasoning should be the opposite: there *are* callers that omit it, which is why a meaningful default is worth having. Also `so an missing input` → `so a missing input`.

**13. Trivia:** `Closes: #604` vs CONTRIBUTING.md's documented `Fixes: #467` form — GitHub honours both, but the repo documents one. And the new job name reaches 95 characters for the longest combination, which the checks list will truncate hard once GitHub prefixes it with the caller's matrix job name; something like `debos (trixie, weston-multimedia, performance)` would survive the UI better.

None of the above is a blocker except arguably #3, which will bite the first person who follows the README example. #1 is the one most worth the author's view on before merge, since it determines whether this profile is ever actually booted on hardware.

lool (changes requested):
Hey, so I went through all the commits, really nice split and logical incremental changes – everything made sense and is inline with exactly what I would expect for these features. TL;DR: I think this is close to what we should merge for an initial version of this.

I didn't review line by line, I mostly connected with the approach and found this sound.

In review comments, I'm only flagging a minor cosmetic thing in a comment with naming of variables.

Otherwise it also bugs me a bit that we "variants" and "profiles" and that people might find this confusing. Saying "default" might also be confusing to refer to one or the other default, perhaps we should pick unique names for each of the defaults.

The Makefile duplication is also not something new and unrelated to this PR, but it's starting to get worse.

I also asked claude to review the PR, and it was insightful (will attach), albeit most bits were harmless. From the Claude notes, point 1), 2), 10), 11), 12) are worth considering IMO.

One thing made me pause. We're on a path to grow our CI significantly, yet while the changes are easy to review and maintain, they are the opposite of what we usually thrive for with Debian: an universal OS that adapts to different situations. So I was wondering if we should handle some of these more dynamically at build time or even at runtime. For instance, we could have a command to select between profiles and it would set the kernel cmdline and kernel flavor. Leaving this thought here as

debos-recipes/qualcomm-linux-debian-rootfs.yaml
obbardc: nope, I am happy to change it to `VARIANT` and `PROFILE`.

- `VARIANT_ID` came from https://www.freedesktop.org/software/systemd/man/latest/os-release.html
- and really `BUILD_PROFILE_ID` was because I really wanted to call `PROFILE` `BUILD_PROFILE`.

Happy to change the names to whatever you think makes sense. I'd lean on `VARIANT_ID` and `PROFILE_ID` everywhere, what do you think?

Also, I think having `default` as the default for both and including `default` for both in `/etc/buildinfo` makes the most sense, but opinions welcome...

lool: Fine with me; I had forgotten about the freedesktop spec for this, it doesn't seem to have a concept like profile (a few things come close, but are different), so +1 on PROFILE_ID – it does state unknown fields should be ignored.

I was also considering amending BUILD_ID to include profile, but since we're building out of the same workflow runs, I guess that's alright (technically, there could be package updates between the two flavors being built).

obbardc: OK; changed the variables to `PROFILE=x` and `VARIANT=x`. Further changes welcome ;-).

I have taken an action to open a separate issue about adding metadata into `/etc/os-release` based on https://www.freedesktop.org/software/systemd/man/latest/os-release.html as it doesn't belong in this PR yet I don't think :-)

obbardc:
@lool 

> Otherwise it also bugs me a bit that we "variants" and "profiles" and that people might find this confusing. Saying "default" might also be confusing to refer to one or the other default, perhaps we should pick unique names for each of the defaults.

I agree, I think for `variants` we should make `minimal` or `cmdline` the default over `default`. But that doesn't belong here, it's in #609 and #507 IMO. That clean-up should be next after this.

 
> The Makefile duplication is also not something new and unrelated to this PR, but it's starting to get worse.

Yeah, I am not sure the best way to handle this. Maybe we need to consider cleaning that up very shortly.


> I also asked claude to review the PR, and it was insightful (will attach), albeit most bits were harmless. From the Claude notes, point 1), 2), 10), 11), 12) are worth considering IMO.

I'll look at those one-by-one. Very detailed. I've split the "debug" into a separate branch locally, since that depends on another kernel. That can be proposed/merged after this one which adds "performance" (no kernel changes needed).


> One thing made me pause. We're on a path to grow our CI significantly, yet while the changes are easy to review and maintain, they are the opposite of what we usually thrive for with Debian: an universal OS that adapts to different situations. So I was wondering if we should handle some of these more dynamically at build time or even at runtime. For instance, we could have a command to select between profiles and it would set the kernel cmdline and kernel flavor. Leaving this thought here as

Yeah, I think that this needs to be considered as part of #507 - my vision is that as we get bigger, we should create a shared/generic minimal "rootfs", then extend it more as the recipes increase (may be other options too). I think that cleanup should come as soon as this is in though.

obbardc:
@lool in response to your very in-depth claude review:

## Review 1 - The performance profile gets no hardware boot validation
This is a genuine issue to fix now (and also was an issue with multimedia images!). I have fixed that in this PR, but I will move that to its own PR once validated, it doesn't belong here.

## Review 2 - A whole extra rootfs for a cmdline-only difference
Yeah, this was deliberate. The long-term intent is actually the opposite: profile is expected to affect only the image recipe (e.g. kernel cmdline) not rootfs contents. It's currently passed to the rootfs anyway while the profile set is still settling. Once the profiles have settled and the rootfs is demonstrably profile-independent, we can remove the profile from the rootfs (e.g. the default rootfs can be shared) and only the image and flash stages need repeating per profile. I'd rather do that as a deliberate cleanup after all profiles are implemented.

## Review 3 - stale `rootfs.tar`
I agree - but unfortunately not much to do in scope of this PR. I will open a separate issue to track that.


All other review comments have been resolved.

.github/workflows/build.yml
```diff
@@ -64,8 +69,23 @@ jobs:
...
+      # it is left out for a suite's default image, whose prefix is just the
+      # suite and which lava-test.yml falls back to.
       matrix:
-        suite: [trixie, forky]
+        include:
+          - suite: trixie
+          - suite: forky
+          # boot the Weston/multimedia session rather than the Xfce desktop
+          - suite: trixie
+            artifact_prefix: trixie-weston-multimedia
```
lool: kind of unfortunate to repeat the suite there, I wonder if you should use:
  variant: [default, weston-multimedia]
and compute artifact_prefix from suite + variant

ci/lava/glymur-crd/boot.yaml
```diff
@@ -4,7 +4,7 @@ actions:
     qcomflash:
       headers:
         QLIAuthorization: Q_QLI_S3_TOKEN
-      url: "{{BUILD_DOWNLOAD_URL}}/{{SUITE}}-flash-emmc.tar.gz"
+      url: "{{BUILD_DOWNLOAD_URL}}/{{ARTIFACT_PREFIX}}-flash-emmc.tar.gz"
```
lool, line 7: I did some similar cleanup in a draft some time back, I think we want to bring as little internal variables to these templates. Upstream LAVA changes might not be going in this direction yet, since they start capturing specific list of UFS partitions that should be flashed with qdl method for instance, but I'd like to move towards having a single high-level LAVA template for all boards.

Anyway, I think this is fine, it's one variable being swapped for another here, so fine with me, just wanted to share the perspective in context :)

.github/workflows/test-on-pr.yml
```diff
@@ -70,13 +73,34 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      # One entry per image "Build on PR" builds, so that nothing is built on
+      # a pull request without being booted. Each entry is a full LAVA sweep
+      # across every board, which is what an image costs to validate; a
+      # variant that is built and never booted buys nothing at all, and a pull
+      # request is where a regression is cheapest to catch. Keep this list in
+      # step with the build matrix in build-on-pr.yml. artifact_prefix is
+      # omitted for a suite's default image, whose prefix is just the suite.
+      matrix:
```
lool, line 77: So that's yet another copy of our target matrix; here's another strategy:
why don't we upload the image set along with the build as a metadata file, then use it as input for the tests?

Here's yet another strategy:
why don't we query GH workflows for what was built to decide what to test?

But if I think of implementation, what I'm vaguely picturing is:
- build-on-pr triggers some event as soon as any of the builds is completed
- test-on-pr fires on any of these events, triggers a corresponding LAVA test unless we want to skip testing
- build-on-pr triggers some "all builds complete" event
- test-on-pr fires on this event and triggers a test summary generation

lool (changes requested):
I've left minor remarks, I understand we can't land this yet due to deps on kernels

The main thing I'd like to discuss/revisit is how we go about building the set of LAVA jobs. My gut feeling is that we want to convey the build set information from the moment we decide what to build down to the moment we want to send things for testing, rather than maintaining list in multiple places. I do want the ability to add/enable or skip/disable specific combos at various points, but by default everything should be tested.

Because of the kernel deps and the complexity of this PR, I'd like to propose merging it in 2-3 PRs:
- first few commits of general cleanups can go in right now (minor comments left)
- image variant concept can IMO land now as well
- LAVA / CI I feel requires further discussion, but could potentially land now
- only the debug profile needs to be deferred to landing later, when the kernel lands

obbardc:
> I've left minor remarks, I understand we can't land this yet due to deps on kernels
> 
> The main thing I'd like to discuss/revisit is how we go about building the set of LAVA jobs. My gut feeling is that we want to convey the build set information from the moment we decide what to build down to the moment we want to send things for testing, rather than maintaining list in multiple places. I do want the ability to add/enable or skip/disable specific combos at various points, but by default everything should be tested.
> 
> Because of the kernel deps and the complexity of this PR, I'd like to propose merging it in 2-3 PRs:
> 
> * first few commits of general cleanups can go in right now (minor comments left)
> * image variant concept can IMO land now as well
> * LAVA / CI I feel requires further discussion, but could potentially land now
> * only the debug profile needs to be deferred to landing later, when the kernel lands

Thanks!
I've created https://github.com/qualcomm-linux/qcom-deb-images/pull/628 for the CI / LAVA changes and we can iterate there. I'll remove those commits from this PR & move the review comments over...

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -3,6 +3,58 @@
...
+# https://github.com/qualcomm-linux/meta-qcom/blob/master/ci/debug.yml
+# commit: 7b744efa47a439c0baee87390e1508c83dff7423
+#
+# this profile is not ready for use yet: the kernel is consumed as a prebuilt
+# package here and the debug image still installs the default one, so the
+# kernel debugging options meta-qcom builds its kernel with are missing. That
+# needs the custom qcom-next-debug kernel package from
+# https://github.com/qualcomm-linux/qcom-deb-images/issues/568 and is left to a
+# follow-up; the kernel command line below is in place for it.
+{{- if eq $profile "debug" }}
```
obbardc, line 47: For debug, we also need `qcom_scm.download_mode=1 and RuntimeWatchdogSec=10s`; see https://github.com/qualcomm-linux/qcom-deb-images/issues/568#issuecomment-5601926503

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -29,6 +30,13 @@
 {{- $suite = "unstable" }}
 {{- end }}
 
+# Validate $profile; keep this list in sync with the list in the
```
obbardc, line 33: This currently builds a separate rootfs for each profile, even though the profile currently only affects the image. We could change that in this PR so profiles reuse the same rootfs; @lool would you prefer that here or as a follow-up? It will need some additional actions plumbing.

obbardc: Modified it to use one rootfs for all profiles.

lool (approved):
Re-read each commit, looks good

On revisiting this, I'm not sure we're not making a mistake by introducing two concurrent dimensions for variants and profiles, instead of a single flat one, time will tell

PR #619, opened 2026-09-07
refactor(debos): pin snapshot builds via native APT config

### LINKED ISSUE #602: Refactor snapshots to use APT config
Our snapshot implementation is a bit complex, juggling between hostnames, when this should all be possible with a couple of lines of APT config.

We should eventually refactor the implementation to use APT config, that would be in line with the README documentation on snapshot usage.

(This is low prio)

Knightkolla:
Title: refactor(debos/rootfs,debos/image): use native APT snapshot config

Fixes: #602

What I did

Refactored how the snapshot build option pins APT package installs to a dated archive, in both debos-recipes/qualcomm-linux-debian-rootfs.yaml and debos-recipes/qualcomm-linux-debian-image.yaml, plus the matching internals documentation in docs/snapshot.md.

Removed:

The apt-snapshot-toggle shell script written into the rootfs, whose only job was flipping Enabled: yes/no between live and snapshot .sources files.
The step that derived a parallel snapshot_*.sources file for each live .sources file, using a hand-maintained per-source table of mirror-URL-to-archive-URL rewrites (one entry each for debian, debian-backports, qli).
The runtime warning loop that scanned .sources files for ones with no snapshot equivalent.

Replaced with:

A single file, /etc/apt/apt.conf.d/99snapshot, containing two lines:
  APT::Snapshot "<TIMESTAMP>";
  Acquire::Check-Valid-Until "false";

This is APT's own native mechanism for resolving packages from a dated snapshot across every configured source automatically, with no per-source knowledge required.

A static warning, gated on the aptlocalrepo variable, since a local APT repo is the one source type that genuinely has no dated-archive equivalent.
In the image recipe, the pin file is recreated (from SNAPSHOT= in /etc/buildinfo, carried forward via rootfs.tar) rather than toggled back on, since the rootfs recipe now deletes it outright at the end instead of leaving it disabled.

Also rewrote the stale # TODO: remove Check-Valid-Until once Debusine-based snapshots are available comment. Per the issue's maintainer, this assumption was wrong — disabling Check-Valid-Until is a permanent, accepted trade-off (a snapshot's Release file is published once and never re-signed, so its validity window is always in the past; HTTPS is the mitigation for authenticity in transit), not a stopgap waiting on a future infra change.

Why the old approach existed and why this is better

The old mechanism worked, but required duplicating every .sources file and hand-maintaining a per-archive rewrite table, purely to work around not having a way to tell APT directly "resolve everything as of this date." That's exactly what APT::Snapshot is for. The refactor became possible because the maintainer confirmed (in the issue thread) that the Debusine-hosted qli archive now has snapshot metadata support too previously only Debian's own archive definitely supported it, so the original implementation had to special-case things.

How it solves the issue

It's a straight replacement of a manual mechanism with the native one, with the same externally-visible behavior (same snapshot variable, same requirement to pass it to both recipes). No .sources file rewriting, no duplication, no toggle script one config file, written before install and deleted after, is the entire mechanism.

What I tested, and what I didn't

No Linux build environment on my end (macOS, 8GB RAM). I ran the godebos/debos:latest container directly (bypassing the Makefile's tty requirement) against the rootfs recipe with real -t snapshot:<timestamp> runs.

Confirmed working:

debian, debian-security: every fetch was transparently redirected to snapshot.debian.org/.../<TIMESTAMP>/ by APT itself, and apt-get full-upgrade completed cleanly against the pinned archive.
qli: for a timestamp after Debusine's snapshot rollout, both Release/indexes and actual .deb packages were fetched successfully through https://deb.debusine.qualcomm.com/qualcomm/qli/<TIMESTAMP>/.... An older timestamp (predating rollout) correctly 404s on qli verified directly with curl that this is an archive-side condition, not a bug in this change: the same URL shape resolves fine for a recent date.

Not tested:

A full rootfs.tar build, the disk-ufs.img build, or booting the result. The local build got wiped out past the qli install step  the part this refactor actually touches  because the host ran out of memory running debos through a software-emulation backend under Docker Desktop. This is a local hardware constraint, not something the code changes caused.
The image recipe's re-creation of the pin file was written by hand, matching the rootfs recipe's logic, but not exercised in an actual image build.
Nothing was boot-tested with run-qemu.py.
How it fits into the pipeline

Pure internals change to one build option. Doesn't touch the flash recipe (no snapshot option there), doesn't change the Makefile, doesn't change anything outside the {{- if $snapshot }} blocks in the two recipe files — EXTRA_DEBOS_OPTS="-t snapshot:<timestamp>" works identically from the outside.

What to review
Whether recreating 99snapshot in the image recipe (rather than toggling a disabled copy left in place) is the right call  it's a small behavior change from the original design: deletes now instead of disabling.
Whether the bootstrap step's separate, unmodified snapshot-mirror-URL override (it runs host-side, before /etc/apt/apt.conf.d/ exists, so it can't use the new mechanism) is fine as-is, or whether reviewers want to explore folding it in too  I deliberately didn't guess at how mmdebstrap's internal APT invocation handles APT::Snapshot.
Whether the aptlocalrepo warning replacement is adequate  the old one checked actual .sources file state at runtime, and the new one is a static check against the aptlocalrepo variable.
A full local build couldn't complete —someone with more RAM, a real Linux box, or CI would need to run the full pipeline before merge.


lool:
@Knightkolla thanks for the PR; did you use am LLM to craft the changes? If so, could you document this in the commit pseudo-headers?

@basak-qcom would you mind handling the PR review?

Knightkolla:
Hi @lool, yes I used Claude Code to help draft these changes. I've amended the commit  with an Assisted-by: Claude Code:claude-sonnet-5 trailer, per the guidelines in CONTRIBUTING.md. I reviewed and tested the changes myself before opening the PR (see the PR description for what I verified locally against snapshot.debian.org and the qli archive), and I'm happy to answer any questions or make adjustments during review.

Thanks for the quick response, and thanks @basak-qcom  for taking a look let me know if you'd like me to change anything or if you have any questions.

basak-qcom (changes requested):
Hi! Sorry for the delay in reviewing. An AI review found a rabbithole and haven't had the time to dig into it. I wonder if you could take a look, please? I'm not sure if the issue is valid. It seems serious enough that it's worth investigating. I would next try to reproduce this behaviour in apt outside qcom-deb-images using mmdebstrap+systemd-nspawn. If it is valid, I wonder if the apt maintainer would consider this a bug in apt, and if so, the previous reproducer would be helpful in filing a bug report upstream in apt if there isn't one there already. If the issue is valid, I think it would be better for apt to detect the case and redirect itself to the snapshot Release file to continue, so that a single initial `apt update` does the right thing automatically.

Given that this is refactoring and not blocking a feature, if the issue in apt is valid then it's probably worth getting a fix for apt in the pipeline while landing this.

I think it's also worth adding some automatic verification that the snapshot selected was actually used, or perhaps that the non-snapshot was _not_ used. IMHO this would be better done in a separate test suite, but I suppose an assertion during the operation also works if that's much easier.

AI report below. I'm marking this "Request changes" since I think more investigation is needed before landing.

### 1. (High, correctness) qli is silently **not** pinned on a fresh build — ordering bug

In the rootfs recipe the pin is written (`qualcomm-linux-debian-rootfs.yaml:277-292`) *before* any `apt-get update` has fetched qli's Release. The first and only guaranteed update runs *with the pin already in place* (`:305`), immediately followed by `apt-get -y full-upgrade` (`:306`).

At that first update, qli has no Release on disk and its host isn't preseeded, so APT can't read qli's `Snapshots:` field and **falls back to live qli, silently** (the `APT::Snapshot` apt.conf path carries a `?` best-effort suffix — `debmetaindex.cc:1331` — and does `goto nosnapshot` when no snapshot server is found, `:1381`). `full-upgrade` — and any qli-sourced package it pulls (e.g. the backport/qli-pinned `fastrpc`, `alsa-ucm-conf`, and the `fastrpc-tests`/`userspace-resource-manager` installs) — therefore comes from the **live** qli archive, defeating reproducibility for precisely the archive this commit claims newly supports.

Debian/backports are unaffected because `deb.debian.org` is in APT's built-in host preseed table (`init.cc:208`), which is likely why the "verified locally … all three sources resolve correctly" check passed: either a second `apt-get update` had happened, or qli's lists were already cached in the test environment. The failure mode is silent, so a green local run doesn't clear it.

**Fix:** follow APT's intended workflow ("run update without snapshot id first"). Create all `.sources`, run one plain `apt-get update` to populate Release files (incl. qli's, with its `Snapshots:` field), *then* write `99snapshot`, *then* `apt-get update` again. Everything after that resolves through the dated archive for all three sources.

### 2. (Medium, correctness/robustness) No verification the pin took effect; the loud assert was removed

The old code hard-asserted each expected mirror URL and failed the build if absent, plus warned for any enabled source without snapshot support. The new code has **no** check that pinning actually happened, and the `?`/apt.conf path fails open. So a broken qli pin (finding #1, or a future Release that drops the `Snapshots:` field) degrades to live **with no signal**.

Given reproducibility is the whole point, add a post-update assertion — e.g. confirm the fetched qli index resolves to a dated URL (grep `/var/lib/apt/lists/` for the snapshot path, or parse `apt-get indextargets`) and fail loudly otherwise. This restores the "fail, don't silently drift" property the table+assert used to provide.

Knightkolla:
Hey  @basak-qcom , thanks for flagging this and reviewing the code .
I tested it myself and also had an AI review the issue independently, and both confirmed it's real.

Repro: rather than mmdebstrap+systemd-nspawn specifically (happy to build
that version too if it's the preferred format for an upstream apt bug
report), I reproduced it more directly with plain apt against a fresh qli
source in an isolated container, mirroring what our recipe does: writing
APT::Snapshot before apt-get update has ever touched qli. Result: apt never
attempts the snapshot-qualified URL for qli, and silently falls back to
live. I then tested the fix one plain apt-get update first (so apt caches
qli's live Release and reads its Snapshots: field), then write the pin,
then update again  and confirmed that resolves qli correctly through the
dated archive on the second pass.

Proposed fix, in both recipes:
1. Reorder: after all .sources files exist, run one plain apt-get update
   (no pin) first, then write 99snapshot, then the existing pinned
   update/full-upgrade proceeds as before. This makes every source resolve
   correctly from the start, rather than depending on some later step
   happening to re-seed it.
2. Add a loud, failing assertion right after the pinned update, checking
   that qli's cached list actually resolved to the dated URL (e.g. grepping
   /var/lib/apt/lists/ for the snapshot-qualified qli path), so a broken pin
   fails the build instead of silently drifting.

Let me know if that approach looks right before I put it up.

On your other points:

- Upstream apt bug: I'd lean toward yes  the silent fallback to live (with
  no error) seems like the part worth reporting, even if it requires a
  prior seed update as intended apt behavior. Happy to prepare the
  mmdebstrap+systemd-nspawn reproducer for that report.
- Should apt redirect itself automatically within a single update? Agreed
  that would be the ideal fix, but it's squarely apt's own behavior to
  change, not something we control from here.
- Should this PR wait on an apt-side fix landing first? I don't think it
  needs to  the recipe-side fix is self-contained (just an extra
  apt-get update call) and doesn't depend on any apt change landing. I'd
  treat the upstream bug report as a separate, non-blocking follow-up.
- Verification: going with the inline assertion for now since it's the
  fast path, open to adding proper test-suite coverage as a follow-up if
  that's preferred long-term.

Knightkolla:
Hi @basak-qcom , pushed a fix for both points, full repro, fix, and verification below.

Root cause confirmed. apt only learns that qli supports snapshots by reading
the Snapshots: field out of its live Release file, and can only do that
after fetching that file at least once in a prior, separate apt-get update.
deb.debian.org has this preseeded into apt itself and works cold; qli
doesn't. The recipe wrote the pin before any update had ever touched
qli.sources, so the first pinned update, and the full-upgrade immediately
after it, silently resolved qli live.

Fix, in both recipes:
1. Added a plain, unpinned apt-get update before 99snapshot is written, so
   qli is seeded before the pin exists.
2. Added a loud, failing assertion right after the pinned update, checking
   the cached APT list filenames for debian, debian-security, and qli all
   resolved through the dated archive, and failing the build if any didn't.
   This restores the coverage the old per-source assert table gave.

Testing, in increasing order of fidelity:
1. Reproduced the original bug in isolation (plain apt against a fresh qli
   source, pin written before any update). Confirmed apt never even
   attempts the snapshot-qualified URL for qli.
2. Confirmed the fix in the same isolated setup: seed update first, then
   pin, then update again. qli resolves correctly on the second update.
3. Broadened the assertion to also cover debian/debian-security (the old
   code checked all three; my first pass only checked qli), and
   independently validated the exact cached-filename markers against a
   real, live apt cache via docker exec before wiring it into the recipes.
4. Validated both changes inside the actual recipe files, not a
   hand-copied script. Ran the real
   debos-recipes/qualcomm-linux-debian-rootfs.yaml end to end. The seed
   step, pin step, and assertion all fired at the correct points with no
   errors; qli resolved through the snapshot-qualified URL from the very
   first pinned update this time.
5. That run completed all the way to a real rootfs.tar (previous attempts
   had died partway through from a local memory limit; this one didn't).
   Confirmed the final tarball has no leftover 99snapshot file and
   /etc/buildinfo correctly records the snapshot timestamp.
6. Fed that rootfs.tar into a full disk-ufs.img build via the image recipe.
   Same seed/pin/assertion pattern, also confirmed working, completed
   successfully.
7. Booted the resulting disk-ufs.img under QEMU (scripts/run-qemu.py
   --headless). Reached a real debian login: prompt, including
   debos-grow-rootfs.service (the first-boot partition grow) completing
   successfully.

So this is now verified at every level: isolated apt behavior, the real
recipes standalone, and a complete rootfs to image to boot pipeline.

On your other questions:
- Upstream apt bug: still think the silent live fallback (no warning) is
  the part worth reporting, separate from whether requiring a seed update
  is intended behavior. Happy to prepare the mmdebstrap+systemd-nspawn
  reproducer for that if you'd like it before or after this merges.
- Should this wait on an apt-side fix? I don't think so, the recipe-side
  fix is self-contained and doesn't depend on anything upstream.
- CI gap: separately worth flagging, I checked and snapshot isn't exercised
  anywhere in the existing on-PR workflows, so this bug wouldn't have been
  caught by CI before or after this fix. Might be worth a follow-up
  snapshot-enabled CI job, matching your "automatic verification" point, as
  a separate PR.

Let me know if you'd like anything else before re-reviewing.

basak-qcom:
Thank you for the update. AI code review still identifies problems, listed below.

If you want to continue using AI, please could iterate directly with AI rather than using me to do it until you have something that is at least acceptable to AI before I spend time reviewing it further?

Here's the AI review:

## Code review: migrate snapshot pinning to native `APT::Snapshot`

This is a good, well-motivated simplification — replacing the hand-rolled `snapshot_*.sources` derivation + toggle helper with apt's native `APT::Snapshot` config is the right direction, and the design reasoning in the comments largely matches what apt actually does. I verified the mechanism against the apt 3.3.3 source. Two findings are, I believe, build-breaking; the rest are minor.

---

### 🔴 1. Rootfs seeding `apt-get update` runs against the still-present bootstrap `sources.list`, which points at the (possibly-expired) snapshot archive with `Check-Valid-Until` re-enabled

`qualcomm-linux-debian-rootfs.yaml:277-290` adds a new *unpinned* `apt-get -y update` **before** the legacy `sources.list` is removed at `:314`.

Sequence in the snapshot path:
1. `mmdebstrap` bootstraps from `https://snapshot.debian.org/archive/debian/<SNAP>/` and writes `/etc/apt/sources.list` pointing at that **snapshot** mirror (the recipe's own comment at `:313` and `rm -vf /etc/apt/sources.list` confirm it exists).
2. `:96-100` removes `99mmdebstrap`, so `Acquire::Check-Valid-Until` reverts to its default **true**.
3. `:290` runs the seeding `apt-get -y update` — at this point `99snapshot` does **not** exist yet (`:302`), so `Check-Valid-Until` is on, and `sources.list` still points at the snapshot archive.

A snapshot archive serves the *original*, never-re-signed `Release`, so its `Valid-Until` is in the past for any snapshot older than the Debian `Release` validity window (~a week — i.e. the common case for a reproducible pin). apt treats that as a hard error, not a warning:

```
apt-pkg/acquire-item.cc:1841  if (... GetValidUntil() > 0)
apt-pkg/acquire-item.cc:1856     return _error->Error(... "Release file for %s is expired ...");
```

`apt-get update` then exits non-zero and, under `set -eux`, aborts the build. This is a **regression**: the old recipe removed `sources.list` *before* its first update, so it never re-validated the stale bootstrap source.

It'll pass CI iff CI happens to pin a recent snapshot, then fail for older pins — an easy trap.

**Fix:** move `rm -vf /etc/apt/sources.list` ahead of the seeding update (into, or before, the `:277` action). `debian.sources` already covers `main`, so nothing is lost. The image recipe is *not* affected — `rootfs.tar` no longer contains `sources.list`, so its seeding update at `:93` sees only live sources.

---

### 🔴 2. Verification markers are hardcoded but the sources they check for are conditional (`unstable`, `qliaptrepo=false`)

Both recipes assert three fixed markers exist in `/var/lib/apt/lists/` (rootfs `:328-331`, image `:115-118`):

```
snapshot.debian.org_archive_debian_${SNAPSHOT}_dists
snapshot.debian.org_archive_debian-security_${SNAPSHOT}_dists
qualcomm_qli_${SNAPSHOT}_dists
```

But whether those sources are even created is gated by templating:
- `debian-security` (and `-updates`) exist only when `{{ ne $suite "unstable" }}` (`:159`). For `suite=unstable` there is **no** security source, so the `debian-security` marker can never appear → verification fails → build aborts.
- `qli.sources` exists only when `{{ and (eq $qliaptrepo "true") (ne $suite "unstable") }}` (`:232`). So `suite=unstable` **or** `qliaptrepo=false` → no qli marker → build aborts.

Net effect: snapshot builds are broken for `unstable` entirely, and for any build with `qliaptrepo=false`. The old `snapshot_*.sources` derivation handled these gracefully (it looped over whatever `*.sources` actually existed). The verification is a genuinely nice addition — apt *silently* falls back to live for an unresolvable source (see finding 3), so failing loudly is the right instinct — but the marker set must track the same template conditions that create the sources. Gate each marker behind the matching `{{- if }}`.

---

### 🟢 3. (Confirmation, not a defect) The silent-fallback risk the verification guards against is real

Worth recording why finding 2's verification matters, since it justifies keeping it (just fixing the conditions). Because the `.sources` files carry no per-source `snapshot=` option, `APT::Snapshot` reaches each source as `"<SNAP>?"` (`debmetaindex.cc:1330-1331`). The trailing `?` means "best-effort": if apt can't resolve a snapshot server for a source, it does **not** error — it silently uses the live URL (`debmetaindex.cc:1370-1382`, `goto nosnapshot`).

`qli`'s snapshot server is resolvable *only* from its `Release`'s `Snapshots:` field (`SnapshotServer()`, `debmetaindex.cc:1307`) — there is no `Host::deb.debusine.qualcomm.com` fallback in `init.cc`, unlike `deb.debian.org` (`init.cc:208`). So from a cold cache, `qli` silently resolves **live**, which is exactly what the seeding update prevents and what the marker check catches. The design reasoning in the comments (`:283-289`) is accurate. So: keep the check, just make its expected-marker set conditional.

One fragility note: the `qualcomm_qli_${SNAPSHOT}_dists` substring assumes the qli `Snapshots:` server lays the path out as `.../qualcomm/qli/<SNAP>/`. If Debusine ever changes that layout the check will false-fail. Fine for 3.3.3/trixie, but it's an undocumented coupling to an archive-controlled value — maybe a one-line comment.

---

### 🟢 4. Minor

- **`apt-get -y update` vs `apt-get update`** — rootfs uses `-y` (`:290`, `:319`), image uses bare `apt-get update` (`:93`, `:108`). `-y` is a no-op for `update`; harmless but inconsistent.
- **`docs/snapshot.md:305-323`** reframes `Check-Valid-Until "false"` from a `TODO` to a "permanent trade-off (confirmed with Debusine maintainers, issue #602)". Good — that matches the archive-side model. It's a build-time-only override now (removed with `99snapshot` before packing), which the doc correctly states; the shipped image no longer disables the check, which is a real improvement over any global override.
- The docs are otherwise accurate to the new behavior and to apt's model. The claim "every configured source ... resolves through the dated archive automatically" is true *given* the seeding update has populated each archive's `Snapshots:` field — worth a half-sentence, since it's the linchpin the whole thing rests on.

---

### Summary

The refactor is a clear net improvement and the comments show the author understood apt's model. Findings **1** and **2** are both build-breaking on realistic inputs (old snapshot dates; `unstable`; `qliaptrepo=false`) and both are regressions against the more defensive old recipe — I'd block on those. Everything else is polish.

Knightkolla:
Hey,@basak-qcom thanks for the detailed review and for being straight with me about
the back and forth. Fair, and honestly needed to hear it.

Going to actually go through this properly now, read the apt source around
the lines you pointed to, trace the recipe step by step myself, and test the
exact edge cases that broke (old snapshot date, unstable, qliaptrepo=false)
before I push anything again. Won't come back until I've actually checked it
myself, not just until a review says it looks ok.

Thanks for taking the time on this.

Issue #620, opened 2026-09-07
Regular qcom-next CI failures on monaco-evk and lemans-evk


lool:
Since our CI revamp to rename jobs and default to qcom-next as the main kernel, the daily build has been mostly red. This ticket is to document the research around current failures, to discuss ignoring a couple of boards temporarily, and to keep an open ticket to re-enable them later.

https://github.com/qualcomm-linux/kernel/issues/758 is the main ticket for the most common monaco-evk issue
https://github.com/qualcomm-linux/kernel/issues/1053 is a new ticket for the lemans-evk issue

lool:
I asked a LLM to review test failures from past weeks; monaco-evk and lemans-evk were the top offenders, so then I asked for a breakdown of issues:

# qcom-deb-images CI boot-test failure summary

This is a summary of the recent `Build - build.yml` GitHub Actions CI runs for:

- Repository: <https://github.com/qualcomm-linux/qcom-deb-images>
- Workflow: <https://github.com/qualcomm-linux/qcom-deb-images/actions/workflows/build.yml>

Data was collected with `gh` from the workflow runs and with `lavacli` from LAVA via `ssh trixie`.

## Scope

- Window inspected: **2026-08-13 through 2026-09-07 UTC**
- Workflow runs inspected: **33** latest `build.yml` runs available at the time of analysis
- Boot jobs counted were jobs named:

  ```text
  test (<suite>) / Submit <board> boot
  ```

- Suites seen in these boot jobs: `trixie`, `forky`

## Overall boot-test failures by board

| Board | Failed boot jobs | Suite breakdown |
|---|---:|---|
| `lemans-evk` | 15 | `trixie`: 9, `forky`: 6 |
| `monaco-evk` | 13 | `trixie`: 7, `forky`: 6 |
| `qrb2210-rb1` | 8 | `forky`: 5, `trixie`: 3 |
| `qcs8300-ride` | 1 | `trixie`: 1 |

## Weekly failure breakdown

| ISO week | Failed boards |
|---|---|
| 2026-W33 | `monaco-evk`: 3, `lemans-evk`: 2 |
| 2026-W34 | `lemans-evk`: 5, `qrb2210-rb1`: 3, `monaco-evk`: 2 |
| 2026-W35 | `lemans-evk`: 5, `qrb2210-rb1`: 4, `monaco-evk`: 2 |
[...]

PR #621, opened 2026-09-07
chore: disable lemans-evk and monaco-evk

lool:
Temporarily disable lemans-evk and monaco-evk which fail CI; see #620. Cleanup a workflow while at it.
- **ci: exclude lemans-evk and monaco-evk**
- **ci: drop inert configs from qcom-next build**


koenkooi (approved):
Looks good to me, but as we discussed, getting the kernel fixed and reinstating the boards is the end goal.

lool:
I checked in today, and relief is still weeks away, so I'm going to ahead and merge this.

PR #627, opened 2026-09-09
fix(debos): use :any qualifier for all APT pins

lool:
When pinning source packages, the :any qualifier is needed to match
arch: all packages, otherwise only packages from the native
architecture are pinned. Add :any to all source pins as we don't
want inconsistencies between packages from the same source.

### DIFF
--- a/debos-recipes/qualcomm-linux-debian-rootfs.yaml
+++ b/debos-recipes/qualcomm-linux-debian-rootfs.yaml
@@ -205,15 +205,15 @@ actions:
 {{- $backportpackages := list
       "src:alsa-lib:any"
       "src:alsa-ucm-conf:any"
-      "src:efivar"
-      "src:fastrpc"
+      "src:efivar:any"
+      "src:fastrpc:any"
       "src:firmware-free:any"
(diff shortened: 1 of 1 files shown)


obbardc (approved):

basak-qcom:
Are you sure this works? `src:fastrpc:any` to me doesn't make sense, since a source package doesn't have an architecture. Are you sure that apt maps this to binary architectures? Have you validated the desired outcome of this PR yet please?

lool:
> Are you sure this works? `src:fastrpc:any` to me doesn't make sense, since a source package doesn't have an architecture. Are you sure that apt maps this to binary architectures? Have you validated the desired outcome of this PR yet please?

Yes, I'm sure we need this, I already researched this when we first introduced the pins. The apt_preferences man page explains this syntax. src:fastrpc actually refers to all the binary packages that came from this source, but the default is src:fastrpc:native which means only for the current native arch (so arm64 on an arm64 system) and doesn't include arch:all packages. I find the APT default a weird and surprising behavior.

We're already using this syntax, but inconsistently – see the diff.

PR #628, opened 2026-09-09
ci: boot every image variant a build produces

obbardc:
The LAVA tests currently only ever boot a suite's default image, so any
non-default variant a build produces (e.g. `weston-multimedia`) currently gets built
but never booted. This series lets callers name the image to test and uses that to give the `weston-multimedia`
variants (and pull requests) real hardware coverage.

Draft as this will need reworking once #405 lands.


.github/workflows/test-on-pr.yml
```diff
@@ -132,13 +142,20 @@ jobs:
...
-          for json_file in $(find ${{ github.workspace }} -name "*test-job-*.json")
+          # sorted so that the jobs of one image stay together in the comment
+          for json_file in $(find ${{ github.workspace }} -name "*test-job-*.json" | sort)
           do
               DEVICE_TYPE=$(cat "$json_file" | jq -r ".requested_device_type")
               URL=$(cat "$json_file" | jq -r ".url")
               JOB_ID=$(cat "$json_file" | jq -r ".id")
-              echo " * [Job $JOB_ID on $DEVICE_TYPE]($URL)"
-              echo " * [Job $JOB_ID on $DEVICE_TYPE]($URL)" >> pr-comment.txt
+              # more than one image is booted, so say which one this job ran.
```
obbardc, line 151: I am not 100% sure about this change and it's not overly simple to validate changes to workflows triggered by `on: pull_request`.

obbardc: I looked at this in more detail and have manually checked the logic. At least from inspection it appears to do the right thing...

.github/workflows/test-on-pr.yml
```diff
@@ -132,13 +142,22 @@ jobs:
...
+          # sorted so that the jobs of one image stay together in the comment
+          for json_file in $(find "$GITHUB_WORKSPACE" -name "*test-job-*.json" | sort)
           do
               DEVICE_TYPE=$(cat "$json_file" | jq -r ".requested_device_type")
               URL=$(cat "$json_file" | jq -r ".url")
               JOB_ID=$(cat "$json_file" | jq -r ".id")
-              echo " * [Job $JOB_ID on $DEVICE_TYPE]($URL)"
-              echo " * [Job $JOB_ID on $DEVICE_TYPE]($URL)" >> pr-comment.txt
+              # more than one image is booted, so say which one this job ran.
+              # lava-action names the file "<ARTIFACT_PREFIX>-<run attempt>-test
```
lool, line 152: Can we rely on this as part of the lava-action contract/API?

lool (commented):
Hey, I did my own review once again, I was happy with most things and left just one comment. Then I also asked for a claude review; I'm pasting the tail of it below as it's not too long.

I'm a little concerned about LAVA pressure, but we can address this as a followup (lead with these jobs, then do these other jobs).

I haven't time to fully digest these yet

---

1. One failed matrix leg now discards the entire PR report (test-on-pr.yml:89). publish-test-results is needs: test with no if:. fail-fast: false stops cancellation but doesn't satisfy a needs gate. Going 1 leg → 4 means any single leg failure — a rejected LAVA submission, a submit-job timeout — yields no test-results check and no job-links comment, despite three images having produced good results. lava-test.yml:191 is ungated the same way. Wants if: ${{ !cancelled() }} on both.

2. The prefix is duplicated across two files resolved from different refs (test-on-pr.yml:80). Both sites carry a "keep in sync" comment, but test-on-pr.yml is workflow_run-triggered, so it — and the ./.github/workflows/lava-test.yml it calls — comes from main, while debos.yml runs from the PR head. Change the naming scheme in a future PR and the test half computes the old prefix: every qcomflash download 404s, all boards fail, no obvious cause.

This bites the PR in front of you: the job-links comment on #628 lists 7 jobs with no image label, i.e. main's test-on-pr.yml ran. None of this PR's test-on-pr.yml changes have been exercised by CI — the prefix-recovery loop and both new download patterns are untested. That's the thing I'd want addressed before merge, less for the bug than because the riskiest file in the series is unvalidated and structurally can't self-validate.

3. Four nested reports land on main instead of the PR (test-on-pr.yml:63). report_test_results_externally defaults true and isn't overridden, so all four nested publish jobs report externally alongside the new aggregate: five check runs named "Test Results", and since the nested ones pass no commit they default to github.sha — the default-branch head under workflow_run. build-debian.yml:80 already sets this false.

4. LAVA job_name no longer identifies the image (ci/lava/*/boot.yaml:65, all nine). Four jobs per board per run with byte-identical name and metadata — in the LAVA UI, the queue, and lavacli jobs list you can't tell which image failed without opening the job and reading the qcomflash URL. {{ARTIFACT_PREFIX}} is already substituted into these files, so this is nearly free.

5. PR hardware cost quadruples (test-on-pr.yml:74). ~28 concurrent LAVA jobs against a lab with one device per type. Four serialise per board while four runners sit in wait_for_job: true; job timeout 210 min vs. step timeout 240 min, so contention can time a leg out — which by finding 1 then discards everything. The commit message argues the cost honestly rather than hiding it, so this is a capacity call for you, not a defect.

Also noted but not filed: needs: build in build.yml:73 / build-on-push.yml:33 requires success, so one broken variant build skips hardware testing for all variants (pre-existing, costlier now); build-on-push.yml:47 reads artifacts_url from a matrix job where outputs collide, benign only because every leg uploads to the same S3 directory — which is what makes the whole prefix scheme work, so it deserves a sentence in the debos.yml comment; and lava-test.yml:235's near-identical loop wasn't updated alongside the test-on-pr.yml one, so the two have diverged.

.github/workflows/test-on-pr.yml
obbardc: Agree; I removed that for now and left it how it was.

obbardc:
This will need reworking once #405 lands. Suggest we land the other PR first. Marking this as a draft for now.

Issue #572, opened 2026-08-14
RB1 boot failures on mainline/linux-next


obbardc:
Issue to diagnose RB1 boot failures.

## Initial background

- qcom-next appears to boot fine
- Debian tests are disabled for this board, will bring-up in separate issue
- mainline/linux-next appears to work fine on qrb2210-arduino-imola, which is similar hardware... 
  - https://lava.infra.foundries.io/scheduler/job/362786
  - https://lava.infra.foundries.io/scheduler/job/362585
  - https://lava.infra.foundries.io/scheduler/job/362579 (trixie)

## Warning: incorrect FAT sector size

U-Boot complains about incorrect FAT sector size:
```
FAT sector size mismatch (fs=4096, dev=512)
```

Note the `MMC: no card present` in U-Boot is a red herring; it is the same on working jobs.

- Failure: Appears to hang/reboot loop

https://lava.infra.foundries.io/scheduler/job/363164
https://lava.infra.foundries.io/scheduler/job/363231
https://lava.infra.foundries.io/scheduler/job/363331

Possibly due to fastrpc?

```
[   15.020516] qcom,fastrpc-cb ab00000.remoteproc:glink-edge:fastrpc:compute-cb@3: mem mmap error, fd 11, vaddr ffff901f0000, size 262144
         Starting [0;1;39mdbus.service[0m - D-Bus System Message Bus...

Format: Log Type - Time(microsec) - Message - Optional Info
```

This also causes a boot-loop:
```
[    0.000000] Booting Linux on physical CPU 0x0000000000 [0x51af8014]
[    0.000000] Linux version 7.2.0-rc7-g2f1baf1fc892 (root@83b8b46678f7) (aarch64-linux-gnu-gcc (Debian 14.2.0-19) 14.2.0, GNU ld (GNU Binutils for Debian) 2.44) #1 SMP PREEMPT Fri Aug 14 13:02:05 UTC 2026
...
login-action timed out after 169 seconds
```

obbardc:
OK qrb2210-arduino-imola does seem to boot with U-Boot and thus is red herring; the debug output just appears to not show from UART :-).

The arduino boards appear to be rock solid; I wonder if their U-Boot is newer or different config against ours on RB1? Would be good to verify https://github.com/arduino/u-boot (waiting confirmation that this is the correct branch etc).

[...]

PR #630, opened 2026-09-10
fix!: disable fastrpc audiopd service on QRB2210 devices

obbardc:
fastrpc audiopd is currently broken on QRB2210 devices (RB1 and
Arduino Uno Q), pending a fix in the kernel.

Add a workaround-fastrpc-audiopd-qrb2210 overlay dropping a systemd
unit override that skips the service on those devices and enable it
in the kernel image builds which doesn't yet have the patch:
linux-arduino, linux-mainline and linux-next.

The override is keyed on the device-tree compatible, so it is inert on
every other board and can be dropped once the underlying issue is
fixed.

See: https://github.com/qualcomm-linux/qcom-deb-images/issues/572


debos-recipes/overlays/workaround-fastrpc-audiopd-qrb2210/usr/lib/systemd/system/adsprpcd_audiopd.service.d/10-qrb2210-disable-audiopd.conf
```diff
@@ -0,0 +1,4 @@
+# Disable fastrpc audiopd on QRB2210 devices (RB1 and Arduino Uno Q); see
+# https://github.com/qualcomm-linux/qcom-deb-images/issues/572
+[Unit]
+ConditionFirmware=!device-tree-compatible(qcom,qrb2210)
```
lool, line 4: wow that's a neat trick

if relevant, I'd prefer landing that in pkg-fastrpc though, and in Debian

Is there a way to test for specific DT properties so that we only disable it on older DTs that don't have the fix?

obbardc: unfortunately, systemd has: `"device-tree-compatible(value)" matches systems with a device tree that are compatible with "value".`

I mean, we could check the DTB with some scripting magic, but that's not a neat trick anymore :p.

happy to move this to pkg-fastrpc and properly check if the DT is fixed, if you like?

lool: If we're going to do it, let's do it on the fastrpc side; happy with the hammer you have right now, and removing it later, or with a more subtle hammer that disables itself when the DT is fixed.

basak-qcom: I agree, but I have a reason that I don't see called out here. If we do this in the image generator, a system flashed with such an image will be stuck with it forever. If we do it in packaging, a packaging update can remove it when it is no longer required.

I don't know if we are for sure moving to a model where we maintain this property across the board. I guess that's a decision we need to make.

obbardc: I agree, this was initially created simply to test the idea. I have created a PR for the packaging: https://github.com/qualcomm-linux/pkg-fastrpc/pull/97

I will close this PR now, but I think the first cleanup commit is still useful. I will open another PR for that commit.

PR #631, opened 2026-09-10
feat: multimedia variant improvements

obbardc:
Multimedia variant changes:
- prefer greetd from backports to allow password change on first login from westonsession
- install more camx packages in multimedia image & increase multimedia image size
- only build multimedia image for trixie


obbardc:
There are also these packages in QLI; I left them out of this PR as they are not strictly "QTI multimedia" apps, but arguably the audioreach/tinyalsa/tinycompress packages could be...

```
# audioreach-conf
# - audioreach-conf
# - audioreach-calibration-data

# tinyalsa
# - tinyalsa                                # provides /usr/bin/tinycap, /usr/bin/tinymix, /usr/bin/tinypcminfo, /usr/bin/tinyplay
# - tinycompress                            # provides /usr/bin/cplay, /usr/bin/crecord, /usr/bin/sofprobeclient


# Test suites
# - tinyalsa-tests                          # /usr/libexec/tinyalsa/tinyalsa_test
# - userspace-resource-manager-tests        # provides /usr/bin/UrmComponentTests, /usr/bin/UrmIntegrationTests
# - qmi-framework-utils                     # not sure what this does

# Broken apps - diag
# - qcom-diag                               # broken test app?
# - qcom-diag-router                        # broken service?

# broken sensor apps
# - qcom-sensors-test-apps                  # broken app?
# - sensinghub-session-client               # broken app?
```

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -815,11 +816,25 @@ actions:
...
       # camx-dkms fails to build in fakemachine; see https://github.com/qualcomm-linux/pkg-camx-dkms/pull/42
       #- camx-dkms
       - camx-firmware-glymur
       - camx-firmware-hamoa
+      # CamX camera service, exposes the cameras to userspace
+      - qti-camera-server
+{{- if eq $suite "trixie" }}
+      # device-specific CamX libraries; only pulled in on trixie as the
+      # camx-glymur and camx-hamoa packages it depends on are not ready
+      # for forky yet; see
```
obbardc: thinking about it, the prebuilts are only released for trixie so this may make sense and won't be fixed for a while?

another separate thought, perhaps we should gate the multimedia image on trixie only ?

lool (changes requested):
looks like there are some build/install issues, but otherwise happy with the changes

lool (approved):
Clean, each commit made sense

For multimedia, I think it's ok for now, but we want to add back a forky flavor with less components "soon"

PR #635, opened 2026-09-16
ci: allow kernel workflows to select rootfs overlays

obbardc:
debos.yml already takes an "overlays" input but linux.yml has no way to set it: linux-arduino.yml had to smuggle it in through `debos_extra_args` as `-t overlays:...`.

Add an "overlays" input to linux.yml and forward it to debos.yml so kernel workflows can select rootfs overlays directly. Switch linux-arduino.yml over to it and describe the input the same way in all workflows.


gagath (approved):
LGTM; we probably will need to revisit the "overlays" name in the future to something more specific like "rootfs_overlays" as we also have package overlays and (unrelated but similar name) device-tree overlays.

lool:
@obbardc sorry, conflicts, please rebase

lool (approved):

Issue #227, opened 2026-02-03
Support for device tree autoloading for Compute targets


lool:
Hi

Currently, qcom-deb-images debos-recipes use systemd-boot and supports mainly Qualcomm IoT boards (and QEMU).

With regular QLI Qualcomm IoT boards, the device tree is supposed to be loaded from the boot firmware. It can either be provided by a dtb.bin partition (supported by current debos-recipes) or a FIT image partition for some boards (not yet implemented in debos-recipes).

In debos-recipes, systemd-boot can also be configured to load a specific device tree instead of using the DT from the boot firmware, overriding the boot firmware DT entirely.

For Compute targets (laptops/desktops) such as Snapdragon laptops, there is often no DT coming from the boot firmware (typically designed for Windows), or an improper one. Ubuntu has initially implemented GRUB changes + configs to select the right DT automatically, and later implemented stubble, a mini-systemd-boot fork to do this within a simplified UKI.

qcom-deb-images should support automatic loading of the target device tree to demonstrate booting on Compute targets, and we should work with Debian to support this in regular boot media somehow.

lool:
> > > > stubble (ubuntu's loader) is in debian unstable now, I can backport it if there's interest.
> > > 
> > > 
> > > Oh wow, I checked like daily last week to see if it was still in NEW; thanks!! I'll try with that one
> > > Yeah, a trixie backport would be awesome, happy to help there too if you like :)
> > 
> > 
> > ...
> > > fwiw I uploaded the latest upstream release (7) to Debian just now, see https://tracker.debian.org/pkg/stubble
> > 
> > 
> > I just uploaded `stubble_7-1~bpo13+1_arm64.changes` into backports-new :-)
> 
> I don't know if I should be able to see it there: https://ftp-master.debian.org/backports-new.html but I'm also guessing it needs to migrate to testing/forky before being considered for trixie-backports.

Indeed:
https://lists.debian.org/debian-backports/2026/02/msg00001.html

lool:
@gagath FYI

gagath:
I was able to reproduce all the steps with the stubble version in trixie-backports and boot the generated UKI on my T14s! :cake:

[...]

PR #642, opened 2026-09-20
feat: Support for Windows on Arm laptops

lool:
Initial support for Windows on Arm laptops:
- **fix(debos/image): don't wait for TPM2 devices**
- **feat(debos): Add dtb-loader overlay**
- **feat(ci): Enable dtb-loader overlay**


lool:
(see also #227)

@gagath @obbardc would love some feedback on this one (keeping it draft while you review the approach)! We could branch initramfs-tools and stubble in our QLI APT overlay, however I think it would be nice to see how well vanilla Debian packages work, except a bit of a glue is missing.

Known opens:
- hook not tested with GRUB
- not tested on trixie
- need to confirm if we can disable waiting for TPM2 everywhere
- initramfs-tools improvements missing

lool:
I boot tested the current image on a hamoa T14s, it mostly starts, but I get a black screen after initramfs, it seems due to:
https://github.com/qualcomm-linux/kernel/issues/1148

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -2,13 +2,20 @@
...
 
 # kernel command line appended by the filesystem-deploy action below
 # XXX these kernel options might be specific to a kernel version or board
 {{- $kernelcmdline := "clk_ignore_unused pd_ignore_unused audit=0 deferred_probe_timeout=30" }}
 
+# Snapdragon Windows-on-ARM laptops advertise a TPM2 in the EFI
+# variables but expose no device for it ("TPM2 Support: firmware
+# only, driver unavailable" in bootctl status), so the systemd units
+# that wait for one timeout TODO check if this is detrimental on IoT
+# systems
```
gagath, line 16: This explanation seems to be a bit contradictory to what is stated here: https://fedoraproject.org/wiki/Snapdragon_WoA_Laptop_Install#systemd.tpm2_wait=0

The comment says that the laptop does not expose a device, the wiki says the kernel does not support the device. Which one is it?

debos-recipes/overlays/dtb-loader/etc/kernel/install.d/95-dtb-loader.install
```diff
@@ -0,0 +1,118 @@
...
+        echo "$*" >&2
+    fi
+}
+
+# This hook is likely boot critical when present, but it's technically not
+# needed on some platforms, so make sure to log some warnings before bailing
+# out.
+fatal_w() {
+    echo "Warning: $*. Not embedding device trees." >&2
+    exit 0
```
gagath, line 48: I do not understand why its not exit 1 if it is a fatal error. It should not be installed on platforms that do not need it in the first place? Also this kind of function has historically been called `die` which is well-known that it will exit the program after displaying a message.

debos-recipes/overlays/dtb-loader/etc/kernel/install.d/95-dtb-loader.install
```diff
@@ -0,0 +1,118 @@
...
+}
+
+# With layout=auto or no layout set, recent kernel-install will set the layout
+# to bls when installing kernels that use UAPI.1 Boot Loader Specification[2]
+# Type #1 layout which is the only case where the post-processing done here is
+# useful.
```
gagath, line 54: This is an implementation detail of kernel-install, why is it stated here instead of just saying we only support bls? (and removing the comment as well)

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -257,6 +264,39 @@ actions:
       # enable unit
       systemctl enable debos-grow-rootfs
 
+  - action: run
+    description: Verify device trees were embedded in the kernel image
+    chroot: true
+    command: |
+      set -eux
```
gagath, line 271: can we extract this in a dedicated script so we can have syntax highlighting, be able to run it locally as well as running shellcheck against it?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -767,6 +767,28 @@ actions:
...
+    recommends: true
+    packages:
+      # provides ukify, which embeds the device trees into the kernel image
+      - systemd-ukify
+      # provides the hardware ID database and the finddtbs script
+      - stubble
+      # provides linuxaa64.efi.stub, which is what ukify wraps the kernel in.
+      # systemd-boot's dependency is satisfied by systemd-boot-efi-arm64-signed,
+      # which ships the signed boot loader but no stub, and systemd-ukify only
+      # recommends this, so ask for it explicitly.
```
gagath, line 789: If systemd-ukify recommends it, and we install with recommends, why do we also ask explicitly?

gagath (commented):
I need to test it on a real image and do a test-boot on T14s, here are some comments in the meantime from a first quick look at the implementation.

debos-recipes/overlays/dtb-loader/etc/kernel/install.d/95-dtb-loader.install
obbardc: My 2c: I don't think installing this based on platform is a good idea as part of this PR, since currently we build generic images.

Instead it may be worthwhile bubbling the error up at runtime IF this platform needs the dtb-loader installed ?

debos-recipes/overlays/dtb-loader/etc/kernel/install.d/95-dtb-loader.install
```diff
@@ -0,0 +1,118 @@
...
+fi
+
+# 90-loaderentry.install has already copied the kernel to ENTRY_DIR_ABS and
+# pointed the entry at it; replace it in place with the wrapped image.
+KERNEL_DEST="$ENTRY_DIR_ABS/linux"
+[ -n "$KERNEL_IMAGE" ] || KERNEL_IMAGE="$KERNEL_DEST"
+
+# Prefer stubble's hardware ID database which is currently being released more
+# frequently than systemd.
+# TODO Should we load from both databases? Perhaps a patch to finddtbs.py
```
obbardc, line 67: We should create an issue to follow-up on this TODO after merging.

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -2,13 +2,20 @@
 {{- $imagesize := or .imagesize "6GiB" }}
 {{- $imagetype := or .imagetype "ufs" }}
 {{- $image := printf "disk-%s.img" $imagetype }}
-{{- $snapshot := or .snapshot "" }}
 {{- $profile := or .profile "default" }}
+{{- $snapshot := or .snapshot "" }}
```
obbardc, line 6: spurious change?

debos-recipes/qualcomm-linux-debian-rootfs.yaml
```diff
@@ -767,6 +767,28 @@ actions:
 {{- range $kernelpackage := split "," $kernelpackages }}
       - {{$kernelpackage}}
 {{- end }}
+{{- end }}
+
+{{- range $overlay := split "," $overlays }}
```
obbardc, line 772: Just wondering if it makes sense to install these packages in the image recipe unconditionally, since we already install some bootloader packages in the image recipe?

(may have to force-disable ukify hook scripts too, when `dtb-loader` is not enabled)

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -2,13 +2,20 @@
...
 
 # kernel command line appended by the filesystem-deploy action below
 # XXX these kernel options might be specific to a kernel version or board
 {{- $kernelcmdline := "clk_ignore_unused pd_ignore_unused audit=0 deferred_probe_timeout=30" }}
 
+# Snapdragon Windows-on-ARM laptops advertise a TPM2 in the EFI
+# variables but expose no device for it ("TPM2 Support: firmware
+# only, driver unavailable" in bootctl status), so the systemd units
+# that wait for one timeout TODO check if this is detrimental on IoT
```
obbardc, line 15: the TODO really means we may have to append this only when `dtb-loader` is not enabled...

debos-recipes/qualcomm-linux-debian-image.yaml
```diff
@@ -257,6 +264,39 @@ actions:
       # enable unit
       systemctl enable debos-grow-rootfs
 
+  - action: run
+    description: Verify device trees were embedded in the kernel image
```
obbardc, line 268: also, should this run when the `dtb-loader` overlay is not enabled? maybe there is a better way of doing that.

obbardc (approved):
LGTM, some minor comments.

debos-recipes/qualcomm-linux-debian-image.yaml
lool: it's two ways of saying the same thing; there is no device exposed to userspace because there is no kernel driver for it

debos-recipes/overlays/dtb-loader/etc/kernel/install.d/95-dtb-loader.install
lool: > I do not understand why its not exit 1 if it is a fatal error. It should not be installed on platforms that do not need it in the first place? Also this kind of function has historically been called `die` which is well-known that it will exit the program after displaying a message.

I didn't call that die() or fail() because I didn't want to exit 1.

> My 2c: I don't think installing this based on platform is a good idea as part of this PR, since currently we build generic images.

Yes, I want us to build generic images; since I know laptops need a number of workarounds, and some of them belong in packages, my initial thought was to put these behind a flag so that we have a way to trace them back, and return to clean images.

Very keen to hear about better / alternate ways of approaching this!

> Instead it may be worthwhile bubbling the error up at runtime IF this platform needs the dtb-loader installed ?

The reason I'm not failing unconditionally is because I wasn't sure how to deliver that without breaking other legitimate uses, for instance:
- I didn't want to break people like Agathe that have a working local UKI setup already
- I didn't want to break chroot builds
- I find it hard to recover from package postinsts failing, and these hooks are typically run from package postinsts

I'm trying to weigh the pros and cons :)

debos-recipes/overlays/dtb-loader/etc/kernel/install.d/95-dtb-loader.install
lool: When I developed this file and was reviewing it, I had to learn about the different layouts supported by the config, and then understand that not all of them would be reflected at runtime. This was my way of bringing it back in context here. Some day, there might be new layout types added.

debos-recipes/overlays/dtb-loader/etc/kernel/install.d/95-dtb-loader.install
obbardc: All good for me, let's come up with a clever solution to be more generic later when we need to, not blocking now :-)!

debos-recipes/qualcomm-linux-debian-image.yaml
lool: took the opportunity to sort alphabetically as I was adding an option

debos-recipes/qualcomm-linux-debian-rootfs.yaml
lool: it's the kind of functionality that we require, so while the package has a recommends, we have a depends, so I thought it was proper to list the dependency

debos-recipes/qualcomm-linux-debian-image.yaml
lool: right; I think I just want an ukify <something something list contents> command instead

debos-recipes/qualcomm-linux-debian-image.yaml
lool: you're right, that's ugly there, I'll drop that

